PR #235, opened 2026-04-10
[codex] Harden build signing preflight

r3dbars:
## Summary
- move local and beta entitlements into committed files under `config/entitlements/`
- harden `build.sh` and `build-beta.sh` with stricter prereq checks, explicit signing identity selection, and signature verification
- document the split between local builds and distribution packaging in `README.md` and `docs/release-packaging.md`

## Why
The local build flow could leave behind a misleading `.app` bundle when dependency artifacts were missing or stale. That made signing failures look like entitlement or Gatekeeper problems when the real issue was earlier in the build pipeline.

## Impact
- local builds now fail fast before signing when `TranscriptedCore` dependency artifacts are missing
- both build scripts require a real app binary before signing
- signing configuration is reviewable in version-controlled entitlement files instead of inline heredocs
- release packaging docs now make the local-vs-distribution flow explicit

## Root Cause
`build.sh` and `build-beta.sh` mixed artifact assembly, dependency assumptions, inline entitlements, and signing in a way that could mask the first real failure. When the app binary was never produced, the scripts still left behind a partial bundle that invited the wrong diagnosis.

## Validation
- `bash -n build.sh build-beta.sh`
- `plutil -lint config/entitlements/local.plist`
- `plutil -lint config/entitlements/beta.plist`
- `bash build.sh` now fails early with a clear missing-dependency message on a clean checkout

## Follow-up
`build-deps.sh` still needs separate stabilization work before full end-to-end app builds succeed from a fresh checkout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,13 @@ bash build-deps.sh
 bash build.sh
 ```
 
+`build.sh` is the local development path. It expects the unified dependency
+artifacts from `build-deps.sh` and then signs the app for stable local
+permissions on the current machine.
+
+For signed DMG packaging and notarization workflow notes, see
+`docs/release-packaging.md`.
+
 ## Tests
 
 ```bash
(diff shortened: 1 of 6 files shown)


r3dbars:
Already included in `main` via merged branch work, so I’m closing this duplicate PR.

Issue #312, opened 2026-04-14
Identify multiple local speakers


TeaPowered:
**What's the job to be done?**
When I have meetings I there's often multiple people in the same space whilst others are remote. I want to separate transcripts into different speakers even if they're local.

**What's awkward today?**
Transcripted treats everyone local as "me" regardless of who was speaking.

**Area**
- Dictation
- Meetings

**What would feel better?**
Speakers could be identified separated in the same way as remote meeting speakers

**Why this matters**
I would be able to use Transcripted more frequently in everyday meetings

**Alternatives considered**
I was using Notion in the past, which transcribed meetings effectively, but I wanted something local.

**Anything else?**
I'm a big fan of Transcripted! It's made meeting notes SO much easier already

r3dbars:
Thank you for the feedback! This is a really pointed feedback and I am going to get to work on testing this! 

Just so I understand, this is when you are in a meeting with others in the same room that come through the mic correct?

TeaPowered:
That’s correct. For example, I had a meeting earlier today with 4 people in the same room and 3  others connecting over Google Meet.

Transcripted could separately identify the 3 online participants but not the different people in the room.

r3dbars:
I'll get to work on it! Thanks for the feedback!

r3dbars:
I added this! Let me know how it works! It's using the same speaker diarization methods as the system audio.

r3dbars:
Implemented on April 15, 2026 via commit d388eb0 and merged in PR #329. Transcripted now supports identifying multiple local speakers on the mic channel behind the Identify multiple local speakers setting. The review flow also includes the People in the room section and a Keep as You fallback. Closing as shipped.

TeaPowered:
It's great! When it's identifying speakers after the recording, it would be great to select 'You' from the speakers menu

r3dbars:
That is a great idea! I will change that!

PR #329, opened 2026-04-15
feat: identify multiple local speakers on the mic channel (#312)

r3dbars:
## Summary

Closes the feature requested in **GitHub #312** ("Identify multiple local speakers"). Meetings with several people physically in the room + remote participants now split the local mic into individual speakers instead of collapsing everyone into a single "You."

**Default OFF** behind a new settings toggle. We'll flip the default after real-world DER on room recordings validates quality.

**Stacked on #327.** Base is `claude/happy-margulis` (the foundation PR). Merge that first, then this can retarget to `main`.

## What this does

| | |
|---|---|
| **New preference** | `LocalSpeakerPreferences.splitLocalSpeakersEnabled` (default `false`), toggled from Settings → People in the Room |
| **Pipeline** | When the flag is on, `transcribeMultichannel` runs PyAnnote offline diarization on the mic track too; mic utterances get per-speaker IDs + persistent UUIDs and go through the same DB-matching / ghost-speaker / adaptive-threshold path as system audio |
| **Runner** | Duplicates the system-speaker classification for mic; extracts clips from the mic WAV via the source-agnostic `SpeakerClipExtractor`; merges mic entries (tagged `.mic`) into the `SpeakerNamingRequest` |
| **Sheet UI** | Entries partition by channel. "People in the room" section renders above "Remote participants" when both are present. New **"Keep as You"** batch button next to the local header dims all mic rows and emits `.collapsedToMe` updates on save (toggles to "Split into speakers" to undo) |
| **Coordinator** | On `.collapsedToMe`: deletes profiles in `newlyCreatedMicProfileIds` (never touches previously-named profiles — Keep-as-You shouldn't destroy prior naming work), then calls new `TranscriptSaver.collapseMicSpeakersToYou` to rewrite `[Mic/X]` body labels to `[Mic/You]`, restore the single-speaker header, remove the Local Speaker Breakdown section, and strip mic entries from YAML |
| **Formatter** | Mic header becomes "Microphone (People in the Room)" when >1 local speaker; adds "Local Speaker Breakdown" analogous to the existing remote one; YAML speakers block gets per-entry `channel:` field; `speakerDbIds` is now channel-qualified to avoid `mic_0` / `system_0` collisions |
| **UX fix** | `SpeakerNamingRequest.knownPeople` is now seeded from named DB profiles. Previously the combobox was always empty — users typed blind. Benefits remote-speaker flow too |

## Architecture decisions called out in the code

- **No reverse contamination gate.** Codex's adversarial review (session `019d8ef5-79d8-7740-a870-7f858c006f17`) flagged symmetric weighting as a recall killer — normal in-room cross-talk would be silently dropped. The mic path trusts PyAnnote's own post-processing and relies on the user-driven "Keep as You" escape for bad cases.
- **No made-up "confidence gate."** My original plan had `max-pairwise-similarity > 0.85 means oversplit` — tautological against the existing 0.78 merge threshold in `EmbeddingClusterer.postProcess`. Dropped. The honest signal is `micSpeakerCount > 1`, period.
- **Classification logic inlined in a helper, not refactored.** `processMicChannelWithDiarization` is ~150 lines of near-duplicate code with the system block. Extracting a shared helper from 200+ lines of tightly-tuned speaker logic was deemed too risky for this PR. Can DRY up after the feature stabilizes.
- **Existing `applyNameReplacement` is used only for system.** Mic collapse uses a scoped regex rewrite (`[Mic/...]` → `[Mic/You]`) to avoid touching system labels or existing wiki links.

## Test plan

- [x] `bash build.sh` — clean
- [x] `bash run-tests.sh` — 289/289 pass (foundation renames + this change)
- [x] `swift test` — 53/53 pass (package boundary)
- [x] `bash run-integration-smoke.sh` — green
- [ ] Manual: real 2+ person room meeting with the flag on — verify sheet shows "People in the Room" section, clips playback is audible, naming one "Me" → next meeting auto-matches via cross-meeting DB matching
- [ ] Manual: same recording, click "Keep as You" — verify saved transcript reads "You" everywhere on the mic channel, YAML has no mic speaker entries, `speakers.sqlite` no longer contains the meeting's newly-created mic profiles

## Known limitations (documented for follow-up)

1. **Retroactive renames** (`RetroactiveSpeakerUpdater.retroactivelyUpdateSpeaker` from Settings) is still system-only for the "rename a speaker months later" flow. The initial-save path handles mic speakers correctly via `updateSpeakerNames` + `collapseMicSpeakersToYou`; historical transcripts won't update when a mic speaker is renamed in Settings. Follow-up PR will symmetrize.
2. **Cross-channel matching** (same person on mic in one meeting + system in another) is uncalibrated. WeSpeaker embeddings share a 256-dim space but matching thresholds (`0.85 / 0.78 / 0.70` in `TranscriptionPipeline.swift`) weren't tuned for the close-mic vs. far-field acoustic gap. Likely failure: duplicate profiles per channel. User can manually merge in Settings.
3. **Retries** from the failed-transcriptions queue always run with `splitLocalSpeakers=false` regardless of the original task's flag. Acceptable given retries are rare and the feature is default-off.
4. **First-meeting "me" UX tax.** User sees `Speaker 0 / 1 / 2` with no hint which is them until they play clips. Accepted one-time cost rather than building a voice-enrollment onboarding step. If dogfood shows it's confusing, that becomes the follow-up.

## Follow-ups (out of scope)

- Calibrate confidence gating if PyAnnote DER on MacBook-mic room audio is worse than expected.
- Symmetrize `RetroactiveSpeakerUpdater` for mic speakers.
- Lightweight voice-enrollment onboarding if the first-time UX proves rough.
- Tests for the mic naming path + `collapseMicSpeakersToYou` (none added in this PR — existing 289 suite covers regression on the default-off path).


r3dbars:
**Verdict**: REQUEST CHANGES

**Summary**: Solid architecture — `processMicChannelWithDiarization` mirrors the system path cleanly, the `.collapsedToMe` "Keep as You" flow is well thought through, the channel-qualified keying (`mic_X` / `system_X`) in `speakerDbIds` / `speakerMappings` / `speakerSources` is the right call, and the runner's Phase 1.6 mic classification block reads well. However, the per-mic-speaker *naming* path (user names mic speakers individually without clicking "Keep as You") is broken and will either crash or silently corrupt transcripts. The PR description's claim that "the initial-save path handles mic speakers correctly via `updateSpeakerNames` + `collapseMicSpeakersToYou`" isn't true.

- **Duplicate-key crash / wrong-entry rewrite in `updateSpeakerNames`.** `Sources/TranscriptedCore/Speaker/RetroactiveSpeakerUpdater.swift:259` builds `updatesBySystemKey` with `"system_\($0.diarizerSpeakerId)"` *regardless of channel*:
  ```swift
  let updatesBySystemKey = Dictionary(uniqueKeysWithValues: updates.map {
      ("system_\($0.diarizerSpeakerId)", ...)
  })
  ```
  When local split is on and the user names individual mic speakers (any path that isn't `.collapsedToMe`), `regularUpdates` contains mixed mic+system entries. PyAnnote numbers each channel from 0, so mic diarizer 0 + system diarizer 0 is the common case → `Dictionary(uniqueKeysWithValues:)` precondition-fails → **runtime trap**. Even without a collision, mic updates land in the dict keyed as `system_X`, so `rewriteLegacyTranscriptSection` / `rewriteStyledTranscriptSection` (which correctly look up `"mic_\(speakerId)"` at line 836/881) never find a match — **mic body labels `[Mic/Speaker 0]` stay as placeholders** and the user's typed name is silently lost in the body. Fix: key by `"\(update.channel.rawValue)_\(update.diarizerSpeakerId)"`.

- **`updateFrontmatterSpeakerMetadata` matches YAML entry without channel.** Lines 574–584 find the first `id: "X"` in the frontmatter. Now that YAML entries are channel-qualified (two entries can share `id: "0"` with different `channel:` values), naming mic_0 when system_0 exists will rewrite the wrong speaker's `db_id` / `name` / `source`. Fix: match on `id:` **and** `channel:` lines within the same block.

- **"Local Speaker Breakdown" never rewritten on individual mic naming.** `rewriteRemoteSpeakerBreakdown` is hardcoded to the remote section and only uses `system_` keys. The Local Speaker Breakdown added by `TranscriptFormatter` at save time keeps showing "Speaker 0 / Speaker 1" forever for named mic speakers. Either add a mic-breakdown rewriter or reuse the remote one parameterized by channel.

- **Stale comment + dead code in `.collapsedToMe` branch.** `SpeakerNamingCoordinator.swift:257-262` says *"Wiring lands in the follow-up PR that enables mic diarization end-to-end; for now, skip silently so … Default (off) behavior never produces this action."* This IS that follow-up PR. The comment is misleading and the branch returns a value that's never used (coordinator partitions collapsed updates out of `regularUpdates` before this runs). Either delete the case or update the comment to reflect that collapsed updates are handled upstream.

- **No tests for the new path.** PR description admits this. At minimum I'd want direct tests for `collapseMicSpeakersToYou` (regex + YAML stripping), `stripYAMLSpeakerEntries`, and `processMicChannelWithDiarization` classification branches (ghost remap, adaptive threshold by sample count, cross-cluster merge). The 289-test green run only proves the default-off path still works.

- **Retry resets `splitLocalSpeakers` to false.** Acknowledged in PR description and the comment at `TranscriptionTaskManager.swift:199-201`, but worth persisting the flag on `FailedTranscription` rather than silently dropping it — mic-split-enabled users will get unexpected single-"You" behavior after a retry, with no UI feedback that it happened.

- **Fragile Speakers-Detected scrubber in `collapseMicSpeakersToYou`.** The regex + "find next `\n\n### `" scoping relies on the section terminator being exactly that. If a future formatter change adds a trailing heading variant or an intermediate `####`, this silently no-ops. A one-line anchored replace keyed on the specific mic stat (e.g. capturing only within the mic stats block) with an explicit assert/log when not found would be more durable.

**Nits**:

- `LocalSpeakerPreferences.isEnabled()` has a redundant `object(forKey:)` guard — `bool(forKey:)` already returns `false` for unset keys. Drop the first check.
- `SpeakerPeopleSettingsSection` tracks `@State private var splitLocalSpeakersEnabled` but doesn't subscribe to `.localSpeakerPrefsDidChange`. Inconsequential today (no other toggle site), but the notification is wired up for a reason — if you keep it, the view should observe it for free consistency.
- `AppLogger.transcription.info("Post-processed mic speaker segments", ["diarizer": "\(rawSpeakerCount)", "after": "\(postProcessedSpeakerCount)", ...])` — key `"diarizer"` is ambiguous; `"beforePostProcess"` matches the system block's convention better.
- `processMicChannelWithDiarization` really is ~150 lines of near-duplicate code. PR description owns this as a deliberate non-refactor, which is fine for landing, but the "mic and system differ in one gate" structure is exactly the kind of thing that bit-rots asymmetrically. Worth a tracking issue.

**What looks good**:

- Channel-qualified keying (`mic_0` / `system_0`) in `speakerDbIds`, `speakerMappings`, and `speakerSources` — the right fix for collisions that this feature would otherwise have introduced everywhere.
- YAML `channel:` field per speaker entry in `TranscriptFormatter` — clean, diffable, and exactly what `stripYAMLSpeakerEntries` needs to work precisely.
- The "Keep as You" escape hatch plus the `newlyCreatedMicProfileIds` constraint on DB deletion (only delete profiles this meeting created, never prior named ones) is the right safety invariant. Glad that's called out explicitly in comments.
- Seeding `SpeakerNamingRequest.knownPeople` from named DB profiles — real fix for a pre-existing paper cut on the remote flow, called out honestly in the PR description.
- The quality gates (`≥0.3 quality`, `≥1.0s duration`) + ghost-speaker remap + adaptive threshold by embedding count are all brought over faithfully from the system path, which is what you want for symmetric behavior on the mic channel.
- Honest documentation of trade-offs (no contamination gate, no fake confidence gate, no retroactive symmetrization, cross-channel matching uncalibrated) in the PR description. That kind of forthrightness about deferred work is genuinely helpful for review.

---

CI hasn't reported any checks for this SHA (`get_check_runs` returned zero). Worth confirming `bash build.sh`, `bash run-tests.sh`, `bash run-integration-smoke.sh`, and `swift test` actually ran locally — the test-plan checkboxes suggest they did but nothing verifies that on the PR.

---

r3dbars:
Retargeted this branch onto `main` and folded in the mic speaker rewrite fixes from #335.

Addressed the blockers from review:
- keyed transcript rename state by `channel + diarizerSpeakerId`
- matched frontmatter speaker entries by both `id` and `channel`
- rewrote the local speaker breakdown for named mic speakers
- added regression coverage for mic rename and `Keep as You` collapse
- fixed the local breakdown rewrite boundary so it no longer truncates the rest of the transcript

Verification on the repaired branch:
- `bash build.sh`
- `bash run-tests.sh`
- `swift test`

PR #330, opened 2026-04-15
Fix review findings across capture flows

r3dbars:
## Summary
- fix analytics, dictation, and meeting-start flows called out in the review findings
- harden audio/session cleanup, speaker naming persistence, and system-audio permission handling
- add a launch smoke check to `build.sh` and update the affected tests

## Verification
- `bash build.sh`
- `bash run-tests.sh`
- `swift test`

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -30,6 +30,7 @@ final class MeetingCaptureBridge: ObservableObject {
     @Published private(set) var systemLevel: Float = 0         // system audio level (latest frame from Core's rolling history)
     @Published private(set) var recordingDuration: TimeInterval = 0
     @Published private(set) var systemAudioStatus: SystemAudioStatus = .unknown
+    @Published private(set) var errorMessage: String?
 
     // MARK: - Underlying capture
 
@@ -42,6 +43,7 @@ final class MeetingCaptureBridge: ObservableObject {
     /// Fulfilled when Core's Audio reports recording completed. Cleared each
     /// time `startRecording` runs so back-to-back sessions do not leak state.
     private var completionContinuation: CheckedContinuation<(micURL: URL?, systemURL: URL?), Never>?
+    private var startContinuation: CheckedContinuation<Bool, Never>?
 
     init(audio: Audio = Audio()) {
         self.audio = audio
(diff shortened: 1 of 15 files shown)


r3dbars:
## Code review

**Verdict:** APPROVE (with a couple of minor should-fix nits)

**Summary:** This PR tightens a handful of capture-flow reliability gaps: meeting start now reports success/failure so the prompt-accept path only marks prompts as accepted when recording actually begins; speaker naming mutations are planned up front and only applied after the transcript rewrite succeeds (no more partial DB commits on rewrite failure); the system-audio permission state grows a proper `unknown` bucket instead of conflating "never checked" with "denied"; and tear-down paths (`ParakeetEngine`, `systemAudioFileURL`, naming-sheet close) get cleaner. Tests and a launch smoke in `build.sh` back the changes.

### Blockers
None.

### Should fix

- **`Sources/Meeting/MeetingSessionController.swift:281-288`** — stray indentation in the `.recording` case: the `DiagnosticsTrail.record(...)` block is over-indented relative to the `return true` below it. Not a correctness issue but it'll show up as noise next time someone re-formats the file.
- **`Sources/Meeting/MeetingCaptureBridge.swift:78`** — the 5-second start timeout is a raw literal. Everywhere else in this repo, audio timing lives in `TranscriptedConstants` (e.g. `audioWatchdogTimeout`). Consider hoisting this value there so it's discoverable and easy to tune alongside the sibling watchdog constants.
- **`Sources/UI/Overlay/DictationSessionController.swift:450-459`** — small duplication: `AppSoundPlayer.shared.play(.dictationDelivered)` fires in both branches of the `.pasted` case. Factoring the sound call out above the `if let saveFailureMessage` is marginally cleaner.

### Nits

- **`Sources/TranscriptedCore/Speaker/SpeakerDatabase.swift:283`** — `let newId = existingId ?? UUID()` quietly changes the semantics of `addOrUpdateSpeaker(embedding:existingId:)`: previously, a missing `existingId` match always minted a fresh UUID; now the caller's UUID is honored. This is correct for the new coordinator path (which pre-generates the UUID), and the other call-sites (`TranscriptionPipeline.swift:251,272,283` and the test at `DatabaseFilePermissionsTests.swift:31`) all either pass `nil` or pass an ID that already exists, so they're unaffected. Worth a one-line doc comment on the method to avoid future callers tripping on it.
- **`Sources/Meeting/MeetingCaptureBridge.swift:183-191`** — you now have both `audio.$isRecording.receive(on: RunLoop.main).assign(to: &$isRecording)` and a second `.sink` subscription on the same publisher. Works fine, but two subscriptions to the same `@Published` is a smell; you could piggyback the continuation-resume logic onto a single subscription if you prefer. Optional.
- **`scripts/entrypoints/build.sh:84`** — the smoke runs the binary directly and kills it after 5s. For a menubar app that finishes init quickly this is fine, but any slow model preload could race the check. If that ever bites, consider grepping `smoke_log` for a "launched" marker instead of the pure liveness test.

### What looks good

- **Atomicity in `SpeakerNamingCoordinator`** is the highlight. Separating `PlannedNamingChanges` from application and only mutating the DB after `TranscriptSaver.updateSpeakerNames` succeeds is exactly the right shape — no more half-applied renames on rewrite failure. The updated test (`rejectedProfile?.disputeCount` unchanged on failure) locks in the invariant.
- **`NamingWindowController.didComplete`** guard neatly handles the "user closes window without save/cancel" path — prior code could have invoked `onComplete` twice.
- **Returning `Bool` from `MeetingCaptureBridge.startRecording`** and propagating it through `MeetingSessionController.startRecording(trigger:)` up to `TranscriptedAppDelegate` so `meetingPromptDetector.markAccepted(...)` only fires on real starts is a clean fix.
- **`SystemAudioPermissionState` (granted/denied/unknown)** is a much better model of the cached-default reality than the boolean it replaces, and the test additions exercise all three paths.
- **`systemAudioFileURL` cleanup on capture failure** (nil-out, remove partial file, clear `systemAudioFile`) closes a real foot-gun where failed setup left a dangling URL pointing at a zero-byte file.
- **`pendingSamples` tear-down parity** across `mediaServicesReset`, `stopRecording`, `cancel`, and the watchdog is the kind of unglamorous cleanup that prevents weird stale-state bugs after wake/reset.
- **`cancelAudioWatchdog()` helper** — small but good de-duplication.
- **`AnalyticsPayloadSanitizer` on `session_id`** is a reasonable defense-in-depth; dropping the key when sanitization empties it avoids sending garbage.

Overall a solid, focused fix PR.

---

PR #331, opened 2026-04-15
Safe quick-wins from deep code review

r3dbars:
## Summary

Low-risk fixes distilled from a 20-agent deep review of `main`. Each change was verified in source before editing; agent claims that didn't hold up on inspection were dropped rather than landed blindly.

### Commit 1 — `fix: safe quick-wins from deep code review`

- **Menubar icon template flag.** `Sources/TranscriptedApp.swift` — mark the status-item `NSImage` with `isTemplate = true` so it tints correctly under light-mode menu bars and dynamic appearance changes.
- **Bearer sanitizer tightened.** `Sources/Observability/AnalyticsPayloadSanitizer.swift` — regex now matches `Bearer\s+` and is case-insensitive, so tabs, multi-space, and lowercase `bearer` all get redacted.
- **Hotkey refs nilled on failure.** `Sources/Capture/ContextCaptureEngine.swift` — Carbon leaves the out-pointer undefined when `RegisterEventHotKey` fails; we now nil `hotkeyRef` / `meetingHotkeyRef` on that path so a later unregister can't dereference an undefined pointer.
- **Dead draft-mode stubs removed.** `Sources/UI/Overlay/DictationSessionController.swift` — `startSession(imageData:sourceApp:)` and `stopSessionAndDraft()` had zero callers. `cancelSession()` kept because `ContextCaptureEngine` still invokes it on interrupt paths.

### Commit 2 — `fix: scrub feedback logs and reject secure text fields`

- **Feedback log scrubbing.** `Sources/UI/MenuBar/MenuBarSettingsView.swift` — the "Send Feedback" button embeds the last 80 log entries into a GitHub issue URL. Those lines can carry `/Users/*` paths, emails, or tokens that happened to land in a log. Logs now pass through `AnalyticsPayloadSanitizer.sanitizeText` before the URL is built.
- **Secure-text-field rejection.** `Sources/Accessibility/AccessibilityBridge.swift` — `focusedTextElement` accepts anything whose role contains `"Text"`, which also matches `AXSecureTextField`. Explicit `role == "AXSecureTextField"` guard added so the AX bridge can't surface password / PIN targets to callers.

## Not included (agent claims that didn't hold up)

- MeetingCaptureBridge continuation swap — theoretical, requires Audio callback delay between two `@MainActor` sequences currently serialized by MSC.
- `ModelDownloadService.checkNetworkReachability` double-resume — both path handler and timeout dispatch on the same serial queue; `ResumeGuard` is fine.
- Meeting trigger attribution logged as `.unknown` — `let recordingTrigger = activeRecordingTrigger` captures locally *before* the reset, so the subsequent telemetry uses the correct value.
- `stopMonitoring` missing `engine.isRunning` guard — already present.

## Still outstanding (needs more than a quick-win)

- Paste `CGEvent.post` verification (needs AX-observe or shorter clipboard-restore heuristic)
- Pipeline task-completion refactor — defer `handleTaskCompletion` until `SpeakerNamingCoordinator` resolves
- Three-observer wake-notification race — needs a centralized wake coordinator
- `AudioDeviceRecovery` tap-reinstall guarantee on early-return — real but needs integration tests
- `Info.plist` `1.0.8` vs appcast latest `1.0.7` — release-owner decision
- `SQLite synchronous=NORMAL` / missing `fsync` after atomic transcript writes
- Analytics API key in POST body, no TLS pinning

## Test plan

- [x] `bash build.sh` — clean build
- [x] `bash run-tests.sh` — 289 tests, 289 passed
- [ ] Manual: confirm menubar icon renders correctly in light-mode menu bar
- [ ] Manual: confirm "Send Feedback" still opens a pre-filled GitHub issue
- [ ] Manual: confirm dictation still pastes to normal text fields
- [ ] Manual: confirm dictation refuses to target a password field (if reachable)

### DIFF
--- a/Sources/Accessibility/AccessibilityBridge.swift
+++ b/Sources/Accessibility/AccessibilityBridge.swift
@@ -24,6 +24,11 @@ struct AccessibilityBridge {
         var roleRef: AnyObject?
         AXUIElementCopyAttributeValue(axElement, kAXRoleAttribute as CFString, &roleRef)
         let role = roleRef as? String ?? ""
+
+        // Never expose secure text fields (password, PIN, biometric auth). The
+        // broader "contains Text" match below would otherwise accept AXSecureTextField.
+        if role == "AXSecureTextField" { return nil }
+
         let textRoles = ["AXTextArea", "AXTextField", "AXWebArea", "AXTextView", "AXComboBox"]
         guard textRoles.contains(role) || role.contains("Text") else { return nil }
 
--- a/Sources/Accessibility/AccessibilityBridge.swift
+++ b/Sources/Accessibility/AccessibilityBridge.swift
@@ -25,9 +25,13 @@ struct AccessibilityBridge {
         AXUIElementCopyAttributeValue(axElement, kAXRoleAttribute as CFString, &roleRef)
         let role = roleRef as? String ?? ""
 
-        // Never expose secure text fields (password, PIN, biometric auth). The
-        // broader "contains Text" match below would otherwise accept AXSecureTextField.
+        // Never expose secure text fields (password, PIN, biometric auth). Some
+        // AppKit hosts surface secure fields as AXTextField with subrole
+        // AXSecureTextField, so we check both the role and the subrole.
         if role == "AXSecureTextField" { return nil }
+        var subroleRef: AnyObject?
+        AXUIElementCopyAttributeValue(axElement, kAXSubroleAttribute as CFString, &subroleRef)
+        if let subrole = subroleRef as? String, subrole == "AXSecureTextField" { return nil }
 
         let textRoles = ["AXTextArea", "AXTextField", "AXWebArea", "AXTextView", "AXComboBox"]
         guard textRoles.contains(role) || role.contains("Text") else { return nil }
--- a/Sources/Capture/ContextCaptureEngine.swift
+++ b/Sources/Capture/ContextCaptureEngine.swift
@@ -348,6 +348,9 @@ class ContextCaptureEngine: ObservableObject {
             &hotkeyRef
         )
         if dictationStatus != noErr {
+            // Carbon leaves the ref undefined on failure — nil it so unregister
+            // doesn't later call UnregisterEventHotKey on a bogus pointer.
+            hotkeyRef = nil
             errors.append("Dictation shortcut")
             EventReporter.shared.capture(level: .error, engine: "capture", event: "hotkey_register_failed",
                 message: "Dictation hotkey registration failed", context: ["os_status": "\(dictationStatus)"])
@@ -364,6 +367,7 @@ class ContextCaptureEngine: ObservableObject {
             &meetingHotkeyRef
         )
         if meetingStatus != noErr {
+            meetingHotkeyRef = nil
             errors.append("Meeting shortcut")
             EventReporter.shared.capture(level: .error, engine: "capture", event: "hotkey_register_failed",
                 message: "Meeting hotkey registration failed", context: ["os_status": "\(meetingStatus)"])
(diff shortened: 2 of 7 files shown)


r3dbars:
**Verdict**: REQUEST CHANGES

**Summary**: Six small, well-scoped fixes derived from a deep review. Five of them are clean wins — the menubar template flag, Bearer regex tightening, Carbon out-pointer nil-on-failure, dead-stub removal, and the `AXSecureTextField` guard all check out against the surrounding code. The sixth — routing feedback logs through `AnalyticsPayloadSanitizer.sanitizeText` — is well-intentioned but silently breaks the "Send Feedback" flow.

- `Sources/UI/MenuBar/MenuBarSettingsView.swift:102` — `AnalyticsPayloadSanitizer.sanitizeText` truncates its output to `maxValueLength = 80` characters plus `"..."` (see `AnalyticsPayloadSanitizer.swift:4,75-77`). That sanitizer is designed for single analytics property values, not for an 80-line log blob. After this change, the GitHub issue body `Logs:` section will contain only the first ~80 characters of the joined 80-line suffix — i.e. almost always just the first line, truncated. This defeats the feature. Two reasonable fixes:
  1. Sanitize line-by-line: `rawLogLines.split(separator: "\n").map(AnalyticsPayloadSanitizer.sanitizeText).joined(separator: "\n")`. Caveat: each line still gets capped at 80 chars, which may still be too aggressive for log lines but is strictly better than capping the whole blob.
  2. Extract the redaction pipeline into a separate `redact(_:)` helper on the sanitizer that applies the regex substitutions without the length cap, and have `sanitizeText` call that plus truncate. Use the uncapped helper here.
  Option 2 is cleaner — it gives the feedback path a real "scrub but don't truncate" primitive and keeps the analytics cap intact. Either way, please add a unit test that sends a multi-line blob with secrets through the feedback path and verifies both redaction and preservation of length.

- `Sources/Accessibility/AccessibilityBridge.swift:30` — the explicit `role == "AXSecureTextField"` check is good, but secure fields can also surface as `AXTextField` with subrole `AXSecureTextField` in some AppKit hosts. Worth reading `kAXSubroleAttribute` and also rejecting when the subrole is `AXSecureTextField`, so the broader `role.contains("Text")` match doesn't slip a subrole-tagged secure field through. Minor belt-and-suspenders.
- Consider adding a regression test for the sanitizer's new `Bearer\s+` + case-insensitive behavior (`bearer\ttoken`, `BEARER   abc.def_ghi`, etc.). The change is one line but the intent is explicit and cheap to lock down.

**Nits** (take or leave):

- `ContextCaptureEngine.swift:352-353` — the comment is clear but long; `// Carbon leaves the out-pointer undefined on failure` would be enough given the adjacent nil assignment is self-documenting.
- `DictationSessionController.swift:127-128` — the explanatory comment for removed stubs is helpful for now, but once the draft-mode transition is fully closed out, it's the kind of archaeological note that can go away. Not a blocker.
- The PR's "Not included" and "Still outstanding" sections are genuinely useful reviewer context — keep doing this on quick-win sweeps.

**What looks good**:

- The Carbon `RegisterEventHotKey` fix is correct and matches the existing unregister idiom (`if let ref = hotkeyRef` at lines 307, 431, 441 — nilling on failure is the right invariant).
- The `Bearer\s+` + case-insensitive change is a clean, small widening of a known-brittle regex.
- The `isTemplate = true` menubar icon fix is exactly right for SF Symbol status items.
- The `AXSecureTextField` guard addresses a real paste-target-leak risk and is placed in the right spot (before the permissive `role.contains("Text")` fallthrough).
- The dead-stub removal is properly cross-checked — `startSession(imageData:sourceApp:)` / `stopSessionAndDraft()` really do only appear in `docs/archive/`, and `cancelSession()` is correctly preserved.
- Commits are well-split and described. The PR body is unusually honest about what was dropped and why, which makes review much easier.

---

PR #334, opened 2026-04-16

r3dbars:
### Audio capture lifecycle
- cleaned up abandoned system-audio files when capture startup races against a stop or session swap, so failed/stale starts do not leave bogus scratch artifacts behind
- made both system-audio capture engines fully reset their prepared state and counters when `stop()` is called before capture is marked active, which avoids stale session state leaking into the next run

### Transcript health metadata
- fixed recording-health classification so failed system-audio capture is no longer reported as healthy/excellent
- folded recording gaps and device switches back into the capture-quality rating, so transcript metadata better reflects degraded real-world recordings

### Tests
- added focused tests covering failed system-audio health classification and quality downgrades for gaps/device switches

## Verification
- `xcodebuild -project Transcripted.xcodeproj -scheme Transcripted -configuration Debug build 2>&1` could not be used as requested because this checkout does not contain `Transcripted.xcodeproj`
- verified with the repo's supported build lane instead: `bash build.sh`

### DIFF
--- a/Sources/TranscriptedCore/Audio/AudioFileManager.swift
+++ b/Sources/TranscriptedCore/Audio/AudioFileManager.swift
@@ -38,6 +38,7 @@ extension Audio {
             try? FileManager.default.createDirectory(at: captureDir, withIntermediateDirectories: true)
             let timestamp = DateFormattingHelper.formatFilenamePrecise(Date())
             let fileURL = captureDir.appendingPathComponent("meeting_\(timestamp)_system.wav")
+            let sessionGeneration = recordingSessionGeneration
             AppLogger.audioSystem.info("System audio file URL", ["file": fileURL.lastPathComponent])
 
             DispatchQueue.global(qos: .userInitiated).async { [weak self] in
@@ -46,12 +47,34 @@ extension Audio {
                     return
                 }
 
+                func cleanupAbandonedSetup() {
+                    strongSelf.systemAudioFileQueue.sync {
+                        strongSelf.systemAudioFile = nil
+                    }
+                    capture.stopSync()
+                    try? FileManager.default.removeItem(at: fileURL)
+                }
+
+                func sessionIsCurrent() -> Bool {
+                    sessionGeneration == strongSelf.recordingSessionGeneration
+                }
+
                 AppLogger.audioSystem.info("Starting system audio capture on background thread")
 
                 do {
+                    guard sessionIsCurrent() else {
+                        cleanupAbandonedSetup()
+                        return
+                    }
+
                     // Step 1: Prepare the tap (creates aggregate device, gets format)
                     // This does NOT start the I/O proc yet
                     try capture.prepare()
+
+                    guard sessionIsCurrent() else {
+                        cleanupAbandonedSetup()
+                        return
+                    }
 
                     // Step 2: Get the format from the tap (now corrected to match device nominal rate)
                     guard let tapFormat = capture.audioFormat else {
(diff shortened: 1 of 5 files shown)


r3dbars:
## Review

**Verdict**: APPROVE

**Summary**: Cleans up three real lifecycle/reporting bugs: orphaned system-audio scratch files when a session swap races with capture setup, stale `isPrepared`/stats state leaking across runs, and recording-health metadata over-reporting quality. The session-generation guards, `cleanupAbandonedSetup()` helper, and `resetStats()` on teardown are targeted and consistent with the existing generation-counter pattern in `AudioDeviceRecovery.swift`.

**Blockers**: none.

**Should fix**:

- `Sources/TranscriptedCore/Audio/SCKAudioCapture.swift:113` — `errorMessage = nil` is now assigned on the userInitiated background queue (via `AudioFileManager.startAudioCapture`). It's a `@Published` property and the existing error set on line 148 is explicitly `DispatchQueue.main.async`'d for a reason. This is consistent with `SystemAudioCapture.swift:194` which also does a cross-thread direct assign, so not a regression, but while you're in this area it'd be worth wrapping both in `DispatchQueue.main.async` to keep `objectWillChange` on the main runloop.

**Nits**:

- `Sources/TranscriptedCore/Audio/AudioFileManager.swift:114-116` — the buffer callback now reads `self.recordingSessionGeneration` which takes an `NSLock`. For the SCK path this runs on the `SCKAudioCapture.output` userInitiated queue (fine), but the legacy CoreAudio path's `AudioDeviceIOProc` is real-time. The rest of this callback already hits several locks (`systemBufferCount`, level calc, etc.), so this is in keeping with the existing style, but if you ever want to tighten RT-safety, a plain `UInt64` with `OSAtomic`/`os_unfair_lock` for the generation counter would avoid the Obj-C lock.
- `Tests/TranscriptedCoreTests/TranscriptMetadataBuilderTests.swift:31` — the test name says "for gaps and device switches" but only exercises both together. Splitting into "gap only" and "device-switch only" cases would pin the `||` branch in `TranscriptMetadataBuilder.swift:48` more precisely (cheap follow-up, not worth blocking).
- `Sources/TranscriptedCore/Models/TranscriptMetadataBuilder.swift:37-42` — minor readability: the IIFE for `successRate` is fine but a plain `let failed = audio.systemAudioFailed || audio.systemAudioStatus == .failed` + ternary would be a touch easier to read. Take or leave.

**What looks good**:

- `cleanupAbandonedSetup()` is scoped inside the setup closure so it captures `strongSelf`, `capture`, and `fileURL` cleanly — no retain cycles, no leaked scratch `.wav`. The guard placement before `prepare()`, after `prepare()`, after `AVAudioFile` creation, and after `capture.start(...)` covers every await-equivalent checkpoint correctly.
- Moving `addStreamOutput` inside the `do/catch` in `SCKAudioCapture.start` and calling `cleanup()` on failure matches the behavior `SystemAudioCapture.start` already had after this PR — consistent engine lifecycle across both backends.
- Handling "prepared but not capturing" in both engines' `stop()`/`stopSync()` closes a real leak where `prepare()` succeeded but `start()` threw and the tap/stream was never torn down.
- `resetStats()` on cleanup prevents `bufferSuccessRate` from carrying results from a prior failed run into the next `RecordingHealthInfo` computation — which is exactly what the metadata fix on the other side needs.
- `RecordingHealthInfo.from` now honors `systemAudioFailed`/`.failed` status, and the downgrade ladder for device switches/gaps is well-bounded (switches >= 3 clamps to `.degraded`, otherwise at most one tier down, and `.degraded` can't be downgraded further — clean).
- Tests exercise the two observable behavior changes directly and use `@testable import` correctly to reach internal `AudioGap`/`appendRecordingGap`.

Nice, focused cleanup — ship it.

---

PR #335, opened 2026-04-16
[nightly-review] Preserve mic speaker name rewrites

r3dbars:
## Finding

Speaker naming updates were being rewritten with a hard-coded `system_*` lookup key even after the diarization refactor added per-channel speaker metadata. That meant a mic-channel rename could update YAML metadata but leave `[Mic/Speaker X]` lines unchanged once local speaker naming is enabled.

## Fix

- preserve `SpeakerNameUpdate.channel` through `handleNamingComplete`
- key clip lookup and transcript rewrite state by `channel + diarizerSpeakerId`
- add an end-to-end regression test that renames a mic speaker and verifies the transcript body becomes `[Mic/<name>]`

## Verification

- `bash build.sh`
- `bash run-tests.sh`
- `swift test`

### DIFF
--- a/Sources/TranscriptedCore/Speaker/RetroactiveSpeakerUpdater.swift
+++ b/Sources/TranscriptedCore/Speaker/RetroactiveSpeakerUpdater.swift
@@ -256,9 +256,9 @@ extension TranscriptSaver {
             }
 
             let obsidianEnabled = isObsidianFormatted(content)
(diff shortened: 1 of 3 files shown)


r3dbars:
Closing this as folded into #329.

The channel-aware mic speaker rewrite, the frontmatter channel match, the local speaker breakdown rewrite, and the regression coverage all landed as part of the repaired `#329` branch, which is now merged into `main`.

PR #339, opened 2026-04-16

r3dbars:
## Summary

- **`SentryPayloadSanitizer`**: Mirror the `bearerRegex` fix from PR #331 — update to `Bearer\s+` with `.caseInsensitive` so mixed-case variants and multi-whitespace separators are redacted. The old pattern only matched literal `Bearer <token>` with a single space, leaving `BEARER  token`, `bearer\ttoken`, etc. unredacted in Sentry payloads.
- **`AccessibilityBridge`**: Extract `secureTextFieldRole` string constant (was duplicated across two guards), promote `textRoles` to a `static let Set<String>` (O(1) lookup, single allocation vs. per-call array), and reorder the subrole IPC call so it only fires for elements that already passed the text-role check — avoids a cross-process AX call for non-text focused elements (buttons, sliders, etc.).
- **`AnalyticsPayloadSanitizer` / `SentryPayloadSanitizer`**: Remove unnecessary `.sorted()` on all sanitize-properties iterations. The result is written into an unordered `[String: String]`/`[String: Any]` dictionary, so the O(n log n) sort produced no observable effect and just wasted cycles.
- **`MenuBarSettingsView`**: Remove a redundant inline comment that restated the `redact` doc-comment added in PR #331.

## Test plan

- [x] `bash build.sh` — clean build, signed app
- [x] `bash run-tests.sh` — 315/315 passed
- [x] `bash run-integration-smoke.sh` — all smoke checks passed


r3dbars:
**Verdict**: APPROVE

**Summary**: Tight nightly simplification sweep — four unrelated but independently-justified micro-cleanups in observability, accessibility, and docs. Each change has a clear rationale in the PR description, and I verified the diffs do exactly what they claim. No behavior changes beyond the stated intent.

**Blockers**: None.

**Should fix**:
- `SentryPayloadSanitizerTests.swift` only exercises `"Bearer abc123"` (single-space, capitalized). PR #331 added mirrored case-insensitive / multi-whitespace coverage for the analytics side (`Tests/AnalyticsPayloadSanitizerTests.swift:67-74`), but no equivalent was added here when the same regex fix landed. Since the regex is identical to the already-tested analytics one, it's not strictly a gap — but the PR's stated intent is to "mirror the `bearerRegex` fix from PR #331," and the analytics mirror included a regression test. Adding the same four-case loop (`"bearer\ttokenabc"`, `"BEARER   abc.def_ghi"`, etc.) against `SentryPayloadSanitizer.sanitizeText` would make the mirroring complete.

**Nits**:
- `AccessibilityBridge.swift:34` — the behavior-preservation here is subtle (reordering the subrole check to run only after the role-filter passes). I traced every `role × subrole` combination and it's equivalent, but a one-line comment noting "role-filter is a superset of subrole-only rejects, so moving it up is safe" would save the next reader the same walk.
- The removed inline comment in `MenuBarSettingsView.swift:99` was actually useful as a *callsite* hint about why `redact` is used instead of `sanitizeText` (length cap). The doc-comment on `redact` itself captures the same info, so removal is defensible, but I'd call it a coin-flip rather than clearly redundant.

**What looks good**:
- The subrole IPC reorder in `AccessibilityBridge` is a real win, not just cosmetic — `AXUIElementCopyAttributeValue` is a cross-process hop, and `focusedTextElement` runs on every focus change. Skipping the second IPC for buttons/sliders/etc. is exactly the kind of micro-opt that adds up on hot paths.
- Promoting `textRoles` to `static let Set<String>` is a legit O(1) vs O(n) improvement with a single allocation — low-risk, high-obviousness refactor.
- The `.sorted()` removal is correctly justified: output is written into `[String: Any]` / `[String: String]`, so the sort had zero observable effect and was pure waste. Five spots cleaned uniformly.
- `Bearer\s+` + `.caseInsensitive` is the right fix and the rationale ("`BEARER  token`, `bearer\ttoken` previously slipped through") matches the analytics test cases from PR #331.
- Good discipline on commit hygiene: two focused commits (docs sync + code sweep), each change justified in the PR body, build + tests + smoke reported as green.

Clean PR. Ship it.

---

PR #340, opened 2026-04-16

r3dbars:
## Summary

Automated nightly security review — 5 findings fixed, 0 CRITICAL or HIGH.

### MEDIUM (3)

**`isStarting` race in `Audio.swift`**
- `isStarting` was a plain `Bool` read and written from both the main thread and background queues (e.g. the `AVCaptureDevice` permission callback). A narrow race between two concurrent `start()` calls could let both pass the `!isStarting` guard and set up two concurrent recording sessions. Fixed by adding an `NSLock`-backed computed property matching the `isMicRecovering` pattern already used in this class.

**`isDatabaseOpen` unsynchronized init in `SpeakerDatabase` and `StatsDatabase`**
- Both database classes set `isDatabaseOpen = true` inside `configureOpenDatabase()`, called from `init()` outside the serial queue. Public methods check `isDatabaseOpen` inside `queue.sync`/`queue.async` blocks. There was no memory barrier between the non-queue init path and the queue-protected reads. Fixed by wrapping `openDatabase()` + `createTables()` in `queue.sync` in both `init` variants.

**Unvalidated `modelId` interpolated into download URLs in `ModelDownloadService.swift`**
- `modelId` was string-interpolated directly into HuggingFace API and CDN URLs without any validation. A value containing `..` or extra slashes could cause the request to hit an unintended endpoint on the mirror. Fixed by adding an `isSafeModelId` allowlist validator (alphanumerics + `-_.`, max 2 slash-separated components, no `..`) applied at both URL construction sites in `fetchModelFileList` and `downloadFileWithMirrorFallback`.

### LOW (2)

**`BetaConfig` placeholder could ship unreplaced**
- `BetaConfig.userToken` defaulted to the literal string `"BETA_TOKEN_PLACEHOLDER"`. If `build-beta.sh` forgot the substitution step, the literal would be sent as a credential. Fixed by adding a `precondition(!token.hasSuffix("PLACEHOLDER"), ...)` that crashes the beta build at startup rather than silently shipping with a broken credential.

**`validateSavePath` did not block `~/Library`**
- The forbidden-prefix list blocked `/Library` (system) but not `~/Library` (user). A tampered `UserDefaults` value could direct transcript writes into `~/Library/Preferences` or `~/Library/Keychains`. Fixed by appending `NSHomeDirectory() + "/Library"` to the forbidden list at runtime.

## Not Fixed (informational only)

- **Sentry DSN / PostHog key in `Info.plist`**: Standard for client analytics SDKs in macOS apps; no code change needed. Mitigate via rate-limiting in the Sentry/PostHog project settings.
- **Sparkle appcast URL on mutable `main` branch**: EdDSA signature requirement is in place, so full RCE would require the private key. Consider moving to a versioned CDN path in a follow-up.

## Test Plan

- [x] `bash build.sh` — passes, signed successfully
- [x] `bash run-tests.sh` — 315 tests, 0 failures
- [x] `bash run-integration-smoke.sh` — all smoke checks pass

### DIFF
--- a/Sources/Beta/BetaConfig.swift
+++ b/Sources/Beta/BetaConfig.swift
@@ -9,7 +9,17 @@ import Foundation
 enum BetaConfig {
     /// Per-user beta token, baked in at build time.
     /// build-beta.sh replaces BETA_TOKEN_PLACEHOLDER with the actual token.
-    static let userToken = "BETA_TOKEN_PLACEHOLDER"
+    static let userToken: String = {
+        let token = "BETA_TOKEN_PLACEHOLDER"
+        // Security: fail fast if build-beta.sh forgot to substitute the token.
+        // Shipping the literal placeholder would allow unauthenticated access to beta
+        // endpoints under the BETA_TOKEN_PLACEHOLDER identity.
+        precondition(
+            !token.hasSuffix("PLACEHOLDER"),
+            "BetaConfig.userToken was not replaced by build-beta.sh — do not ship this build"
+        )
+        return token
+    }()
 
     /// Proxy base URL for beta-only proxy traffic.
     static let proxyBaseURL = "https://draft-proxy.tz427gsydr.workers.dev"
--- a/Sources/TranscriptedCore/Audio/Audio.swift
+++ b/Sources/TranscriptedCore/Audio/Audio.swift
@@ -43,7 +43,16 @@ public enum SystemAudioStatus: Equatable {
 public class Audio: ObservableObject {
     @Published public var isRecording: Bool = false
     @Published public private(set) var isMonitoring: Bool = false  // Lightweight level metering without file recording
-    private var isStarting: Bool = false  // Prevents double-start during async setup
+    // Security: isStarting is read and written from both the main thread and
+    // background queues (e.g. the AVCaptureDevice permission callback). Protect it
+    // with a lock to prevent a narrow race where two concurrent start() calls both
+    // pass the guard and set up two recording sessions.
+    private var _isStarting: Bool = false
+    private let isStartingLock = NSLock()
+    private var isStarting: Bool {
+        get { isStartingLock.lock(); defer { isStartingLock.unlock() }; return _isStarting }
+        set { isStartingLock.lock(); defer { isStartingLock.unlock() }; _isStarting = newValue }
+    }
     @Published public var audioLevel: Float = 0.0
     @Published public var recordingDuration: TimeInterval = 0.0
     @Published public var audioLevelHistory: [Float] = Array(repeating: 0.0, count: 15)
--- a/Sources/TranscriptedCore/Services/ModelDownloadService.swift
+++ b/Sources/TranscriptedCore/Services/ModelDownloadService.swift
@@ -222,6 +222,30 @@ public enum ModelDownloadService {
         return true
     }
 
+    /// Validate a HuggingFace model ID before interpolating it into download URLs.
+    /// Security: modelId is string-interpolated directly into HTTPS URLs. An attacker-controlled
+    /// or misconfigured value containing path traversal sequences (e.g. "../../foo") or
+    /// URL-breaking characters could cause the request to hit an unexpected endpoint.
+    /// Allowlist: "namespace/model-name" — alphanumerics, hyphens, underscores, dots, and
+    /// exactly one optional slash separating namespace from model name.
+    static func isSafeModelId(_ modelId: String) -> Bool {
+        guard !modelId.isEmpty, modelId.count <= 200 else { return false }
+        // Must not contain ".." traversal components
+        let components = modelId.components(separatedBy: "/")
+        guard !components.contains(".."), !components.contains(".") else { return false }
+        // At most two path components (namespace/model) — extra slashes are suspicious
+        guard components.count <= 2 else { return false }
+        // Each component must be non-empty and consist only of safe characters
+        let safeChars = CharacterSet.alphanumerics.union(CharacterSet(charactersIn: "-_."))
+        for component in components {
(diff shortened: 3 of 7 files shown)


r3dbars:
## Review

**Verdict**: APPROVE with one should-fix

**Summary**: Nightly security audit sweep — 3 MEDIUM and 2 LOW findings addressed across `Audio`, `SpeakerDatabase`, `StatsDatabase`, `ModelDownloadService`, `RecordingValidator`, and `BetaConfig`, plus a docs-only sync of `Sources/UI/CLAUDE.md`. Most of the changes are genuinely useful defense-in-depth. One fix (the `isStarting` lock in `Audio.swift`) doesn't actually close the race it describes, though in practice the race isn't currently reachable.

### Blockers
None.

### Should fix

- **`Audio.swift:46-55` — the `isStarting` lock fix is a TOCTOU shim, not a real lock.** The computed property only protects individual reads and writes. The actual check-and-set sequence at `Audio.swift:427` / `Audio.swift:447`
  ```swift
  guard !isRecording, !isStarting else { ... return }
  ...
  isStarting = true
  ```
  releases the lock between the guard and the assignment. Two concurrent `start()` calls on different threads would still both see `false`, both pass the guard, and both set `true`. Either:
  1. Make the check-and-set atomic:
     ```swift
     private func tryBeginStart() -> Bool {
         isStartingLock.lock(); defer { isStartingLock.unlock() }
         guard !_isStarting else { return false }
         _isStarting = true
         return true
     }
     ```
     then use `guard tryBeginStart() else { ... return }` in both `start()` and `startAudioCaptureAsync()`.
  2. Or delete the lock entirely and document the main-actor invariant. Today the only in-tree caller is `MeetingCaptureBridge.startRecording()` at `MeetingCaptureBridge.swift:75`, which is `@MainActor`, and every `isStarting = false` write is wrapped in `DispatchQueue.main.async` / `MainActor.run`. The stated race ("main thread and background queues (e.g. the AVCaptureDevice permission callback)") doesn't actually happen — the permission callback re-dispatches to main before touching `isStarting`.

  As it stands the lock costs runtime cycles for a race that a) doesn't exist in the current call graph and b) the lock wouldn't fix if it did. Note the same pattern applies to the pre-existing `isMicRecovering` property, so this is consistent with surrounding code — but that doesn't mean it's correct.

### Nits

- **`BetaConfig.swift:18`** — `!token.hasSuffix("PLACEHOLDER")` is slightly loose. Comparing against the full literal (`token != "BETA_TOKEN_PLACEHOLDER"`) would be stricter and avoid a theoretical false-positive crash if an actual token ever ended in `PLACEHOLDER`. Probably never matters in practice.
- **`BetaConfig.swift:12-22`** — the `precondition` only fires on first access to `BetaConfig.userToken`, not at app launch. A beta build that boots but never exercises the token-consuming code path (unlikely but possible in a crash-before-first-API-call scenario) could in principle ship without tripping it. Adding a no-op `_ = BetaConfig.userToken` near startup would make the fail-fast more reliable. Very minor.
- **`ModelDownloadService.swift:235`** — the explicit `components.contains("..")` / `components.contains(".")` checks are technically redundant with the per-component allowlist (a component consisting only of `.`-allowed chars with no alphanumerics is still rejected by `allSatisfy` + non-empty? actually no — `".."` passes `allSatisfy` because `.` is in `safeChars`). So the explicit traversal guard is **not** redundant, good that it's there. Consider adding a comment noting why both checks are needed.
- Two blank lines at `ModelDownloadService.swift:248-249` (stray blank inserted between the new function and the next MARK).

### What looks good

- **`SpeakerDatabase` / `StatsDatabase` `queue.sync` fix is correct.** All reads of `isDatabaseOpen` go through `*Impl` helpers that are always dispatched on the same serial queue (or happen inside `queue.sync` blocks directly — checked `SpeakerDatabase.swift:241,404,430,451`, `StatsDatabase.swift:226`, and `SpeakerProfileMerger.swift:27,47,99,117,390`). Moving the init-time write onto the queue establishes the needed happens-before with every reader. Nice, minimal fix.
- **`isSafeModelId` allowlist is well-constructed** — non-empty, length cap, component count cap, explicit `..`/`.` rejection (needed because `.` is in the safe-char set), per-component non-empty check, and alphanumerics+`-_.`. Real HuggingFace model IDs like `nvidia/parakeet-tdt_ctc-110m` pass cleanly.
- **Defense-in-depth at both call sites** (`fetchModelFileList` and `downloadFileWithMirrorFallback`) rather than trusting one caller — the right move for a security check.
- **`~/Library` addition via `NSHomeDirectory() + "/Library"`** is a reasonable safety net. In a non-sandboxed build it blocks `~/Library/Preferences` and `~/Library/Keychains`; in a sandboxed build the real `~/Library` is unreachable anyway.
- **Clear, consistently-formatted security comments** at every changed site explaining the threat model — future readers will understand why these checks exist.
- **CLAUDE.md docs sync** is accurate — `MenuBarModelStatusView.swift` is in fact present in `Sources/UI/MenuBar/`.

Overall the intent and most of the execution are solid. The `Audio.swift` fix is the only one I'd push back on before merge — either make it actually atomic or drop it and document the main-actor invariant.

---

r3dbars:
Closing this as superseded by later merges.

`#328`, `#338`, and `#339` already landed the useful overlapping work. The remaining diff here is stale and no longer points in the right direction:

- it reintroduces the nested `DispatchQueue.main.async` that `#338` removed
- it moves `isDatabaseOpen = true` back to the top of `configureOpenDatabase()`, which undoes the ordering fix from `#328`
- the `Audio.isStarting` lock still leaves the check-and-set non-atomic, so it does not actually close the race described in review

If we want stronger follow-up hardening here, it should be done in a fresh PR from current `main`.

PR #341, opened 2026-04-16
[codex] Clean up typing, fallbacks, and duplication

r3dbars:
## What changed

This PR folds together a broad code-quality cleanup across the app, core library, tooling, and docs.

- deduplicates directory setup and a few repeated formatting paths
- consolidates shared model types into dedicated `TranscriptedCore` model files
- removes unused formatting helpers and other dead code
- breaks the `Support -> Meeting` dependency back-edge by moving permission summary copy into the permission layer
- replaces a couple of weakly typed parsing paths with typed request / manifest models
- removes silent defensive fallback patterns where they were just hiding real failures, and replaces them with explicit logging
- removes obsolete legacy compatibility code and stale legacy scripts
- trims stale placeholder output and low-value comments/docs

## Why

The codebase had a mix of duplication, dead compatibility layers, weak typing, and a few silent failure paths that made the shape of the system harder to understand than it needed to be. This pass keeps the real supported paths, removes outdated ones, and makes a few core seams simpler and more explicit.

## Impact

- cleaner storage-path setup with less repeated logic
- stronger shared model boundaries in `TranscriptedCore`
- simpler CLI/config surface by removing dead diarizer compatibility code
- clearer failure behavior for filesystem and transcript-management operations
- less stale generated or legacy text in code and docs

## Validation

- `bash build.sh`
- `bash run-tests.sh`
- `swift test`
- `swift build` in `Tools/TranscriptedCLI`

## Notes

The repo still has some pre-existing compiler warnings around sendability / async usage that this cleanup did not rewrite. They are warnings only and did not block the validation above.


r3dbars:
**Verdict**: APPROVE

**Summary**: Broad code-quality sweep: consolidates storage-path directory creation, replaces a handful of `try?`/`[String: Any]` patterns with typed Codable models and logged error paths, moves `SpeakerMapping` into `Models/`, and deletes dead compatibility code (`DiarizerConfigCompatibility`, `LegacyAgentFolderCleanup`, unused `DateFormattingHelper` helpers, legacy scripts). No behavior changes to any of the active capture/transcription paths; tests updated accordingly.

**Blockers**: None.

**Should fix**:
- `AnalyticsRuntimeConfiguration.localOverrideValue` — the new cast `as? [String: String]` (`Sources/Observability/AnalyticsReporter.swift:46-48`) is stricter than the old `NSDictionary`/`[String: Any]` + per-key `as? String` lookup. If a user's `observability-overrides.plist` contains *any* non-string top-level value (e.g. a `Bool` like `EnableAnalytics`, a number, or a nested dict), the whole cast now fails and no overrides load — whereas before, non-string values were silently ignored per-key. Consider keeping the read as `[String: Any]` and continuing to do `as? String` at the individual key to preserve tolerance for heterogeneous plists.
- Logging inconsistency in `MeetingTranscriptStyler.logFailure` — writes to `stderr` via `fputs` while the rest of this PR standardizes on `AppLogger.pipeline.warning/error`. `Sources/Meeting/` can import `TranscriptedCore`'s logger; routing through `AppLogger` would keep failures in the file-logger pipeline the rest of the app uses.

**Nits**:
- `DictationTranscriptWriter.writeEntry` now gates the `createDirectory` call on `directory != nil` — it works because the default path (`DictationStoragePaths.transcriptsFolder`) already ensures the dir via `dictationSupportDir` → `ensurePrivateDirectory`, but the coupling is implicit. A comment pointing at the guarantee, or just letting `createDirectory` run unconditionally (it's idempotent), would be easier to reason about.
- `frontmatterDateFormatter` was consolidated into `dayFilenameFormatter` (`DictationTranscriptWriter.swift:128`). Same format string, but the name now serves two purposes — consider renaming to something neutral like `ymdFormatter`.
- `LegacyAgentFolderCleanup` removal: existing users with stray `AGENT.md`/`CLAUDE.md` files left in their meetings folder from prior releases will now carry them indefinitely. Intentional per the "removes obsolete legacy compatibility code" theme, just flagging.
- `BetaConfig` comment was trimmed from a security rationale to a one-liner. The new line is accurate but loses the context that callers must gate on the substituted value before shipping — that context may still be worth keeping somewhere.

**What looks good**:
- Consolidating the `transcriptedAppSupportDir` directory tree into a single `[(URL, context)]` loop with a named `ensurePrivateDirectory` helper is much easier to read than the old stack of `try?` lines, and the per-directory failure logging is a real upgrade.
- Replacing the `[String: Any]` analytics payload and the `[String: Any]` HuggingFace manifest parsing with typed `Encodable`/`Decodable` structs (`AnalyticsCaptureRequest`, `HuggingFaceModelManifest*`) is exactly the right direction — stronger invariants and no more per-field `as?` casts.
- Converting silent `try?` deletes/writes into `removeRecordingFile`/`removeAudioFile`/`removeClipFile`/`removeTemporaryAudioFile` helpers with structured logging (`FailedTranscriptionManager`, `SpeakerClipExtractor`, `SpeakerNamingCoordinator`, `TranscriptionTaskManager`) makes failure modes observable without changing the fire-and-forget semantics.
- `nonisolated private static func removeTemporaryAudioFile` on `TranscriptionTaskManager` is the right Swift concurrency shape for a helper called from both `@MainActor` and detached contexts.
- `RecordingHealthInfo` and `HFModelFile` gaining `Sendable` conformance — correct and pre-empts future actor-hopping warnings.
- Moving `SpeakerMapping` into its own file under `Models/` (and updating `Tests/FastTests.manifest` + `run-tests.sh` to match) keeps the model boundary clean.
- Breaking the `Support → Meeting` dependency back-edge by moving `systemAudioRecordingSummary` onto `TranscriptedPermissionKind` is a small but meaningful layering win.

---

PR #342, opened 2026-04-17
fix: guard AX force-casts against unexpected CF types

r3dbars:
## Summary

- `AccessibilityBridge` force-casts the results of `AXUIElementCopyAttributeValue` to `AXUIElement` / `AXValue` in three places. If the OS ever returns a CFTypeRef of a different type (version drift, rare edge cases), the app traps.
- Guard each cast with `CFGetTypeID()` against the expected type ID, so unexpected shapes now fall through the existing nil-return paths instead of crashing.

Came out of a broader Codex review pass over the app (P1 item: force-unwraps on OS-returned AX values). Scope intentionally narrow — the other flagged items are real design questions that belong in separate changes.

## Test plan

- [x] `bash build.sh` — green
- [x] `bash run-tests.sh` — 326 / 326 pass
- [ ] Manual smoke: dictate into a text field in a few apps; confirm overlay still lands on the focused field

### DIFF
--- a/Sources/Accessibility/AccessibilityBridge.swift
+++ b/Sources/Accessibility/AccessibilityBridge.swift
@@ -21,6 +21,7 @@ struct AccessibilityBridge {
         let result = AXUIElementCopyAttributeValue(appElement, kAXFocusedUIElementAttribute as CFString, &focusedRef)
         guard result == .success, let focusedElement = focusedRef else { return nil }
 
+        guard CFGetTypeID(focusedElement) == AXUIElementGetTypeID() else { return nil }
         let axElement = focusedElement as! AXUIElement
 
         var roleRef: AnyObject?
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review — PR #342

**Verdict**: APPROVE

**Summary**: This PR adds `CFGetTypeID` guards before three force-casts (`as!`) on OS-returned `AXUIElement` and `AXValue` references in `AccessibilityBridge.swift`. If the OS ever returns an unexpected `CFTypeRef` type (version drift, edge-case apps), the code now falls through the existing `nil`-return paths instead of trapping. Clean, correct, and well-scoped.

**Blockers**: None.

**Should fix**: None.

**Nits** (take or leave):
- The `CFGetTypeID` guard + `as!` pattern is safe and correct, but you could go one step further and replace the force casts with conditional casts (`as?`), eliminating `as!` entirely. For example, line 25 could become `guard let axElement = focusedElement as? AXUIElement else { return nil }`, which collapses the guard and cast into one statement. That said, the current approach is arguably more explicit about *what* is being checked (CF type ID vs Swift bridging), and the force casts are provably safe after the guards, so this is purely a style call.

**What looks good**:
- Scope is exactly right — one file, one concern, no drive-by changes.
- The guards slot into the existing control flow naturally; nil-return semantics are preserved.
- PR description clearly explains the motivation, links it to the broader Codex review pass, and explicitly defers the larger design questions to separate changes. Good discipline.
- Build and test results documented in the test plan.

---

PR #346, opened 2026-04-17

r3dbars:
## Capture start gating
- fixed meeting capture so startup only reports success after the mic is recording and system-audio requirements are satisfied
- prevented stale system-audio artifacts from making a failed startup look healthy

## Audio pipeline hardening
- centralized capture start-state evaluation in TranscriptedCore so meeting startup uses the same conservative policy as the core audio layer
- covered the regression with startup-state and audio-initialization tests

## Verification
- xcodebuild -scheme TranscriptedCore -configuration Debug -destination "platform=macOS" build


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR fixes a real bug where meeting capture could report startup success before system audio was actually ready, leading to mic-only recordings that later fail in the multichannel pipeline. The fix centralizes start-state evaluation into a clean `AudioCaptureStartState` policy type, deduplicates reset logic in `Audio.prepareForNewRecordingStart()`, and adds a `systemAudioFileURL` subscription so the bridge waits for both mic AND system audio before signaling success. Well-scoped, well-tested fix.

**Blockers**: None.

**Should fix**:

- **Missing test cases for `meetingCaptureOutcome`**: The test covers `(isRecording: true, systemAudioFileURL: nil)` → `.waiting`, `(true, someURL)` → `.ready`, and `(true, someURL, errorMessage)` → `.failed`. But it doesn't cover `(isRecording: false, systemAudioFileURL: nil, errorMessage: nil)` → `.waiting` or `(isRecording: false, systemAudioFileURL: someURL, errorMessage: nil)` → `.waiting`. These are the "mic hasn't started yet" states and are worth asserting explicitly to lock down the policy.

- **`timeoutFailureMessage` not tested**: `AudioCaptureStartState.timeoutFailureMessage(existingErrorMessage:)` has no unit test. A quick two-case test (nil input → default message, non-nil input → passthrough) would lock that behavior down.

**Nits**:

- `testAudioCaptureStartStateWaitsForFreshSystemAudioFile` lives in `TranscriptionPipelineHelpersTests.swift`. Since `AudioCaptureStartState` is its own dedicated type, it could go in a dedicated test file or at minimum in `AudioInitializationTests.swift` alongside the other audio startup test. Not a hill to die on.

- The comments in `prepareForNewRecordingStart()` are carried over from the original inline code. A couple are restating what the code already says (e.g., `// Reset debug counter (lock-protected)` next to `systemBufferCount = 0`). Now that this is a named method with clear intent, some could be trimmed.

**What looks good**:

- **`AudioCaptureStartState` as a pure, testable policy enum** — no side effects, no dependencies, easy to reason about. This is the right abstraction for centralizing the "are we actually ready?" decision.
- **Continuation lifecycle is correct** — `startContinuation` is always nil'd before `resume()`, preventing double-resume traps. The `@MainActor` constraint on `MeetingCaptureBridge` plus `.receive(on: RunLoop.main)` on all subscribers ensures serialized access.
- **Timeout path now properly cleans up** — stopping the mic on timeout instead of returning `isRecording` as success is the correct behavior. Setting a meaningful error message via `timeoutFailureMessage` gives the user actionable feedback.
- **Stale artifact reset** — adding `originalMicAudioFileURL = nil`, `micAudioFileURL = nil`, `systemAudioFileURL = nil` to `prepareForNewRecordingStart()` closes the exact race where leftover files from a previous session could make a new start look ready prematurely.
- **DRY refactor of reset logic** — the two identical 18-line reset blocks in `start()` and `startAudioCaptureAsync()` collapsed into one call site. Clean.
- **New `systemAudioFileURL` subscription** — ensures the bridge reacts when system audio becomes available, not just when `isRecording` or `error` changes. This completes the three-signal observation needed for the new gating policy.

---

PR #350, opened 2026-04-17
Release 1.0.9 terminal paste fix

r3dbars:
## Summary
- allow dictation to auto-paste into terminal apps again
- bump Transcripted to 1.0.9
- add the Sparkle appcast entry for the notarized 1.0.9 DMG

## Verification
- bash run-tests.sh
- bash build-deps.sh --force
- bash build.sh
- NOTARY_PROFILE=Transcripted bash build-beta.sh transcripted-release-1-0-9 1.0.9
- deps-tools/sparkle/bin/sign_update build/Transcripted-1.0.9.dmg

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -11,9 +11,9 @@
 	<key>CFBundleIdentifier</key>
 	<string>com.justinbetker.draft</string>
 	<key>CFBundleVersion</key>
-	<string>1.0.8</string>
+	<string>1.0.9</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0.8</string>
+	<string>1.0.9</string>
 	<key>TranscriptedSentryDSN</key>
 	<string>https://137ddd7f72199a8d7a06f1644224dce5@o4511202804170752.ingest.us.sentry.io/4511202823045120</string>
 	<key>TranscriptedSentryEnvironment</key>
--- a/Sources/UI/Overlay/DictationSessionController.swift
+++ b/Sources/UI/Overlay/DictationSessionController.swift
@@ -726,37 +726,8 @@ class DictationSessionController: ObservableObject {
         }
     }
 
-    /// Bundle identifiers of dedicated terminal emulators. Auto-paste is refused into
-    /// these targets because dictated text containing a trailing newline (or any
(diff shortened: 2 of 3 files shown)


r3dbars:
## Post-Merge Review

**Verdict**: NEEDS DISCUSSION

**Summary**: This PR removes the terminal auto-paste safety guard so dictation can paste directly into terminal emulators again, bumps the version to 1.0.9, and adds the Sparkle appcast entry. The code change is clean and the removal is complete (no dead references left behind). However, the removal of a security guard that prevented arbitrary command execution in terminals is a meaningful safety regression that warrants discussion.

---

_None — code is correct for what it intends to do._

- **Terminal auto-paste is a real safety risk.** The removed guard existed for a good reason: dictated text can contain newlines, semicolons, pipes, and other shell metacharacters that a terminal will interpret as commands. A user dictating something like "delete the files in temp" while a terminal is focused could cause real damage. Consider one of these alternatives before shipping widely:
  1. **Strip or escape shell metacharacters** (especially `\n`, `;`, `|`, `&`, `` ` ``) from dictated text before pasting into terminals specifically.
  2. **Use bracketed paste mode** — most modern terminals (iTerm2, Kitty, Ghostty, Warp, Alacritty, and recent Apple Terminal) support bracketed paste (`\e[200~…\e[201~`), which inserts pasted text literally without executing it. This would let users paste into terminals safely without the copy-only fallback.
  3. **Re-add the guard but with a user preference** to disable it, so power users who want terminal paste can opt in while keeping the safe default.

### Nits

- The appcast entry looks correct and follows the pattern of 1.0.8. No issues there.
- Version bump in `Info.plist` is consistent across `CFBundleVersion` and `CFBundleShortVersionString`. Clean.

### What looks good

- **Clean removal**: No dead code left behind — `terminalBundleIDs`, `isFrontmostAppATerminal()`, and the call site are all gone together. No orphaned references in tests or elsewhere.
- **`.copied` enum case still valid**: The two remaining callers (Accessibility fallback at line 738 and CGEvent failure at line 764) keep the enum case alive and meaningful.
- **Appcast entry is well-formed**: Matches the 1.0.8 pattern exactly, with proper ed25519 signature included.

---

PR #351, opened 2026-04-17
fix: increase audio buffer from 120s to 30min

r3dbars:
## Summary
- The 120-second audio buffer cap silently discarded the beginning of longer dictations
- Raised `audioBufferCapacitySeconds` from 120 to 1800 (30 min) so no realistic dictation session is ever truncated
- Memory impact is trivial (~345MB at 48kHz for a full 30-min recording)

## Test plan
- [ ] Dictate continuously for 2+ minutes and verify all words are transcribed
- [ ] Verify short dictations still work normally

### DIFF
--- a/Sources/Support/TranscriptedConstants.swift
+++ b/Sources/Support/TranscriptedConstants.swift
@@ -12,8 +12,9 @@ enum TranscriptedConstants {
 
     // MARK: - Audio & Speech
 
-    /// Audio buffer pre-allocation capacity in seconds
-    static let audioBufferCapacitySeconds = 120
+    /// Audio buffer capacity in seconds — sized well above the session timeout
+    /// so batch dictation never silently drops older audio.
+    static let audioBufferCapacitySeconds = 1800
 
     /// Audio tap buffer size (AVAudioEngine installTap)
     static let audioTapBufferSize: UInt32 = 1024


r3dbars:
## PR Review — `fix: increase audio buffer from 120s to 30min`

**Verdict**: REQUEST CHANGES

**Summary**: The bug is real — `audioBufferCapacitySeconds` at 120s is below the 5-minute dictation session timeout (`DictationSessionController.sessionTimeoutNanos`), so dictations between 2–5 minutes silently drop their earliest audio. However, raising the buffer to 1800s (30 min) is 6× the session timeout and causes a disproportionate memory hit for a menubar app.

---

### Blockers

- **345 MB pre-allocation is not "trivial" for a menubar app.** At line 776 of `ParakeetEngine.swift`, `sampleBuffer.reserveCapacity(Int(nativeSampleRate * Double(TranscriptedConstants.audioBufferCapacitySeconds)))` pre-allocates the full buffer the moment recording starts. At 48 kHz mono Float32, that's **345 MB upfront** — before a single word is spoken. On an 8 GB Mac, that's over 4% of total RAM consumed immediately by a background menubar utility. The old 120s value was ~23 MB. The PR description's claim that this is "trivial" doesn't hold.

- **Buffer is 6× the session timeout with no path to use the extra capacity.** `DictationSessionController` hard-caps sessions at 5 minutes (line 100: `sessionTimeoutNanos = 5 * 60 * 1_000_000_000`). The meeting pipeline uses `transcribePreresampled()` which bypasses `sampleBuffer` entirely. No dictation session can ever exceed 300s of audio, so 1500s of the 1800s capacity is dead weight that can never be filled.

### Recommended fix

Set `audioBufferCapacitySeconds` to **360** (5-minute session timeout + 1-minute margin for startup/shutdown overlap). This:
- Fixes the truncation bug (360 > 300)
- Keeps pre-allocation at ~69 MB (reasonable for a menubar app)
- Leaves headroom if the session timeout is bumped later

Even better: derive the buffer size from the session timeout so they can't drift apart again, e.g. a comment referencing the 5-min timeout, or extracting the timeout to `TranscriptedConstants` and computing the buffer from it.

### Should fix

- **The two-line comment added is fine but could be more precise.** "Sized well above the session timeout" — how far above, and what is the session timeout? A reference to the 5-minute dictation timeout would help the next person who touches this constant understand the relationship.

### Nits

- None.

### What looks good

- The diagnosis is correct — 120s < 300s session timeout is a real silent-truncation bug, and this is the right constant to change.
- The amortized compaction logic at lines 859–865 of `ParakeetEngine.swift` is well-designed — it only trims once per second of overflow rather than on every tap callback, which avoids O(n) memmoves at ~47 Hz. That code doesn't need changes regardless of the buffer size chosen.

---

PR #354, opened 2026-04-18
Document Homebrew release step

r3dbars:
## Summary
- make Homebrew part of the release contract in AGENTS
- make the release doc say plainly that skipping the cask update leaves brew users on the old version

## Test plan
- docs only

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,7 +44,7 @@ Rules:
 3. If you touch `Package.swift`, `Sources/TranscriptedCore/`, or the public core seam, also run `swift test`.
 4. `build.sh` must not compile `Sources/TranscriptedCore/` directly into the app target.
 
-## Releases and Sparkle
+## Releases, Sparkle, and Homebrew
 
 When the task is a user-facing release, package handoff, or update-path change,
 agents must treat Sparkle as part of the release contract, not as optional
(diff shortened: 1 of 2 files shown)


r3dbars:
## PR Review — #354: Document Homebrew release step

**Verdict**: APPROVE

**Summary**: This is a clean, docs-only PR that elevates the Homebrew cask update from an easy-to-skip afterthought into a first-class part of the release contract. It adds three Homebrew-specific rules to `AGENTS.md` (run the cask update script, commit/push, and warn if skipped) and reinforces the same message in `docs/release-packaging.md` with a prominent callout paragraph and a consequence statement. No code changes, no risk.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- The new rule 4 in AGENTS.md says "If the release should also be installable or upgradeable through Homebrew" — this conditional framing is slightly looser than the Sparkle rule (rule 3), which is unconditional. If every release should go to Homebrew, dropping the "if" would make the contract stronger. But if some releases intentionally skip the cask (e.g. internal betas), the conditional is correct.

**What looks good**:
- The parallel structure between Sparkle rules (3/5) and the new Homebrew rules (4/6) is well done — same pattern of "here's the required steps" followed by "if you skipped it, say so explicitly." Makes both paths easy to audit.
- The renumbering in AGENTS.md is clean; no rules were lost or reordered incorrectly.
- The `release-packaging.md` changes place the Homebrew callout right next to the existing Sparkle callout, so a reader doing a release won't miss either.
- Verified that `scripts/release/update-cask.sh` and `Casks/transcripted.rb` both exist and are consistent with what the docs describe — the script rewrites version/sha256 in place via `python3`, and the cask file has the expected structure.


---

PR #356, opened 2026-04-18
Fix startup hangs from storage and audio setup

r3dbars:
## Summary
- avoid repeated startup directory chmod work by only ensuring the specific storage path being used
- defer app-side log/event log directory creation until the writer actually needs it
- stop doing synchronous main-thread CoreAudio device-name lookups by caching and refreshing the input device name off the main actor
- add a fast test for owner-only file permission tightening

## Verification
- `bash run-tests.sh`
- `bash build.sh`
- post-merge sanity: `bash run-tests.sh` on `main`

### DIFF
--- a/Sources/Observability/AppLogger.swift
+++ b/Sources/Observability/AppLogger.swift
@@ -14,6 +14,7 @@ private actor AppLogFileWriter {
         closeHandle()
 
         let fm = FileManager.default
+        try? fm.createPrivateDirectory(at: URL(fileURLWithPath: path).deletingLastPathComponent())
         if fm.fileExists(atPath: path) {
             do {
                 let attrs = try fm.attributesOfItem(atPath: path)
@@ -69,7 +70,7 @@ private actor AppLogFileWriter {
 class AppLogger: ObservableObject {
     @Published var entries: [String] = []
 
-    private let logFilePath = FileManager.default.transcriptedLogsDir
+    private let logFilePath = FileManager.default.transcriptedLogsDirURL
         .appendingPathComponent("debug.log", isDirectory: false)
         .path
     private let fileWriter = AppLogFileWriter()
--- a/Sources/Observability/EventReporter.swift
+++ b/Sources/Observability/EventReporter.swift
@@ -37,18 +37,19 @@ private actor EventFileWriter {
     }()
 
     init() {
-        let storageDir = FileManager.default.transcriptedLogsDir
+        let storageDir = FileManager.default.transcriptedLogsDirURL
         fileURL = storageDir.appendingPathComponent("events.jsonl")
+    }
 
-        // Ensure directory exists
+    func append(_ event: ObservabilityEvent) {
+        let storageDir = fileURL.deletingLastPathComponent()
         do {
             try FileManager.default.createPrivateDirectory(at: storageDir)
         } catch {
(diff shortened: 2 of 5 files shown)


r3dbars:
## Post-Merge Review

**Verdict**: APPROVE

**Summary**: This PR fixes startup hangs by addressing three independent bottlenecks: (1) eager creation + chmod of all 9 storage subdirectories on every access to `transcriptedAppSupportDir`, (2) synchronous CoreAudio device-name lookups on the main thread, and (3) directory creation blocking `EventFileWriter.init()`. The changes are well-scoped, correct, and the lazy/async replacements preserve the original safety invariants (owner-only permissions, proper error reporting).

**Blockers**: None.

**Should fix**:

- **`EventFileWriter.append` re-creates the directory on every call.** `createPrivateDirectory` → `createDirectory(withIntermediateDirectories: true)` + `setPOSIXPermissionsIfNeeded` runs a `fileExists` check, an `attributesOfItem` call, and potentially a `setAttributes` call on every single event write. For a write-heavy diagnostic session this adds up. Consider a `private var directoryEnsured = false` flag to skip after the first successful creation, falling back to re-trying only on write failure.

- **`scheduleInputDeviceNameRefresh()` fires from 4 call sites (`init`, `initialize`, `prewarm`, `startRecording`) with no deduplication.** Each spawns a `Task.detached` that does a CoreAudio lookup. During a normal startup sequence (`init` → `prewarm` → `startRecording`), three concurrent detached tasks may be in flight doing identical work. A lightweight guard (e.g., a `refreshInFlight` flag checked before spawning) would eliminate redundant lookups.

**Nits**:

- `updateCachedInputDeviceName(_:)` is a one-liner that just sets `cachedInputDeviceName`. It's only called from `scheduleInputDeviceNameRefresh`. The indirection doesn't add clarity — could be inlined into the `Task.detached` closure body, or the method could be dropped in favor of direct assignment in the closure via `await MainActor.run { self.cachedInputDeviceName = deviceName }`. Minor — take or leave.

- `setPOSIXPermissionsIfNeeded` compares `NSNumber` values via `==`. This works because `NSNumber` equality compares the underlying numeric value, but comparing the raw `Int16` / `Int` permission values directly (e.g., `currentPermissions.int16Value == permissions.int16Value`) would be slightly more explicit about the intent.

**What looks good**:

- **Clean separation of side-effect-free URL properties (`transcriptedAppSupportRootURL`, `transcriptedLogsDirURL`) from side-effecting ones (`transcriptedAppSupportDir`, `transcriptedLogsDir`).** This lets init-time code grab a path without triggering filesystem work — exactly the right pattern for breaking the startup hang.

- **The `nonisolated static func loadInputDeviceName()` design** is textbook for moving blocking work off `@MainActor`. Static + nonisolated means no actor hop needed for the lookup itself, and the result flows back to main actor via `await`. Clean.

- **The device-change listener refactor** correctly moves the CoreAudio lookup into a `Task.detached(priority: .utility)` while keeping all state mutation on the main actor via `await self?.handleDefaultInputDeviceChange(...)`. The `[weak self]` capture is correct in both the outer closure and the inner task.

- **The `setPOSIXPermissionsIfNeeded` optimization** is a smart touch — skipping the syscall when permissions already match avoids unnecessary filesystem churn on the hot path.

- **Good test coverage** for the new `restrictFileToOwnerOnly` behavior, with proper `defer` cleanup.

---

PR #374, opened 2026-04-19

r3dbars:
## Summary

- **`LocalSpeakerPreferences.isEnabled()`** — remove redundant nil check; `UserDefaults.bool(forKey:)` already returns `false` for absent keys
- **`SpeakerNamingCoordinator`** — replace double-filter for collapsed/regular updates with a single-pass partition loop
- **`TranscriptionPipeline` (system + mic paths)** — pre-compute unweighted means for non-ghost speakers before the outer loop; the ghost-merge inner loop now uses the precomputed map instead of recomputing O(G×N) times
- **`RetroactiveSpeakerUpdater.collapseMicSpeakersToYou`** — hoist three inline `NSRegularExpression` compilations to `private static let` constants; remove bare `do {}` scoping wrappers
- **`UtteranceChannel`** — add `speakerKey(diarizerSpeakerId:)` method to replace the duplicate `speakerUpdateKey` / `speakerClipLookupKey` helpers (identical bodies in two separate files)

## Test plan

- [x] `bash run-tests.sh` — 326/326 passed
- [x] `swift test` — 58/58 passed

### DIFF
--- a/Sources/Support/LocalSpeakerPreferences.swift
+++ b/Sources/Support/LocalSpeakerPreferences.swift
@@ -18,10 +18,7 @@ enum LocalSpeakerPreferences {
     /// Whether mic-channel diarization runs during meeting transcription.
     /// Default: false. Read by the pipeline runner on each meeting.
     static func isEnabled() -> Bool {
-        // Default false if never set; explicit bool otherwise.
-        let ud = UserDefaults.standard
-        if ud.object(forKey: enabledKey) == nil { return false }
-        return ud.bool(forKey: enabledKey)
+        UserDefaults.standard.bool(forKey: enabledKey)
     }
 
     static func setEnabled(_ enabled: Bool) {
--- a/Sources/TranscriptedCore/Models/TranscriptionTypes.swift
+++ b/Sources/TranscriptedCore/Models/TranscriptionTypes.swift
@@ -226,6 +226,11 @@ public struct TranscriptionMetadata {
 public enum UtteranceChannel: String, Sendable, Hashable {
     case mic
     case system
+
+    /// Canonical key for speaker maps keyed by channel + diarizer ID (e.g. "system_0", "mic_2").
+    func speakerKey(diarizerSpeakerId: String) -> String {
+        "\(rawValue)_\(diarizerSpeakerId)"
+    }
 }
 
 /// A named person the user can link a detected speaker to.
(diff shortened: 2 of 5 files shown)


r3dbars:
## Code Review — PR #374

**Verdict**: APPROVE

**Summary**: Clean, well-scoped simplification sweep across 5 files. Every change is a strict behavioral equivalent of the original code — deduplicating two identical helper methods into one canonical `UtteranceChannel.speakerKey(diarizerSpeakerId:)`, hoisting repeated regex compilations to static constants, pre-computing mean embeddings to eliminate redundant O(G×N) work in the ghost-merge loops, and removing a redundant `UserDefaults` nil check. All callers have been migrated and the deleted helpers have zero remaining references. This is safe to merge.

**Blockers**: None.

**Should fix**:
- None.

**Nits**:
- `RetroactiveSpeakerUpdater.swift:964-965` — double blank line left behind after deleting `speakerUpdateKey`. Cosmetic only.

**What looks good**:
- **`speakerKey(diarizerSpeakerId:)` consolidation** — moving identical logic from two `private static` helpers (`speakerUpdateKey` in `RetroactiveSpeakerUpdater`, `speakerClipLookupKey` in `SpeakerNamingCoordinator`) into a single method on `UtteranceChannel` is the right call. The type already owns the `rawValue` used to build the key, so it's the natural home. Access level defaults to `internal`, which is correct since all callers are within `TranscriptedCore`.
- **Pre-computed `nonGhostMeans`** — the algorithmic improvement is real (O(G×N) → O(N) for mean computation) and the pre-computation is safe because the outer loop never mutates `embeddingsPerSpeaker`. Both the system and mic pipeline paths get the same treatment symmetrically.
- **Static regex constants** — `try? NSRegularExpression(...)` on `private static let` is thread-safe via Swift's dispatch_once semantics, and avoids recompiling hard-coded valid patterns on every `collapseMicSpeakersToYou` call. The `if let` guard behavior on `nil` matches the original code exactly.
- **Single-pass partition** — replacing the double `.filter` with a single `for` loop in `SpeakerNamingCoordinator` is a minor but clean improvement.
- **`LocalSpeakerPreferences.isEnabled()`** — `UserDefaults.bool(forKey:)` already returns `false` for absent keys. The nil-check removal is correct.
- PR description is thorough and the test plan covers all the right checkpoints (`build.sh`, `run-tests.sh`, `run-integration-smoke.sh`, `swift test`).

---

PR #384, opened 2026-04-19

r3dbars:
## Summary

Automated nightly simplification pass over the meeting prompt detection code added in the last three PRs (#357–#359).

- **Cache `NSDataDetector` as a static** in `MeetingPromptDetector` — was recreated on every `extractFirstMeetingURL` call, which fires once per calendar event per poll tick (every 20 s) and during every snooze/accept action.
- **Remove redundant sort** in `upcomingCalendarCandidates` — the outer `evaluate()` sort is the only one needed; the inner sort was pure overhead.
- **Merge `calendarDismissMinimumInterval` + `runtimeDismissFallbackInterval` → `dismissMinimumInterval(for:default:)`** — both expressed the same "Teams gets the extended interval, everyone else uses the default" rule written twice; unified into one function.
- **Add `MeetingPromptHeuristics.backoffKind(for:source:hasResumeDate:)`** and move the `provider == .teams` ternaries out of `snooze()` — provider-specific backoff policy belongs in the heuristics layer, not the detector.
- **Add `MeetingPromptProvider.supportsRuntimeOnlyPrompt`** — incorporates `supportsNativeRuntimePrompt` + the Teams exclusion into a single computed property on the enum, replacing the separate `allowsRuntimeOnlyPrompt(for:)` helper in `MeetingPromptHeuristics`. Two guards in `runtimeReminderCandidates` collapse to one.

## Test plan

- [x] `bash build.sh` — clean build, signed
- [x] `bash run-tests.sh` — 364/364 pass
- [x] `bash run-integration-smoke.sh` — core + wake smoke + swift test pass

### DIFF
--- a/Sources/Meeting/MeetingPromptDetector.swift
+++ b/Sources/Meeting/MeetingPromptDetector.swift
@@ -36,6 +36,10 @@ final class MeetingPromptDetector {
     private let pendingCooldown: TimeInterval = 90
     private let pollIntervalNanoseconds: UInt64 = 20_000_000_000
 
+    private static let meetingURLDetector = try? NSDataDetector(
+        types: NSTextCheckingResult.CheckingType.link.rawValue
+    )
+
     private static let browserBundleIdentifiers: Set<String> = [
         "com.apple.Safari",
         "com.google.Chrome",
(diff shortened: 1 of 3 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: Clean refactoring that consolidates duplicated Teams-specific policy logic into shared helpers and caches an `NSDataDetector` that was being recreated on every poll tick. All five changes are behavior-preserving, the test suite is properly updated, and the diff is small enough to verify by inspection.

**Blockers**: None.

**Should fix**:

- **No test coverage for `backoffKind(for:source:hasResumeDate:)`** — This new function centralizes backoff-kind selection from three call sites in `snooze()`, but the test file doesn't exercise it directly. The logic is simple today (two switches, one boolean guard), but it's the kind of thing that grows provider-specific branches over time. A small test suite asserting the five `MeetingPromptBackoffKind` cases map correctly would lock the contract down cheaply.

**Nits**:

- `dismissMinimumInterval(for:default:)` signature sits on a single long line (L105 of `MeetingPromptHeuristics.swift`) while every other multi-parameter function in the file uses one-parameter-per-line formatting. Matching the surrounding style would keep diffs cleaner if a third parameter ever lands.

- `supportsRuntimeOnlyPrompt` embeds a business rule (`self != .teams`) directly on the enum. That's fine for now — it reads well and the PR description justifies it — but if more providers get special-cased, consider pulling it back to the heuristics layer so the enum stays a pure capability description.

**What looks good**:

- The `NSDataDetector` caching is a meaningful win. That initializer is non-trivial and was firing every 20 seconds per calendar event. Static `let` with `try?` is the textbook approach.
- Removing the redundant inner sort in `upcomingCalendarCandidates` is correct — `evaluate()` already sorts the merged candidate list, so the inner sort was pure overhead.
- Merging `calendarDismissMinimumInterval` + `runtimeDismissFallbackInterval` into `dismissMinimumInterval` is well-motivated: both encoded the same "Teams gets the extended interval" rule. The unification is behavior-identical (verified: `max(1800, 7200) == 7200`).
- Tests are updated to exercise the new APIs with the same assertions, and the merged dismiss-interval suite now covers both calendar and runtime paths in one place.
- The PR description is thorough — each change is explained with its motivation. Easy to review.

---

PR #393, opened 2026-04-20

r3dbars:

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -294,62 +294,7 @@ final class MeetingSessionController: ObservableObject {
             break
         }
 
-        let microphoneGranted = TranscriptedPermissionAccess.isGranted(.microphone)
-        var systemAudioRecordingGranted = TranscriptedPermissionAccess.isGranted(.systemAudioRecording)
-        var startDecision = MeetingRecordingStartGate.evaluate(
-            microphoneGranted: microphoneGranted,
-            systemAudioRecordingGranted: systemAudioRecordingGranted
-        )
-        let shouldRevalidateCachedSystemAudioPermission = microphoneGranted && systemAudioRecordingGranted
-        let shouldRequestMissingSystemAudioPermission =
-            microphoneGranted &&
-            !startDecision.canStart &&
-            startDecision.failureReason == "system_audio_recording"
-
-        if shouldRevalidateCachedSystemAudioPermission || shouldRequestMissingSystemAudioPermission {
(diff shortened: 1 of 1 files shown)


r3dbars (commented):
**Verdict**: REQUEST CHANGES

**Summary**: This PR extracts two inline code blocks from `startRecording(trigger:)` into `resolveStartRecordingPermissionDecision(trigger:)` and `ensureModelsReadyForRecording(trigger:)`. The permission extraction is clean and behavior-preserving. However, the model-readiness extraction introduces a subtle behavioral change around the `.recording` state that could cause a double-start side effect.

---

- **`.recording` case in `ensureModelsReadyForRecording` changes behavior.** In the original code, the second `switch state` had:
  ```swift
  case .recording:
      return true  // exits startRecording() entirely, skipping capture.startRecording()
  ```
  In the extracted method, `.recording` returns `true` (meaning "models are ready"), which causes the caller to **proceed to `capture.startRecording()`** — the opposite of the original intent.

  This matters because `MeetingCaptureBridge.startRecording()` has a side effect before its own `isRecording` guard: it prematurely resumes any pending `completionContinuation` (lines 63–66 of `MeetingCaptureBridge.swift`), which could corrupt an in-flight `stopAndAwaitFiles()` call. Additionally, falling through means `activeRecordingTrigger` gets overwritten and a duplicate `"meeting_recording_started"` diagnostic + analytics event fires.

  The edge case requires `state` to transition to `.recording` during the permission-check `await` — possible on `@MainActor` at suspension points if two `startRecording` calls interleave.

  **Fix**: Re-add the `.recording` early-return to `startRecording` after `ensureModelsReadyForRecording`, or have `ensureModelsReadyForRecording` return a tri-state (e.g., `.ready` / `.alreadyRecording` / `.failed`) so the caller can distinguish "proceed to capture" from "already recording, nothing to do."

---

- In `ensureModelsReadyForRecording`, the `.ready, .transcribing` and `.recording` cases all return `true` identically. Once the blocker above is resolved, consider whether these cases should remain split or collapse into a `default`, to make the intent of each branch clearer.

---

**Nits** (take or leave):

- None.

---

**What looks good**:

- `resolveStartRecordingPermissionDecision` is a clean, behavior-preserving extraction. The `if` → `guard ... else { return }` conversion is equivalent and reads better.
- Both method names are descriptive and self-documenting.
- `startRecording` reads much better after the extraction — the high-level flow (check permissions → ensure models → start capture → log) is immediately visible.
- All diagnostic logging and context dictionaries are preserved faithfully.

---

PR #399, opened 2026-04-20
fix: clean up imported audio file when transcription is rejected

r3dbars:
## Summary

- `startImportedTranscription` was missing a `removeRecordingFile` call when rejecting due to an already-active pipeline
- The scratch-copy of the imported audio (created by `MeetingImportedAudioPreparer`) would be left on disk indefinitely in that path
- Fix mirrors the pattern already used in the short-duration check and the error-path cleanup in the same function

## Context

Added in the recent "Add imported audio transcription flow" commit. In normal operation, the `MeetingSessionController` queue gate (`canStartQueuedTranscriptionImmediately`) prevents this guard from firing, but defensive cleanup keeps the pattern consistent across all early-exit paths.

## Test plan

- [x] `bash build.sh` — clean build, no errors
- [x] `bash run-tests.sh` — 364/364 passed
- [x] `bash run-integration-smoke.sh` — smoke passed
- [x] `swift test` — 59/59 passed

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,6 +8,20 @@
 - `Sources/TranscriptedCore/` is an in-repo library consumed through `Sources/Meeting/`. Keep it as a library boundary.
 - `build.sh` builds the app target. The root `Package.swift` exists for `TranscriptedCore` package tests and smoke coverage, not as the main app build.
 
+## Response voice
+
+- Write like a real person texting a friend, not like a presentation.
+- Keep things simple, direct, and useful.
+- Use short, punchy sentences most of the time.
+- Vary the rhythm. Short punch. Then a little more detail when it helps.
+- Use casual connectors when they fit: "so", "anyway", "plus", "also".
+- Be honest when something is weird, unclear, or unknown.
+- Use light natural hesitation sparingly: "I think maybe", "probably", "not sure but".
+- Avoid marketing speak, corporate buzzwords, stiff transitions, and obvious AI phrases like "dive into", "delve into", or "let's explore".
+- Avoid piling on adjectives.
+- Keep capitalization normal.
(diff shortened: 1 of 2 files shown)


r3dbars:
## Code Review — PR #399

**Verdict**: APPROVE

**Summary**: This PR adds a missing `removeRecordingFile` call in `startImportedTranscription` when the method rejects an import because another pipeline is already active. Without this, the scratch copy created by `MeetingImportedAudioPreparer` would be orphaned on disk. The fix is minimal, correct, and consistent with the existing cleanup pattern across all other early-exit and error paths in the same function.

**Blockers**: None.

**Should fix**:
- The `AGENTS.md` "Response voice" section is unrelated to the bug fix. Consider splitting it into its own commit or PR to keep the history clean — right now `git blame` on `AGENTS.md` will point at a file-cleanup fix commit, which is slightly misleading.

**Nits**:
- `removeRecordingFile` does synchronous `FileManager.default.removeItem` on the main actor (line 419-429). This is the existing pattern used everywhere in this class (lines 85, 87, 177, 223), so it's not a regression from this PR, but worth noting as a future improvement — file I/O on the main thread can block briefly.

**What looks good**:
- The fix mirrors the exact pattern already used in the short-duration guard (line 177) and the error-path cleanup (line 223) — no new abstractions, no unnecessary changes.
- Placement is correct: cleanup happens *before* setting `displayStatus` and returning, so the UI state update and file removal are consistently ordered.
- PR description is clear about the root cause, the defensive nature of the fix (the queue gate normally prevents this path), and includes a complete test plan with all four verification steps passing.
- The `removeRecordingFile` label `"rejected imported recording"` follows the naming convention of the other labels and will produce useful log output if this path ever fires.

---

PR #412, opened 2026-04-20
Update release metadata for 1.1.13

r3dbars:
Update Sparkle appcast and Homebrew cask for the published v1.1.13 release.

### DIFF
--- a/Casks/transcripted.rb
+++ b/Casks/transcripted.rb
@@ -1,6 +1,6 @@
 cask "transcripted" do
-  version "1.1.12"
-  sha256 "547b697195c39d7819a06ba9e80874ad6a40e5611668a463849034c1301fe5c7"
+  version "1.1.13"
+  sha256 "a679ef073ba92e3669136470b0126d0aa0f6db386e93d8d089dc1da1a8baccbe"
 
   url "https://github.com/r3dbars/transcripted/releases/download/v#{version}/Transcripted-#{version}.dmg",
       verified: "github.com/r3dbars/transcripted/"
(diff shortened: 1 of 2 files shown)


r3dbars:
## Post-merge review

**Verdict**: APPROVE

**Summary**: Straightforward release metadata update for v1.1.13. The Homebrew cask version and SHA256 are bumped correctly, and the new Sparkle appcast entry has all required fields (version, pubDate, download URL, edSignature, minimumSystemVersion). The core change is correct and does what the PR description says.

**Blockers**: None.

**Should fix** (for next release metadata PR):

- **Missing trailing newline in `docs/appcast.xml`** — The file now ends without a newline (`\ No newline at end of file` in the diff). Most tools and POSIX conventions expect a trailing newline. Worth restoring in a follow-up.

- **Gratuitous reformatting inflates the diff** — The entire `appcast.xml` was re-indented from 2-space to 4-space, turning a ~12-line content change into a 121-addition / 110-deletion diff. This makes it harder to review what actually changed and pollutes `git blame`. Next time, either reformat in a separate commit or keep existing indentation.

**Nits**:

- The new v1.1.13 entry uses `sparkle:fullReleaseNotesLink` while most other entries (1.1.12, 1.1.11, 1.1.10, etc.) use `sparkle:releaseNotesLink`. Both are valid Sparkle elements, and v1.0.7 also uses `fullReleaseNotesLink`, but the inconsistency is worth noting. Sparkle treats them slightly differently (inline vs. external browser).

- `standalone="yes"` was added to the XML declaration. This is valid but unnecessary — the document doesn't reference an external DTD. Harmless, just unexplained churn.

**What looks good**:

- Cask update is clean: version string, SHA256, and the `livecheck` / `verified` URL all line up correctly.
- The new appcast item has proper `sparkle:edSignature`, correct DMG URL pattern, and `minimumSystemVersion: 26.0` consistent with recent releases.
- All existing appcast entries are preserved with no content changes (only whitespace).


---

PR #421, opened 2026-04-21

r3dbars:
### Logic / lifecycle bug
- Fixed a stale meeting-capture start timeout in `MeetingCaptureBridge`.
- Before this change, a timeout task from an earlier start attempt could fire during a later recording start if the earlier timeout window had not elapsed yet. That could incorrectly stop a fresh recording and surface a false timeout failure.

## What I changed
- Scoped the timeout to a specific start attempt with an attempt ID.
- Cancel and clear the timeout task when the start attempt resolves successfully or fails.
- Prevent older timeout tasks from touching newer start continuations.

## Validation
- `bash build.sh`
- `swift test --filter TranscriptionPipelineHelpersTests`

## Notes
- The requested `xcodebuild -project Transcripted.xcodeproj -scheme Transcripted -configuration Debug build` command was not runnable in this repo snapshot because `Transcripted.xcodeproj` is not present here, so validation used the repo's supported build entrypoints instead.

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -44,6 +44,8 @@ final class MeetingCaptureBridge: ObservableObject {
     /// time `startRecording` runs so back-to-back sessions do not leak state.
     private var completionContinuation: CheckedContinuation<(micURL: URL?, systemURL: URL?), Never>?
     private var startContinuation: CheckedContinuation<Bool, Never>?
+    private var startAttemptID: UUID?
+    private var startTimeoutTask: Task<Void, Never>?
 
     init(audio: Audio = Audio()) {
         self.audio = audio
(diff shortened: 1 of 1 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR fixes a real lifecycle bug where a stale timeout `Task` from an earlier `startRecording()` attempt could fire during a later recording session, falsely stopping a fresh capture and surfacing a spurious timeout error. The fix scopes each timeout to an attempt UUID and stores/cancels the timeout task explicitly. The approach is correct, minimal, and well-contained to a single file.

**Blockers**: None.

**Should fix**:
- The 4-line cleanup sequence (`startContinuation = nil` / `startAttemptID = nil` / `startTimeoutTask?.cancel()` / `startTimeoutTask = nil`) is repeated verbatim in the `.ready` case (lines 157–160), the `.failed` case (lines 163–166), and partially in the timeout closure (lines 91–93). A small private helper like `cleanupStartAttempt()` would reduce duplication and make the cleanup contract explicit. Not urgent for a single-file fix, but worth considering if this area gets touched again.

**Nits**:
- Inside the timeout closure (line 93), the task sets `self.startTimeoutTask = nil` — the task nils its own reference. This is safe because everything is `@MainActor` (no concurrent mutation), and the attempt-ID guard already prevents stale execution. But it reads a little odd. If the cleanup were extracted to a helper, this would naturally resolve itself.

**What looks good**:
- The attempt-ID pattern is the right tool for this problem — simple, no over-engineering, easy to reason about.
- `[weak self]` on the timeout task prevents retain cycles. The guard at the top of the closure handles the nil-self case cleanly.
- All three resolution paths (timeout fires, ready, failed) consistently clear all related state. No dangling continuations or leaked tasks.
- The `startTimeoutTask?.cancel()` at the top of `startRecording()` before resuming the old continuation is a nice belt-and-suspenders detail — even though the attempt-ID guard would catch it, explicit cancellation avoids unnecessary sleeping tasks.
- Thread safety is solid: the class is `@MainActor`, the timeout task is `@MainActor`, and `finishPendingStartAttemptIfPossible()` is called from main-thread Combine sinks. No data races possible.

---

PR #425, opened 2026-04-21
Harden Parakeet startRecording failure recovery

r3dbars:
## Summary
- harden `ParakeetEngine.startRecording()` failure handling around audio format and `AVAudioEngine.start()` failures
- keep format-ready state consistent on start failure and reset engine state cleanly
- route retry decisions through a test-covered failure policy (`ParakeetStartRecordingFailurePolicy`)
- add BET-88 QA checklist with live log-monitor commands

## Verification
- `bash build.sh`
- `bash run-tests.sh`

## Manual Validation (Required)
- Run `docs/qa-parakeet-start-failure-smoke.md` on real hardware (Bluetooth reconnect + sleep/wake scenarios)

## Commits
- `273a133`
- `7554c1f`


r3dbars:
QA handoff for BET-88 is ready.\n\nPlease run  on real hardware (Bluetooth reconnect + sleep/wake scenarios) and report pass/fail here.\n\nKey checks:\n- no retry thrash / no stuck recording state after start failures\n- dictation recovers after transient device instability\n- Parakeet start-failure events include  context\n\nIf this passes, I’ll close/merge immediately. If it fails, I’ll patch follow-up in this same PR.

r3dbars:
Correction to my previous comment (formatting got mangled):\n\nQA handoff for BET-88 is ready.\n\nPlease run: docs/qa-parakeet-start-failure-smoke.md\non real hardware (Bluetooth reconnect + sleep/wake scenarios), then report pass/fail here.\n\nKey checks:\n- no retry thrash / no stuck recording state after start failures\n- dictation recovers after transient device instability\n- Parakeet start-failure events include is_recovery_attempt context\n\nIf this passes, I’ll close/merge immediately. If it fails, I’ll patch follow-up in this same PR.

r3dbars:
## PR Review: Harden MCP path validation + add ops health probes

**Verdict**: REQUEST CHANGES

**Summary**: This PR bundles three distinct changes: (1) a well-structured `ParakeetStartRecordingFailurePolicy` extraction that controls retry behavior when audio engine start fails, (2) a solid security hardening of `indexSingleFile` in the MCP server to validate file paths against watched roots before indexing, and (3) new ops tooling (health probes + Codex memory indexer). The Swift code is clean and well-tested. The ops tooling has a functional bug in `health-probe.sh` that breaks the `all` mode.

---

- **`scripts/ops/health-probe.sh`: `skip_lane()` and the probe functions use `exit 0` instead of `return 0`** — When `run_all()` calls `probe_sentry`, `probe_posthog`, etc., any `skip_lane()` call (or the `exit 0` in `probe_github` lines 25/31) terminates the **entire script**, not just the current lane. So `bash health-probe.sh all` on a machine missing Sentry creds will silently exit before ever checking PostHog or Cloudflare. Every `exit 0` inside the probe functions and in `skip_lane()` should be `return 0` for `run_all` to work as intended.

- **`run_all()` only reports the last failed lane** (line 190: `failed_lane="$lane"` overwrites). If both sentry and cloudflare fail, only cloudflare is reported. Consider accumulating: `failed_lane="$failed_lane $lane"`.

---

- **Branch name mismatch**: `docs/claude-md-sync-20260417` doesn't reflect the actual content (security hardening + ops tooling + Parakeet failure policy). This makes the git log harder to navigate later. Consider whether this should have been split into 2-3 focused PRs — the MCP security fix, the Parakeet failure policy, and the ops tooling are independent changes.

- **`build-codex-memory-index.py` at 1,465 lines**: This is a substantial ops script committed without any tests. The redaction logic (`sanitize_text`, token patterns) and the session parsing logic would benefit from at least a basic test file, especially since a regression in redaction could leak secrets into the index output.

- **Undocumented behavioral change in `ParakeetEngine` audioEngineStartFailed path**: The PR adds `audioEngine.stop()` and `isEnginePrewarmed = false` to the engine-start-failure catch block (around line 943-944), which were **not** present in the original code. If the engine `.start()` throws, the original code only cleaned up the tap and returned false — it never explicitly stopped the engine or reset the prewarm flag. This is actually a **good fix** (the engine may be in a partial state after a failed start), but the PR description should call it out since it's a behavioral change beyond just "extract policy."

---

### Nits (take or leave)

- `ParakeetStartRecordingFailurePolicy.action(for:isRecoveryAttempt:)` currently returns `markFormatUnready: true` for all cases — the `if startFailureAction.markFormatUnready` check in `ParakeetEngine` is always true. The structure is fine for future extensibility, but be aware it reads like there's a branch that returns false when there isn't one yet.

- The `switch` in the policy collapses both enum cases into one arm (`case .invalidAudioFormat, .audioEngineStartFailed:`). This is fine as scaffolding, but if they never diverge, the enum + switch is more ceremony than a simple boolean parameter would be.

- `docs/ops-credentials.md` line 7: The PostHog probe command in the table is a very long single-line `curl` command. Consider moving it to a code block below the table for readability.

---

### What looks good

- **MCP path validation is solid security work.** The `indexSingleFile` change correctly validates incoming URLs against allowed roots using `standardizedFileURL`, delegates to the existing `PathSecurity.validateExistingFile` for symlink resolution, and properly handles all three outcomes (valid → reindex, missing → remove, invalid → log + remove). The symlink-escape test (`testIndexSingleFileRejectsSymlinkEscape`) is well-constructed.

- **`ParakeetStartRecordingFailurePolicy` is a clean extraction.** Pure value types, no side effects, testable in isolation. All 4 combinations (2 failure reasons × 2 recovery states) are covered in tests. The key invariant — recovery attempts must not chain extra retries — is clearly expressed.

- **The Python indexer's redaction coverage is thorough.** AWS keys, GitHub PATs, Slack tokens, Bearer tokens, emails, and `/Users/*` paths are all redacted. The `--allow-local-model-calls` opt-in gate for MLX summarization is a good safety rail. The 3-failure auto-disable for MLX is practical.

- **Enhanced EventReporter context in failure paths.** Adding `audio_device`, format details, and `is_recovery_attempt` to both failure events makes debugging device-churn issues much more tractable.

- **QA smoke doc (`qa-parakeet-start-failure-smoke.md`)** is well-written — specific scenarios, clear expected behavior, practical pass criteria.

---

r3dbars:
@r3dbars quick ping on BET-88: PR is ready for manual QA validation using docs/qa-parakeet-start-failure-smoke.md. Once you post pass/fail, I’ll immediately either merge (pass) or patch follow-up (fail).

r3dbars:
Post-merge QA tracking moved to #428 to avoid repeated polling in this PR thread.

Active gate:
- #428 (assigned to @r3dbars, labeled `qa-needed`)
- Run `docs/qa-parakeet-start-failure-smoke.md` and post PASS/FAIL there.

Closure rule:
- PASS => close #428 and mark BET-88 complete in Paperclip tracking
- FAIL => open/link follow-up fix from #428

PR #426, opened 2026-04-21
Merge audio start recovery hardening

r3dbars:
Implements the resolved audio start recovery hardening from #416 after reconciling it with the already-merged Parakeet start failure recovery work from #425.\n\nVerification:\n- bash build-deps.sh --force\n- bash build.sh\n- bash run-tests.sh\n- bash run-integration-smoke.sh\n- swift test\n- swift test in Tools/TranscriptedCLI\n- swift test in Tools/TranscriptedMCP\n- swift build in Tools/TranscriptedQA\n- python3 -m py_compile scripts/ops/build-codex-memory-index.py\n- bash -n scripts/ops/health-probe.sh scripts/ops/nightly-transcripted-archive-miner.sh

### DIFF
--- a/Sources/Meeting/FailedMeetingPresentation.swift
+++ b/Sources/Meeting/FailedMeetingPresentation.swift
@@ -6,6 +6,7 @@ enum FailedMeetingPresentation {
         from failed: FailedTranscription,
         isRetrying: Bool
     ) -> MeetingSessionController.FailedMeetingItem {
+        let failureKind = MeetingFailureKind.classify(message: failed.errorMessage)
         let copy = MeetingFailureCopy.make(
             forMessage: failed.errorMessage,
             shortErrorMessage: failed.shortErrorMessage,
@@ -18,6 +19,7 @@ enum FailedMeetingPresentation {
             title: copy.title,
             detail: copy.detail,
             meta: meta(for: failed, isRetrying: isRetrying),
+            failureKind: failureKind,
             isRetryable: failed.isRetryable,
             isRetrying: isRetrying,
             hasAudioFiles: failed.audioFilesExist()
--- a/Sources/Meeting/MeetingFailureCopy.swift
+++ b/Sources/Meeting/MeetingFailureCopy.swift
@@ -20,12 +20,10 @@ struct MeetingFailureCopy: Equatable {
             )
         }
 
-        if message.contains("recording too short")
-            || message.contains("at least")
-            || message.contains("invalid audio data") {
+        if MeetingFailureKind.isRecordingTooShortMessage(message) {
             return MeetingFailureCopy(
-                title: "Recording was too short",
-                detail: "Transcripted needs at least a second of audio. Keep the meeting running a little longer, then retry."
+                title: "Recording ended too soon",
+                detail: "Nothing broke - there just was not enough audio to transcribe. Record at least two seconds before stopping."
             )
         }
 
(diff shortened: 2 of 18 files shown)


r3dbars:
## Post-merge review

**Verdict**: APPROVE

**Summary**: This PR hardens audio engine start recovery in `ParakeetEngine` by adding a retry loop with full graph reset, Sentry report throttling, and richer failure diagnostics. It also tightens the `recordingTooShort` failure classifier to require audio/recording context (fixing false positives from unrelated "at least" language), softens the UX for short recordings (timer icon + idle dot instead of error styling), and adds the `finalizeBackgroundTranscriptionStateIfNeeded()` call on the `.failed` status path. Test coverage is thorough across all new policies.

**Blockers**: None.

**Should fix**:

- **Duplicated `isRecordingTooShortMessage` across module boundary**: `MeetingFailureKind.isRecordingTooShortMessage` (normalizes internally) and `FailedTranscription.isRecordingTooShortMessage` (expects pre-normalized input) contain identical matching logic but behave differently on raw vs. normalized input. Because they live in `Meeting` vs. `TranscriptedCore`, sharing isn't trivial — but if someone adds a new pattern to one and forgets the other, they'll silently disagree. Consider a doc comment on each noting the other's existence, or extracting the pattern list into a shared constant if the library boundary allows it.

**Nits**:

- **Double normalization in `classify` → `isRecordingTooShortMessage`**: `classify()` lowercases and trims, then passes the result to `isRecordingTooShortMessage()` which lowercases and trims again. Harmless (idempotent) but slightly wasteful. The method needs to normalize because `MeetingFailureCopy.make` calls it with raw input, so this is the right trade-off — just noting it.

- **`markStartFailed()` vs `markFormatUnready()`**: Both set `inputFormatReady = false` with identical bodies. `markStartFailed` communicates better intent at call sites, so it's worth keeping as a distinct method, but a one-line doc comment noting the semantic difference from `markFormatUnready` (doesn't imply a device-change context) would help future readers.

**What looks good**:

- The retry loop design is clean: `for attempt in 1...maxAttempts` with `continue`/`break` is easy to follow, and `resetAudioGraphAfterStartFailure()` fully tears down (tap + engine stop + reset + prewarm flag) before retrying.
- `ParakeetAudioStartRecoveryPolicy` and the Sentry throttle are pure-logic, fully testable, and injected via parameters with sensible defaults. No hidden state.
- The `isRecordingTooShortMessage` tightening is the right fix — requiring "audio" or "recording" context alongside "at least" phrases prevents the false-positive on messages like "Upload failed after at least one retry."
- `removeRecordingTap()` with the `inputTapInstalled` guard and `force` parameter centralizes all 6 call sites that previously called `audioEngine.inputNode.removeTap(onBus: 0)` directly, eliminating the double-remove crash risk.
- The UX changes (timer icon, `.dotIdle` color for short recordings) correctly distinguish "nothing broke" from actual errors — users shouldn't be alarmed by short captures.
- The `SentryPayloadSanitizer` test verifies that new diagnostic keys (`attempt`, `start_mode`, `status_domain`, `status_code`) pass through while PII keys (`audio_device`, `error`, `source_app`) are stripped. Privacy posture is solid.
- Test coverage is excellent: 54 new test lines for recovery state/policy, 16 for failure classification, 32 for sanitizer, 20 for retryability, and regression tests for the overmatch fix.

---

Issue #428, opened 2026-04-21
Post-merge manual QA audit for BET-88 Parakeet startRecording failure path


r3dbars:
Blocked on owner action
- Unblock owner: @r3dbars
- Unblock action: post one top-level QA result comment in this issue where first non-empty line is exactly one of:
  - `PASS`
  - `PASS: <summary>`
  - `FAIL`
  - `FAIL: <summary>`

Context
- BET-88 implementation merged via PR #425 on 2026-04-21 (merge commit 17bee15924e988f9116d0c526d35a93a76438c31).
- Manual QA pass/fail evidence was not posted before merge.

Why this issue exists
- Preserve an explicit verification trail for the startRecording failure hardening after merge.
- Avoid repeated status polling loops.

Owner
- @r3dbars (QA)

Required checklist
1. Run `docs/qa-parakeet-start-failure-smoke.md` on real hardware.
2. Cover Bluetooth reconnect churn and sleep/wake scenarios.
3. Capture pass/fail result in this issue.
4. If fail, include reproduction notes and attach relevant `events.jsonl` excerpts (sanitized).

Automation trigger format (required)
- The first non-empty line of your top-level QA result comment must be one of:
  - `PASS`
  - `PASS: <summary>`
  - `FAIL`
  - `FAIL: <summary>`

Acceptance
- A clear PASS or FAIL is posted.
- If FAIL, a follow-up bug is opened and linked here.

Links
- PR: https://github.com/r3dbars/transcripted/pull/425
- Checklist: docs/qa-parakeet-start-failure-smoke.md

r3dbars:
@r3dbars quick reminder: BET-88 is waiting on the post-merge manual QA result here.

Please run `docs/qa-parakeet-start-failure-smoke.md` and post **PASS** or **FAIL** in this issue.

- PASS => I’ll close this issue and mark BET-88 complete in Paperclip tracking.
- FAIL => I’ll open and land the follow-up fix linked from this issue.

r3dbars:
Status check (2026-04-21): still waiting on QA verification for BET-88.

Blocker
- Owner: `@r3dbars` (QA)
- Action: run `docs/qa-parakeet-start-failure-smoke.md` and post a single **PASS** or **FAIL** comment in this issue.

Resolution path
- If **PASS**: Codex Operator closes #428 and marks BET-88 complete in Paperclip.
- If **FAIL**: Codex Operator opens/lands the linked follow-up fix and references it here.

[...]

PR #439, opened 2026-04-21
ops: add BET-88 QA gate auto-close workflow on main

r3dbars:
## Summary
- add `.github/workflows/qa-gate-auto-close-bet88.yml` on default branch (`main`)
- auto-close issue #428 when a PASS comment is posted

## Why
The previous merge landed this workflow on `docs/claude-md-sync-20260417`, but GitHub Actions only picks workflows from the default branch for live issue events.

## Safety
- scoped to issue #428 only
- ignores bot comments
- does not auto-close on FAIL

## Related
- #428

### DIFF
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -1,0 +1,56 @@
+name: BET-88 QA Gate Auto Close
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+jobs:
+  close_on_pass:
+    if: >-
+      ${{
+        github.event.issue.number == 428 &&
+        github.actor != 'github-actions[bot]'
+      }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Detect gate keyword
+        id: detect
+        shell: bash
+        run: |
+          set -euo pipefail
+          body="$(jq -r '.comment.body // ""' "$GITHUB_EVENT_PATH")"
+
+          if echo "$body" | grep -Eiq '(^|[[:space:]])PASS([[:space:]]|$)'; then
+            echo "status=pass" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


r3dbars:
## Post-merge review

**Verdict**: APPROVE

**Summary**: Adds a GitHub Actions workflow to auto-close issue #428 (BET-88 QA gate) when a human posts a comment containing "PASS". This is a re-land onto `main` because `issue_comment` workflows only trigger from the default branch. Clean, well-scoped ops automation.

**Blockers**: None.

**Should fix**:
- **Cleanup after #428 resolves**: This workflow fires (but no-ops at the job `if`) on every `issue_comment` event repo-wide. Once #428 is closed and the QA gate is done, delete this file to avoid unnecessary workflow triggers. Consider adding a reminder or follow-up task.

**Nits**:
- **`echo` vs `printf` for untrusted input** (line 27–29): `echo "$body"` can misinterpret comment bodies starting with `-n`, `-e`, or `-E` as flags. `printf '%s\n' "$body" | grep ...` is more robust. Extremely unlikely to matter for this use case, but worth noting as a habit.
- **SHA pinning on `actions/github-script`**: `@v7` (major tag) is standard practice but a full SHA pin would be more resilient to supply-chain attacks. Low risk given this workflow's narrow scope.
- **Dead FAIL output**: The `status=fail` output is detected but never consumed by a downstream step. Harmless (useful for logs or future extension), just noting it exists.

**What looks good**:
- Minimal `permissions` scope — `issues: write` only, no unnecessary grants
- Bot-loop prevention via `github.actor != 'github-actions[bot]'`
- Word-boundary regex `(^|[[:space:]])PASS([[:space:]]|$)` prevents false positives from words like "bypass" or "password"
- `set -euo pipefail` in the bash step
- PR description clearly explains *why* this re-land is needed (default branch requirement for `issue_comment` triggers)
- Single-issue scope keeps blast radius to zero

---

PR #443, opened 2026-04-21
[codex] BET-88 QA gate: auto-create follow-up on FAIL

r3dbars:
## Summary
- extend BET-88 QA gate comment workflow to handle top-level FAIL
- keep existing PASS auto-close behavior unchanged
- on FAIL, open or reuse a deduplicated follow-up issue and post the link back on #428

## Why
BET-88 is QA-gated and already automated for PASS closeout. This adds durable automation for the FAIL branch so follow-up work is immediately tracked without manual triage lag.

## Test
- workflow logic inspection in `.github/workflows/qa-gate-auto-close-bet88.yml`

### DIFF
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -8,7 +8,7 @@ permissions:
   issues: write
 
 jobs:
-  close_on_pass:
+  handle_gate_comment:
     if: >-
       ${{
         github.event.issue.number == 428 &&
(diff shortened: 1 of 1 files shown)


r3dbars:
## PR Review: BET-88 QA gate — auto-create follow-up on FAIL

**Verdict**: APPROVE

**Summary**: This PR extends the existing BET-88 QA gate workflow to handle the FAIL branch — when someone comments FAIL on issue #428, it creates (or reuses) a deduplicated follow-up issue and posts a link back. The change is focused, follows the patterns already established by the PASS branch, and the job rename from `close_on_pass` to `handle_gate_comment` is the right call now that the job handles both outcomes.

---

None.

---

- **Search API eventual consistency**: GitHub's search index can lag by seconds to minutes. If two FAIL comments land in quick succession, the search in `github.rest.search.issuesAndPullRequests` may not yet index the first follow-up issue, creating a duplicate. This is a known GitHub API limitation and there's no clean fix without external state — but it's worth acknowledging. A pragmatic mitigation would be to add a label (e.g., `qa-followup`) to the created issue and search by label instead of title, since label filtering via `issues.listForRepo` is consistent (not search-indexed). Low probability for a gate on a single issue, so not a blocker.

---

**Nits** (take or leave):

- **Workflow name**: The workflow is still named `BET-88 QA Gate Auto Close` (line 1), but it now also opens follow-up issues. Consider updating to something like `BET-88 QA Gate Handler` for accuracy.
- **No labels on the follow-up issue**: Adding a label like `qa-followup` or `BET-88` to the created issue would make filtering easier and could also serve the deduplication improvement mentioned above.
- **Hardcoded `#428` in body template** (line 89): The rest of the script correctly uses the dynamic `issue_number`, but the body includes a hardcoded `"- Link validation evidence back to #428"`. Consistent with the top-level guard (`github.event.issue.number == 428`), so not wrong — just worth noting if this pattern is ever templated for other gates.

---

**What looks good**:

- **Deduplication via marker title** is a solid pattern for keeping follow-up issues from piling up.
- **`filter(Boolean)`** to strip empty lines when `failComment` is absent — clean approach.
- **Defensive optional chaining** on `context.payload.comment?.html_url` is correct even though the `issue_comment` event always provides it.
- **Bot exclusion guard** (`github.actor != 'github-actions[bot]'`) on the job prevents infinite loops from the bot's own auto-comments. Already existed, but glad it's there.
- **The referenced doc** (`docs/qa-parakeet-start-failure-smoke.md`) exists in the repo, so the follow-up issue body points to a real checklist.
- **Permissions are minimal** — `issues: write` is exactly what's needed, nothing more.

---

PR #444, opened 2026-04-21
[codex] Fix BET-88 QA gate false-positive PASS matching

r3dbars:
## Summary
- fix the BET-88 QA gate matcher in `.github/workflows/qa-gate-auto-close-bet88.yml`
- require explicit first non-empty line gate declaration:
  - `PASS` or `PASS: ...`
  - `FAIL` or `FAIL: ...`
- prevent explanatory text like `PASS => ...` from triggering auto-close

## Why
`#428` was auto-closed by a false positive from explanatory text, not a real QA result. This patch makes automation strict and safe.

## Validation
- manually inspected matcher logic in workflow step `Detect gate keyword`
- reopened `#428` and documented the false-positive + fix path

### DIFF
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -8,7 +8,7 @@ permissions:
   issues: write
 
 jobs:
-  close_on_pass:
+  handle_gate_comment:
     if: >-
       ${{
         github.event.issue.number == 428 &&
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review

**Summary**: This PR fixes a false-positive in the BET-88 QA gate workflow where explanatory text containing "PASS" (like `PASS => ...`) triggered an auto-close on #428. The fix restricts matching to an explicit `PASS` or `FAIL` declaration on the first non-empty line of a comment, and adds a new FAIL-handling path that auto-creates follow-up issues. Clean, focused fix for a real problem.

---

**Blockers**: None.

---

1. **`filter(Boolean)` removes intentional blank-line separators in the FAIL follow-up issue body** (line 103). The array uses empty strings `""` both as paragraph-break separators and as the falsy value from the conditional `failComment` ternary. `filter(Boolean)` removes *all* of them, collapsing the intended paragraph structure. A targeted fix:
   ```js
   // Instead of filter(Boolean), only filter the conditional line:
   .filter(line => line !== false)
   // ...and use `false` instead of `""` in the ternary:
   failComment ? `Source FAIL comment: ${failComment}` : false,
   ```

---

**Nits** (take or leave):

1. **`echo` vs `printf` for first_line** (line 39): `echo "$first_line"` can misinterpret lines starting with `-e`, `-n`, or `-E` as flags on some shells. `printf '%s\n' "$first_line"` is the safer portable pattern. Not a practical risk here since the grep anchors to `^PASS`/`^FAIL`, but worth being consistent — the line above already uses `printf`.

2. **Case-insensitive matching**: The `-i` flag on grep means `pass`, `Pass`, `PASS` all trigger the gate. Probably intentional for usability, but worth documenting in the comment block if so.

3. **GitHub search indexing delay**: The deduplication search in the FAIL handler (line 80) depends on GitHub's search index, which can lag by seconds to minutes. Rapid consecutive FAIL comments could create duplicate follow-up issues. Low risk given this is scoped to a single gate issue, but worth noting.

---

**What looks good**:

- Reading comment body via `jq -r` from `$GITHUB_EVENT_PATH` instead of interpolating into shell — correct security pattern for GitHub Actions.
- The first-line extraction pipeline (`sed` → `awk` → `sed`) is robust: handles `\r\n`, skips blank lines, trims whitespace.
- Job rename from `close_on_pass` to `handle_gate_comment` accurately reflects the expanded scope.
- Bot actor filter (`github.actor != 'github-actions[bot]'`) prevents infinite comment loops.
- Minimum permissions (`issues: write` only).
- The FAIL handler's deduplication-via-search-before-create is a nice touch for idempotency.

---

PR #445, opened 2026-04-21
[codex] Clarify BET-88 QA gate result comment format

r3dbars:
## Summary
- update `docs/qa-parakeet-start-failure-smoke.md` with explicit gate-result comment format for issue `#428`
- document accepted first-line forms:
  - `PASS`
  - `PASS: <summary>`
  - `FAIL`
  - `FAIL: <summary>`
- add copy/paste PASS and FAIL comment examples

## Why
BET-88 gate automation now parses a strict first-line declaration. This doc update removes ambiguity for QA and prevents accidental format mismatches.

### DIFF
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -8,7 +8,7 @@ permissions:
   issues: write
 
 jobs:
-  close_on_pass:
+  handle_gate_comment:
     if: >-
       ${{
         github.event.issue.number == 428 &&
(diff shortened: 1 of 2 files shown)


r3dbars:
## Post-merge review

**Summary**: This PR tightens the BET-88 QA gate automation to only match explicit first-line `PASS`/`FAIL` declarations (preventing false triggers from body text), and adds a new FAIL path that automatically creates or reuses a follow-up tracking issue. The documentation update in the smoke test doc is clear and well-structured. Overall this is clean, well-scoped work.

**Note**: This was merged ~7 seconds after opening with no review. For CI/workflow changes that auto-close issues and auto-create new ones, a review pass is worth the extra minute.

---

### Blockers
None.

### Should fix

1. **GitHub Search API indexing lag (race condition on duplicate FAIL comments)**
   In `.github/workflows/qa-gate-auto-close-bet88.yml` lines 80-86: the `search.issuesAndPullRequests` API has a known indexing delay (seconds to minutes). If two FAIL comments land on `#428` in quick succession, both workflow runs could see zero search results and each create a duplicate follow-up issue. Low probability for a single QA gate issue, but worth noting. A mitigation would be to search using `list` with a title filter on the `issues` REST endpoint instead of the search API, since `list` reads from the primary store with no index lag:
   ```js
   const issues = await github.rest.issues.listForRepo({
     owner, repo, state: 'open',
     creator: 'github-actions[bot]',
     per_page: 100
   });
   const existing = issues.data.find(i => i.title === marker);
   ```

2. **Missing `contents: read` in permissions block**
   The workflow declares only `permissions: issues: write`. The `search.issuesAndPullRequests` call and `actions/github-script` both work because `GITHUB_TOKEN` gets implicit read access by default — but if the repo or org ever sets `permissions: {}` as a default, this will break silently. Adding `contents: read` makes the contract explicit.

### Nits

1. **No labels on auto-created follow-up issues** (line 89-103): Adding a label like `qa-gate` or `automated` to the created issue would make filtering and cleanup easier.

2. **Hardcoded `#428` in the follow-up body** (line 102): The body string `"- Link validation evidence back to #428"` is hardcoded while the rest of the step uses `issue_number` dynamically. Should be `` `- Link validation evidence back to #${issue_number}` `` for consistency, even though in practice `issue_number` is always 428 here.

3. **Commit ordering**: The three commits read bottom-to-top as "add FAIL workflow → tighten matcher → clarify docs", which is a logical progression. Clean history — nice.

### What looks good

- **First-line extraction pipeline** (lines 26-31): The `sed → awk → sed` pipeline for CR-stripping, first-non-blank extraction, and whitespace trimming is robust and handles Windows-style line endings. Good defensive shell.
- **Idempotent follow-up creation**: The search-or-create pattern avoids piling up duplicate issues on repeated FAILs. The `filter(Boolean)` trick for conditional body lines is clean.
- **Bot-comment exclusion**: `github.actor != 'github-actions[bot]'` in the job condition prevents the bot's own auto-comments from re-triggering the workflow. Important and easy to miss.
- **Documentation quality**: The added QA comment format section with copy-paste examples is exactly the right level of detail — prevents format mismatches without over-specifying.

---

PR #447, opened 2026-04-21
[codex] Restrict BET-88 gate workflow to QA owner

r3dbars:
## Summary
- update BET-88 gate workflow job condition to only process comments from `r3dbars`
- keep existing PASS/FAIL parsing and close/follow-up behavior unchanged

## Why
Prevents accidental gate triggers from unrelated commenters on #428 while the issue is in manual QA hold.

### DIFF
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -12,7 +12,8 @@ jobs:
     if: >-
       ${{
         github.event.issue.number == 428 &&
-        github.actor != 'github-actions[bot]'
+        github.actor != 'github-actions[bot]' &&
+        github.event.comment.user.login == 'r3dbars'
       }}
     runs-on: ubuntu-latest
 
--- a/.github/workflows/qa-gate-auto-close-bet88.yml
+++ b/.github/workflows/qa-gate-auto-close-bet88.yml
@@ -8,7 +8,7 @@ permissions:
   issues: write
 
 jobs:
-  close_on_pass:
+  handle_gate_comment:
     if: >-
       ${{
         github.event.issue.number == 428 &&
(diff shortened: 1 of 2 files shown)


r3dbars:
## Post-merge review

**Verdict**: APPROVE

**Summary**: This PR tightens the BET-88 QA gate workflow in three ways: restricts triggers to comments from `r3dbars` only, narrows PASS/FAIL detection from whole-body scan to first-non-empty-line only, and adds a new FAIL path that creates (or reuses) a follow-up tracking issue. It also documents the expected comment format in the QA smoke doc. Overall this is clean, well-scoped work.

**Blockers**: None.

**Should fix** (for a follow-up):
- The `github.actor != 'github-actions[bot]'` condition (line 15) is now redundant — if `comment.user.login == 'r3dbars'`, the actor can never be `github-actions[bot]`. Harmless but worth cleaning up to avoid confusion about what's actually guarding the gate.
- The GitHub Search API has eventual consistency. If two FAIL comments land in quick succession, the `search.issuesAndPullRequests` call (line 81) might not find the just-created follow-up issue, producing duplicates. Low risk given the single-owner gate, but if you ever widen access, consider switching to `issues.listForRepo` with a label filter instead of search — the REST list endpoint is consistent.

**Nits**:
- The workflow `name` (line 1) is still "BET-88 QA Gate Auto Close" but the workflow now also handles FAIL. Consider updating to something like "BET-88 QA Gate" to reflect both paths.

**What looks good**:
- First-line-only matching is a significant hardening over the previous whole-body grep — eliminates false triggers from comments that mention PASS/FAIL in discussion context.
- The shell pipeline (`printf | sed | awk | sed`) handles untrusted comment bodies safely: all variables are properly quoted, patterns are hardcoded, no user input reaches an execution context.
- The `actions/github-script` steps are clean — no comment body text is interpolated into JS template literals, only safe GitHub API objects (`issue.number`, `comment.html_url`).
- Follow-up issue reuse via title marker search is a pragmatic dedup approach.
- The QA doc addition with concrete PASS/FAIL examples is a nice touch — removes ambiguity for anyone running the smoke test.
- Minimal permission scope (`issues: write`) is correct and sufficient.

---

PR #448, opened 2026-04-21
Ignore .claude worktree mirrors in repo command contract scan

r3dbars:
## Summary
- exclude `.claude/worktrees/` from `RepoCommandContractTests` repo scan
- prevents local/mirrored worktree copies from tripping the `Transcripted.xcodeproj` live-doc assertion

## Why
`run-tests.sh` for BET-88 verification was failing on false positives from `.claude/worktrees/*` script mirrors, not from live repo docs/scripts.

## Verification
- `bash run-tests.sh` (633 passed, 0 failed)
- `bash build.sh`

### DIFF
--- a/Tests/RepoCommandContractTests.swift
+++ b/Tests/RepoCommandContractTests.swift
@@ -82,6 +82,7 @@ private func shouldDescendInto(_ url: URL, root: URL) -> Bool {
     let relativePath = url.path.replacingOccurrences(of: root.path + "/", with: "")
     let excludedPrefixes = [
         ".build/",
+        ".claude/worktrees/",
         ".deps-build/",
         ".git/",
         ".swiftpm/",


r3dbars:
## PR Review (post-merge)

**Verdict**: APPROVE

**Summary**: Single-line fix adding `.claude/worktrees/` to the excluded-prefix list in `RepoCommandContractTests.shouldDescendInto()`. This prevents Claude Code worktree mirrors from triggering false positives when the test scans for stale `Transcripted.xcodeproj` / `xcodebuild` references. Clean and correct.

**Blockers**: None.

**Should fix**: None.

**Nits**: None. The entry is alphabetically sorted with the rest of the list, uses the trailing-slash convention consistently, and the scope is appropriately narrow (`.claude/worktrees/` rather than all of `.claude/`) since other `.claude/` files like `settings.json` could legitimately be scanned.

**What looks good**:
- The fix targets only the worktree mirror directory, not the entire `.claude/` tree — good surgical scoping.
- `.claude/` is already in `.gitignore`, so this was purely a filesystem-walk issue in the test harness. The PR description correctly identifies the root cause.
- Verification section in the PR body shows all three check commands were run. Nice.

Solid fix. Nothing to flag.

---

PR #451, opened 2026-04-21
Add one-shot BET-88 QA gate closeout helper

r3dbars:
## Summary
- add `scripts/ops/qa-gate-closeout.sh`
- wrap `qa-gate-check.sh` with explicit closeout messaging for `PASS` / `FAIL` / `PENDING`
- document the helper in `scripts/README.md`

## Why
BET-88 is externally blocked on manual QA evidence in `#428`. This helper gives one deterministic command that reports closeout readiness and always prints unblock owner/action when still pending.

## Verification
- `bash scripts/ops/qa-gate-closeout.sh r3dbars/transcripted 428 r3dbars`
  - output includes:
    - `status: PENDING`
    - `UNBLOCK_OWNER: @r3dbars`
    - `UNBLOCK_ACTION: post top-level comment ... PASS / FAIL`
  - exit code: `3`

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -39,6 +39,9 @@ not have to carry the full operational logic:
 - `scripts/ops/qa-gate-check.sh` — one-shot check for the BET-88 QA gate comment on `#428` using the same strict owner + first-line PASS/FAIL rules as the auto-close workflow
   - Usage: `bash scripts/ops/qa-gate-check.sh [repo] [issue_number] [owner_login]`
   - Returns JSON and exits `0` for `pass`/`fail`, `3` for `PENDING`
+- `scripts/ops/qa-gate-closeout.sh` — closeout wrapper around `qa-gate-check.sh` that prints explicit unblock owner/action when status is still pending
+  - Usage: `bash scripts/ops/qa-gate-closeout.sh [repo] [issue_number] [owner_login]`
+  - Returns `0` for pass/fail closeout-ready, `3` when still blocked/pending
 - `scripts/ops/build-codex-memory-index.py` — build a safe metadata-only index from local Codex session archives for Transcripted memory briefs
   - Usage: `python3 scripts/ops/build-codex-memory-index.py --verbose`
   - Writes:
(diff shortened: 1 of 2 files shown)


r3dbars:
## Post-merge review

**Verdict**: APPROVE

**Summary**: This PR adds a thin closeout wrapper (`scripts/ops/qa-gate-closeout.sh`) around the existing `qa-gate-check.sh` to give human-readable unblock guidance when the BET-88 QA gate on #428 is still pending. It's a clean, well-scoped ops utility — 58 lines, proper error handling, good documentation in `scripts/README.md`.

**Blockers**: None.

**Should fix**:

- **`-x` check vs `bash` invocation** (`qa-gate-closeout.sh:17,25`): The guard `[[ ! -x "$CHECK_SCRIPT" ]]` requires the executable bit, but the script is invoked via `bash "$CHECK_SCRIPT"` which doesn't need `+x`. These two lines disagree on intent. If the executable bit matters, invoke directly as `"$CHECK_SCRIPT"` without `bash`. If it doesn't (and it doesn't, since you use `bash`), relax the check to `[[ ! -f "$CHECK_SCRIPT" ]]`. In practice this won't bite because the file is committed as `755`, but the inconsistency could confuse a future reader or break if someone re-clones without preserving file modes.

- **Mixed JSON + plain text on stdout** (`qa-gate-closeout.sh:29`): `echo "$check_output"` dumps raw JSON to stdout, then the closeout/action lines append plain text to the same stream. Any downstream consumer parsing stdout as JSON will break. Consider either sending the JSON to stderr (`echo "$check_output" >&2`) or wrapping everything into a single JSON envelope. For a human-only ops helper this is fine today, but it's a paper cut waiting to happen if anyone pipes this output.

**Nits**:

- `check_output=""` on line 22 is unnecessary — it's immediately overwritten on line 25 after `set +e`. Harmless, but dead initialization.

**What looks good**:

- Clean delegation pattern: the wrapper doesn't duplicate any of the gate-matching logic from `qa-gate-check.sh`.
- All three exit paths (pass/fail at `0`, pending at `3`, unexpected error at `2`) are handled explicitly — no silent failures.
- `set -euo pipefail` with a controlled `set +e` block for capturing the exit code is the right pattern.
- README documentation is clear and consistent with the existing entries.
- No security concerns: all variables are properly quoted, no eval, no uncontrolled expansion.

---

PR #468, opened 2026-04-22

r3dbars:
## Summary
Nightly automated code review fixes for recent transcription/imported-audio changes.

### Transcript metadata parsing
- fixed `TranscriptionTaskManager.populateSavedMetadata` so it reads YAML frontmatter in bounded chunks instead of only the first 2 KB
- this restores saved title, duration, and speaker-count parsing for newer transcripts whose frontmatter grew past 2 KB because of speaker metadata and gap-event details
- added a regression test covering oversized frontmatter parsing

### Imported audio boundary handling
- reject directories/packages during imported-audio preparation instead of treating them like source files
- clean up partially copied scratch files if the secure permission-hardening step fails

### Speaker metadata consistency
- fixed system-speaker `speakerSources` entries to use the same channel-qualified keys as the rest of the transcript metadata pipeline

## Validation
- `xcodebuild -project Transcripted.xcodeproj -scheme Transcripted -configuration Debug build 2>&1` could not run because this checkout does not contain `Transcripted.xcodeproj`
- `swift test --filter TranscriptionTaskManagerMetadataTests`
- `bash build.sh`

### DIFF
--- a/Sources/Meeting/MeetingImportedAudioPreparer.swift
+++ b/Sources/Meeting/MeetingImportedAudioPreparer.swift
@@ -27,6 +27,29 @@ enum MeetingImportedAudioPreparer {
             )
         }
 
+        let sourceAttributes: [FileAttributeKey: Any]
+        do {
+            sourceAttributes = try fileManager.attributesOfItem(atPath: resolvedSourceURL.path)
+        } catch {
+            throw NSError(
+                domain: "MeetingImportedAudioPreparer",
+                code: 1,
+                userInfo: [
+                    NSLocalizedDescriptionKey: "Transcripted couldn't inspect that audio file."
+                ]
+            )
+        }
+
+        guard sourceAttributes[.type] as? FileAttributeType == .typeRegular else {
+            throw NSError(
+                domain: "MeetingImportedAudioPreparer",
+                code: 1,
+                userInfo: [
+                    NSLocalizedDescriptionKey: "Select an audio file, not a folder or package."
+                ]
+            )
+        }
+
         let destinationURL = uniqueScratchURL(
             for: resolvedSourceURL,
             in: scratchDirectory,
@@ -37,6 +60,7 @@ enum MeetingImportedAudioPreparer {
             try fileManager.copyItem(at: resolvedSourceURL, to: destinationURL)
             fileManager.restrictFileToOwnerOnly(at: destinationURL)
         } catch {
+            try? fileManager.removeItem(at: destinationURL)
             throw NSError(
                 domain: "MeetingImportedAudioPreparer",
                 code: 2,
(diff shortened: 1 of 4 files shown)


r3dbars:
## Code Review — PR #468

**Verdict**: APPROVE

**Summary**: This PR fixes three real bugs: a 2 KB YAML frontmatter parsing limit that silently dropped metadata for speaker-heavy transcripts, a miskeyed `speakerSources` dictionary entry for system speakers, and missing input validation + scratch cleanup in the imported-audio preparer. All three changes are correct, well-scoped, and backed by a solid regression test. Clean PR.

---

### Blockers
None.

---

### Should fix

**`readYAMLFrontmatter` reads up to 64 KB for non-YAML files unnecessarily**
`TranscriptionTaskManager.swift:385` — The `data.starts(with: Data("---".utf8))` check is inside the while loop but its result never changes after the first chunk. If a file doesn't start with `---`, the function still reads up to 64 KB before returning `nil`. In practice this only runs on pipeline-generated `.md` files which always have frontmatter, so it won't bite — but an early-exit after the first chunk would be cleaner:

```swift
// After first chunk, bail immediately if no frontmatter opener
if data.count == chunk.count && !data.starts(with: Data("---".utf8)) {
    return nil
}
```

---

### Nits

1. **Duplicate error codes in `MeetingImportedAudioPreparer`** — The "file not found", "couldn't inspect", and "not a regular file" errors all use `code: 1`. If anything ever switches on error codes programmatically, these would be indistinguishable. The messages are distinct so this is fine for user-facing display, but unique codes would be more defensive.

2. **Test multiline string indentation** — `TranscriptionTaskManagerMetadataTests.swift:29-40` — The test content string works correctly (Swift strips leading whitespace matching the closing `"""`), but at a glance it *looks* like the YAML lines are indented. A comment or a differently-structured literal could prevent a future reader from second-guessing whether the frontmatter starts at column 0. Very minor.

---

### What looks good

- **The `speakerSources` key fix** (`TranscriptionPipelineRunner.swift:144`) is a subtle but important correctness fix. The `TranscriptFormatter` has a fallback (`speakerSources[key] ?? speakerSources[mapping.speakerId]`) that was masking this bug in formatted output, but the dictionary was still inconsistently keyed compared to `speakerMappings`. Good catch.

- **The chunked YAML reader** is well-designed. It accumulates data and searches the full buffer each iteration, so closing markers that straddle chunk boundaries are handled correctly. The `defer { try? handle.close() }` cleanup and the `maxBytes` cap are both solid guardrails.

- **Imported-audio directory rejection** is a sensible hardening — `copyItem` on a directory would silently create a deep copy, wasting disk and confusing the pipeline. The `try? fileManager.removeItem(at: destinationURL)` cleanup on copy failure prevents orphaned scratch files.

- **The regression test** directly exercises the failure mode (frontmatter > 2 KB) with realistic padding that mimics speaker metadata growth. Clean setup/teardown with temp directories. Good coverage for the fix.

- **Single commit, focused scope, clear PR description.** This is how automated fix PRs should look.

---

PR #471, opened 2026-04-22
Add GitHub README banner

r3dbars:
## Summary

- Adds a polished Transcripted banner image for the GitHub README.
- Updates the README to use the repo-hosted banner asset instead of the old attachment.

## Validation

- Previewed the generated banner locally at full size and GitHub-style display size.
- Tests not run because this only changes README/image assets.


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR replaces the old GitHub user-attachments banner URL with a repo-hosted PNG at `docs/assets/transcripted-github-banner.png` and switches from an `` HTML tag to standard Markdown image syntax. Clean, minimal change. The banner looks polished and professional.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- The old `` tag specified explicit dimensions (`width="625" height="329"`), which prevents layout shift while the image loads. The Markdown `alt` syntax doesn't support that. On GitHub this is cosmetic at worst, but if you notice a flash of layout reflow on the README you could switch back to an `` tag pointing at the new local path.
- The PNG is ~1.3 MB. Not a problem for a single banner, but worth keeping in mind if more assets land in `docs/assets/` — at some point Git LFS would be worth considering. No action needed now.

**What looks good**:
- Moving from an ephemeral `user-attachments` URL to a repo-committed asset is the right call — the old URL could break silently if GitHub garbage-collects it.
- Image is already covered by `*.png binary` in `.gitattributes`.
- The banner itself (1600×840, RGBA) is sharp, on-brand, and renders well at GitHub's display width.

---

PR #474, opened 2026-04-22

r3dbars:
## Summary

- **Logic bug (STTRouter.swift:148)**: Removed redundant `parakeetEngine.finishExternalTranscription()` call from the `catch` block in `STTRouter.transcribe()`. Swift's `defer` in the `do` block already calls this when the block exits via throw — it runs _before_ control transfers to `catch`. The redundant call was harmless (double-set of a bool + double-clear of an empty buffer) but misleading, suggesting the `defer` was insufficient.

## Review coverage

Reviewed all commits from the last 7 days:
- `fa56789` — BET-88 workflow: no logic, CI-only
- `97b1323` — Parakeet start failure state reset hardening: correct, adds defensive pre-reset before `rebuildAudioEngine`
- `36e1d90` — Microphone recovery hardening: correct, properly introduces `rebuildAudioEngine` as a full engine replacement path
- `504c2fe` — Dictation shortcut modes: correct, CapsLock-as-toggle behavior is intentional

**No threading violations found.** `Audio.swift` and `SystemAudioCapture.swift` are correctly not `@MainActor`. **No TranscriptedCore boundary violations** — no Draft app imports found in `Sources/TranscriptedCore/`.

## Test plan

- [x] `bash run-tests.sh` — 656/656 passed
- [x] `swift test` — 68/68 passed

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Sources/Speech/STTRouter.swift
+++ b/Sources/Speech/STTRouter.swift
@@ -145,7 +145,6 @@ class STTRouter: ObservableObject {
                 return text.isEmpty ? nil : text
             } catch {
                 print("❌ WHISPER | dictation failed: \(error.localizedDescription)")
-                parakeetEngine.finishExternalTranscription()
                 return nil
             }
         }


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR removes a redundant `parakeetEngine.finishExternalTranscription()` call from the `catch` block of `STTRouter.transcribe()`. The `defer` at line 137–139 is scoped to the `do` block, so it already fires before the `catch` block runs on a throw — making the explicit call in `catch` a no-op duplicate. Clean, correct one-liner.

**Blockers**: None.

**Should fix**: None.

**Nits**: None.

**What looks good**:
- The Swift `defer` semantics analysis is spot-on. I verified: `defer` inside a `do` block runs when execution leaves the `do` scope, which happens *before* the `catch` block receives control. The duplicate call was indeed dead code.
- `finishExternalTranscription()` is idempotent (`isTranscribing = false` + `sampleBuffer.removeAll`), so the double call was harmless at runtime — but removing it eliminates a misleading suggestion that the `defer` doesn't cover the error path.
- PR description is thorough: documents every recent commit reviewed, confirms no threading violations, and lists all test results. Well done for an automated nightly review.

---

PR #475, opened 2026-04-22

r3dbars:
## Summary

- **Remove redundant `fileExists` pre-checks** in `MeetingRecordingCleanup.discardFiles` and `DictationTranscriptWriter.hasExistingContent` — both already handle missing-file errors downstream, so the pre-check is a TOCTOU anti-pattern with no safety benefit.
- **Bound `while true` uniqueness loops** in `MeetingTranscriptStyler` (`uniqueAudioDirectoryURL` and `uniqueTranscriptURL`) to a max of 999 iterations with a UUID-based fallback, eliminating the theoretical infinite-loop risk on pathological filesystems.

## What was reviewed but skipped

- `shouldSurfaceMeetingWarmupFailure` flag refactor — functionally correct, too invasive for a sweep
- `MeetingTranscriptStyler.renameAudioDirectoryIfNeeded` fileExists — intentional conditional logic (skip if no audio directory), not a TOCTOU
- `fileEndsWithBlankLine` double FileHandle open — minor overhead on tiny files, not worth the churn
- `provider(forBundleIdentifier:)` relocation to enum — new feature, not a simplify regression
- DictationTranscriptStore early-exit sort — analyzed and confirmed correct (day files are date-named, so newer day ↔ newer entries)

## Test plan

- [x] `bash build.sh` — clean build, signed
- [x] `bash run-tests.sh` — 656/656 passed

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/Sources/Dictation/DictationTranscriptWriter.swift
+++ b/Sources/Dictation/DictationTranscriptWriter.swift
@@ -161,8 +161,7 @@ enum DictationTranscriptWriter {
     }
 
     private static func hasExistingContent(at url: URL) -> Bool {
-        guard FileManager.default.fileExists(atPath: url.path),
-              let values = try? url.resourceValues(forKeys: [.fileSizeKey]),
+        guard let values = try? url.resourceValues(forKeys: [.fileSizeKey]),
               let size = values.fileSize else {
             return false
         }
--- a/Sources/Meeting/MeetingRecordingCleanup.swift
+++ b/Sources/Meeting/MeetingRecordingCleanup.swift
@@ -11,7 +11,6 @@ enum MeetingRecordingCleanup {
         let urls = Set([micURL, systemURL].compactMap { $0 })
 
         for url in urls {
-            guard fileManager.fileExists(atPath: url.path) else { continue }
             do {
                 try fileManager.removeItem(at: url)
                 discarded.append(url)
--- a/Sources/Meeting/MeetingTranscriptStyler.swift
+++ b/Sources/Meeting/MeetingTranscriptStyler.swift
@@ -402,13 +402,14 @@ enum MeetingTranscriptStyler {
         let stem = preferredURL.lastPathComponent
         var suffix = 2
 
-        while true {
+        while suffix <= 999 {
             let candidate = directory.appendingPathComponent("\(stem) \(suffix)", isDirectory: true)
             if !fm.fileExists(atPath: candidate.path) {
                 return candidate
             }
             suffix += 1
         }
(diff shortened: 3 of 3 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR removes two redundant `fileExists` pre-checks (both TOCTOU anti-patterns) and bounds two `while true` uniqueness loops with a 999-iteration cap and UUID fallback. All three changes are small, well-scoped, and correct. The PR description is thorough, with a clear "reviewed but skipped" section that shows good judgment about what belongs in a sweep versus what doesn't.

---

None.

---

- **No test coverage for the new fallback paths.** The UUID fallback in `uniqueAudioDirectoryURL` and `uniqueTranscriptURL` is untested, and there are no tests verifying `hasExistingContent` or `discardFiles` behavior when the file doesn't exist. The existing test suites only cover happy paths. Since the whole point of this PR is tightening edge-case behavior, consider adding at least one test per change to prove the new paths work:
  - `MeetingTranscriptStylerTests`: a test with a mock/temp directory pre-populated to force the UUID fallback
  - `MeetingRecordingCleanupTests`: a test calling `discardFiles` with a URL that doesn't exist on disk, verifying it returns an empty array
  - `DictationTranscriptWriterTests`: a test calling `hasExistingContent` (or exercising it through `save`) on a non-existent URL

---

**Nits** (take or leave):

- **Off-by-one in `uniqueTranscriptURL`**: Because `candidateStem` is updated at the *end* of each loop iteration, the last stem set (`"preferredStem 999"`) is never actually checked before the loop exits. The function tries candidates up to `"preferredStem 998"` then falls through to UUID. Compare with `uniqueAudioDirectoryURL`, which builds the candidate inline and correctly checks `"stem 999"`. Practically irrelevant (999 collisions won't happen), but the two methods are inconsistent — `uniqueAudioDirectoryURL` checks 998 suffixed candidates (2–999), while `uniqueTranscriptURL` checks 997 suffixed candidates (2–998) plus the unsuffixed stem. If you ever want to align them, changing the bound to `suffix <= 1000` in `uniqueTranscriptURL` would match the other method's coverage.

---

**What looks good**:

- The TOCTOU analysis is spot-on. Both `fileExists` calls were genuinely redundant: `url.resourceValues(forKeys:)` throws on missing files (caught by `try?`), and `removeItem` throws into a `catch` that already continues. Removing them is strictly better.
- The "reviewed but skipped" section in the PR description shows disciplined scoping — especially the call-out that `renameAudioDirectoryIfNeeded`'s `fileExists` is intentional conditional logic rather than a TOCTOU.
- Build, unit tests (656/656), and integration smoke all pass.

---

PR #509, opened 2026-04-24
Harden release QA reliability paths

r3dbars:
## Summary
- preserve active dictation and meeting captures during app shutdown
- refresh hotkeys after onboarding permission changes and retarget meeting storage dynamically
- harden MCP/CLI transcript file handling and add focused regression tests
- smooth menu/settings loading states before release

## Verification
- bash build-deps.sh --force
- bash build.sh
- bash run-tests.sh
- swift test
- swift test (Tools/TranscriptedMCP)
- swift test (Tools/TranscriptedCLI)
- swift test (Tools/TranscriptedQA)
- Computer Use smoke of Transcripted Settings/Home/Storage

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -230,6 +230,7 @@ final class MeetingSessionController: ObservableObject {
             speakerStore: services.speakerStore,
             speakerClipsDirectory: storagePaths.speakerClips,
             retainedAudioDirectory: MeetingStoragePaths.audioArchiveFolder,
+            retainedAudioDirectoryProvider: { MeetingStoragePaths.audioArchiveFolder },
             statsStore: statsDatabase
         )
 
@@ -772,6 +773,44 @@ final class MeetingSessionController: ObservableObject {
         )
     }
 
+    func prepareForTermination() async {
+        guard case .recording = state else { return }
+        guard !isFinishingRecording else { return }
+        isFinishingRecording = true
+        defer { isFinishingRecording = false }
+
+        _ = audioInactivityDetector.stopRecording()
+        audioInactivityWarning = nil
+
+        let recordingTrigger = activeRecordingTrigger
+        let files = await capture.stopAndAwaitFiles()
+        activeRecordingTrigger = .unknown
+
+        if let micURL = files.micURL {
+            failedManager.addFailedTranscription(
+                micAudioURL: micURL,
+                systemAudioURL: files.systemURL,
+                errorMessage: "Transcripted quit before this meeting could be transcribed."
+            )
+            refreshFailedMeetings()
+        }
+
+        state = .ready
+        DiagnosticsTrail.record(
+            level: .warning,
+            engine: "meeting",
+            event: "meeting_recording_saved_for_shutdown",
+            message: "Meeting recording was preserved during app termination",
+            context: baseDiagnosticsContext(
+                extra: [
+                    "trigger": recordingTrigger.rawValue,
+                    "mic_file_present": boolString(files.micURL != nil),
+                    "system_file_present": boolString(files.systemURL != nil)
+                ]
+            )
+        )
+    }
+
     func retryFailedMeeting(id: UUID) {
         guard !isRecording, !hasBackgroundTranscriptionWork else { return }
         guard !retryingFailedMeetingIDs.contains(id) else { return }
@@ -784,7 +823,7 @@ final class MeetingSessionController: ObservableObject {
             guard let self else { return }
             _ = await self.taskManager.retryFailedTranscription(
                 failedId: id,
-                outputFolder: self.storagePaths.transcripts
+                outputFolder: MeetingStoragePaths.transcriptsFolder
             )
             self.retryingFailedMeetingIDs.remove(id)
             self.refreshFailedMeetings()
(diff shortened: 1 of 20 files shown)


r3dbars:
## Code Review — PR #509: Harden release QA reliability paths

**Summary**: This PR bundles four reliability improvements for release readiness: graceful shutdown that preserves active dictation and meeting recordings, hotkey refresh after onboarding permission grants, hardened path security for CLI/MCP transcript file handling, and smoother loading states in the menu bar and settings UI. The changes are well-structured, properly tested, and address real edge cases that would cause data loss or UX jank during release usage.

---

### Blockers
None.

---

### Should fix

1. **`finishDictationForTermination()` calls `stopDictationAndPaste(trigger: .unknown)`** (`DictationSessionController.swift:661`). During app termination, the "paste" half of stop-and-paste will attempt to write to another app's text field via accessibility. If the target app is also shutting down, or the pasteboard round-trip races the termination reply, this could hang or silently fail. Consider a dedicated `stopDictationAndSave(trigger:)` path that persists the transcript without the paste step, or at minimum document why paste-during-termination is acceptable.

2. **10-second polling ceiling in `finishDictationForTermination`** (`DictationSessionController.swift:663-666`) combined with the `await capture.stopAndAwaitFiles()` in `prepareForTermination()` means the worst-case `applicationShouldTerminate` delay could exceed 10 seconds. macOS doesn't enforce a strict timeout for `.terminateLater` during normal quit, but during system shutdown or restart the window is shorter. Consider reducing the poll cap or adding a combined deadline across both cleanup phases.

3. **`retainedAudioDirectoryProvider` alongside `retainedAudioDirectory`** (`TranscriptionTaskManager.swift:27,47`) — both are passed with the same value at the call site. The static property is now shadowed by the provider in every code path. Consider deprecating the static URL parameter in a follow-up to avoid drift where a caller sets one but not the other.

---

### Nits

1. **Duplicate `CLIPathSecurity` vs MCP `PathSecurity`** — the CLI package adds its own `CLIPathSecurity` that is functionally identical to the MCP's `PathSecurity` (same enum shape, same `validateExistingFile` logic). They're in separate packages so some duplication is expected, but the minor differences (MCP has `appendingExtension`, CLI does the `fileExists` check inside `validateExistingFile` only) could diverge over time. Worth extracting to a shared target if the packages ever get a shared dependency.

2. **`paste_available: "unknown"` in diagnostics** (`TranscriptedApp.swift:449`, `MenuBarUtilityActionsView.swift:125`) — changing from a real boolean to `"unknown"` means analytics consumers lose signal on paste availability from these surfaces. If the goal is avoiding the synchronous disk hit, the async-loaded value from `MenuBarPanelController.latestDictation` could be threaded through, or the field could be dropped entirely rather than emitting a sentinel that might confuse dashboards.

3. **`parseBracketedSpeakerLabel` uses `Array(value)` materialization** (`MeetingTranscriptStyler.swift:350`) — for typical short speaker labels this is fine, but `String.Index`-based iteration would avoid the allocation. Very minor given transcript lines are short.

4. **`SpeakerPeopleSettingsViewModel.refreshGeneration`** (`SpeakerPeopleSettingsSection.swift:88`) — the generation counter correctly prevents stale snapshots, but it's an `Int` that will silently wrap on overflow after 2^63 refreshes. Practically impossible, but `UInt64` or a cancellation-based approach (cancel the previous `DispatchQueue.global().async` work item) would be more idiomatic.

---

### What looks good

- **Graceful shutdown flow** — using `applicationShouldTerminate` → `.terminateLater` → async cleanup → `reply(toApplicationShouldTerminate:)` is the correct macOS lifecycle pattern. The `terminationCleanupStarted` guard prevents double-entry. Good.

- **`prepareForTermination()` saves to failed queue** — preserving in-progress meeting recordings as failed transcriptions that the user can retry later is exactly the right design. No data loss, clear diagnostics trail, and it reuses the existing retry infrastructure.

- **`parseBracketedSpeakerLabel` wiki-link depth tracking** — the depth-counted bracket parser correctly handles `[System/[[Alex]]]` without breaking on the inner `]]`. The test coverage with `sampleObsidianTranscript()` validates the real-world use case.

- **Path security hardening** — both the CLI and MCP now reject parent traversal (`../`) and symlink escapes with proper test coverage (`testReadDictationRejectsParentTraversal`, `testReadDictationRejectsSymlinkEscape`). The defense-in-depth approach (standardized path prefix check + resolved symlink check) is solid.

- **FileWatcher batch change detection** — switching from per-file `onChange(URL)` to batch `onChange()` with full reconcile eliminates a class of race conditions where rapid file changes could leave the index inconsistent. Detecting deletions via key-set comparison is a nice touch.

- **`indexSingleFile` priority ordering** — iterating roots in order and taking the first valid match ensures the "current" directory wins over "legacy" for duplicate basenames. The `testIndexSingleFileKeepsHigherPriorityDuplicateBasename` test validates this clearly.

- **Async lazy loading for `latestDictation`** — moving the synchronous `DictationTranscriptStore.latestSavedDictation()` off the main thread with the `latestDictationLoaded` guard + `Task.detached` pattern prevents menu bar jank. The `dictationTranscriptDidSave` notification invalidation keeps it fresh.

- **Settings loading states** — showing "Loading recent meetings..." / "Loading recent dictations..." instead of "No meetings saved yet" during async loads prevents false-negative UX.

- **Test coverage** — every behavioral change has corresponding test coverage: Obsidian links, path traversal, symlink escape, file watcher deletes, index priority, QA validators. Well done.

---

PR #511, opened 2026-04-24
Fix dictation readiness refresh after audio recovery

r3dbars:
## Summary
- refresh dictation input readiness when audio recovery finishes unready
- add a small policy guard/test for the recovery-wait decision
- keep refreshes bounded so the wait loop does not hammer CoreAudio

## Why
Sentry APPLE-MACOS-14 showed dictation timing out after 6s with start_attempts=0 and readiness_refreshes=0 while recovery had left the input unready. That can look to users like dictation starts and then turns itself off without crashing.

## Verification
- bash build-deps.sh --force
- bash build.sh
- bash run-tests.sh


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR fixes a real gap in the dictation readiness wait loop where recovery finishing with `inputFormatReady == false` left the loop doing nothing useful until the 6-second deadline expired — matching the Sentry crash report exactly. The fix extracts the three-state decision into a named policy, adds a throttled refresh path for the previously-unhandled state, and includes clean unit tests.

**Blockers**: None.

**Should fix**:
- The test file covers three of the four input combinations but skips `(isRecovering: true, inputFormatReady: true)`. The policy correctly returns `.waitForRecovery` for that case (recovery should finish cleanly even if format looks momentarily ready), and a fourth test case would document that the "recovery takes priority" invariant is intentional. Low risk but worth adding for completeness.

**Nits**:
- `DictationReadinessWaitPolicy.swift` lines 1–3: the file-header comment ("Small policy for deciding how dictation should react…") is fine but slightly redundant with the type name and enum cases already being self-documenting. Take or leave.

**What looks good**:
- The policy extraction follows the established codebase pattern (`DictationInputDeviceSelectionPolicy`, `ParakeetPrewarmPolicy`, etc.) — pure function, no side effects, testable in isolation.
- Throttling via `nextReadinessRefreshAt` at 300ms vs the 150ms poll interval is a good balance: responsive enough to unstick CoreAudio quickly, bounded enough to avoid hammering it.
- The `readinessRefreshes` counter was already wired into the timeout diagnostics, so the new refresh calls automatically show up in Sentry reports — good observability for free.
- Test manifest, test runner `APP_SOURCES`, and `FastTests.manifest` are all correctly updated. No loose ends.
- The scope is tight: fixes the bug, adds the guard, nothing extra.

---

PR #513, opened 2026-04-24
Fix updater stuck in checking state

r3dbars:
## Summary
- avoid putting the updater UI into Checking unless Sparkle can actually start or observe a check
- reset to an idle state when Sparkle is not ready so the menu does not stay disabled forever

## Verification
- bash build-deps.sh --force
- bash build.sh
- bash run-tests.sh

### DIFF
--- a/Sources/Observability/SparkleUpdaterController.swift
+++ b/Sources/Observability/SparkleUpdaterController.swift
@@ -82,12 +82,10 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
         }
 
         if updaterController.updater.automaticallyChecksForUpdates {
-            beginObservedUpdateCheck()
-
             // Sparkle recommends forcing launch-time background checks, if
             // desired, immediately after the updater has started and only when
             // automatic checks are enabled.
-            guard !updaterController.updater.sessionInProgress else { return }
+            guard beginObservedUpdateCheckIfPossible(allowExistingSession: true) else { return }
             updaterController.updater.checkForUpdatesInBackground()
         } else {
             refreshUpdateStatus()
@@ -100,8 +98,7 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
             return
         }
 
-        beginObservedUpdateCheck()
-        guard updaterController.updater.canCheckForUpdates else { return }
+        guard beginObservedUpdateCheckIfPossible(allowExistingSession: true) else { return }
         updaterController.checkForUpdates(nil)
     }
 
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review: Fix updater stuck in checking state

**Verdict**: APPROVE

**Summary**: This PR fixes a bug where the updater UI gets permanently stuck in "Checking" state when Sparkle can't actually perform an update check. The root cause was that `beginObservedUpdateCheck()` (which sets UI to `.checking`) was called unconditionally *before* verifying Sparkle preconditions (`canCheckForUpdates`, `!sessionInProgress`). When the guards then bailed out, the UI was already stuck. The fix introduces `beginObservedUpdateCheckIfPossible(allowExistingSession:)` which checks preconditions first, only transitions to `.checking` when a check will actually happen, and resets to idle otherwise. Clean, well-scoped fix.

**Blockers**: None.

**Should fix**: None.

**Nits**:

- `markUpdaterIdle` reads `updater.canCheckForUpdates` twice on lines 287-288. Could cache it locally for consistency with how other methods in the file handle it (e.g. `syncReadiness` caches into `let canCheckForUpdates`). Not a real performance concern since it's a KVO property on `@MainActor`, just a style alignment with the rest of the file:
  ```swift
  let canCheck = updater.canCheckForUpdates
  let state: UpdateStatus.State = canCheck ? .readyToCheck : .unknown
  setUpdateStatus(state, canCheckForUpdates: canCheck)
  ```
  (Note: the same double-read pattern pre-exists in `markUpdateCheckFailed` — not introduced by this PR.)

**What looks good**:

- The core insight is exactly right: never set UI to "checking" unless you're confident a check will actually proceed. This eliminates an entire class of stuck-state bugs.
- `beginObservedUpdateCheckIfPossible` is a well-designed consolidation — it replaces four separate inline guard-then-begin patterns with a single gating function, making the state machine easier to reason about.
- The `allowExistingSession` parameter is a good design choice: startup and user-initiated checks should observe an in-progress session, while refresh/auto-download should not piggyback. The boolean name is clear at call sites.
- `markUpdaterIdle` correctly preserves update-available state by deferring to `syncReadiness` when a version is already known, avoiding accidentally hiding a pending update.
- The `didFinishUpdateCycleFor` delegate callback already handles cleanup for the `.checking` state, so this fix focuses on prevention (don't enter `.checking` incorrectly) rather than adding more recovery paths. Good approach.
- Net -12/+35 for a single file — tight scope, no collateral changes.

---

PR #515, opened 2026-04-25
Disable Sentry app hang noise

r3dbars:
## Summary
- Disable Sentry automatic app-hang tracking by default because modal macOS alerts can be misreported as hangs.
- Keep crash reporting, NSException reporting, Sentry scrubbing, and allowlisted manual/non-fatal events enabled.
- Add explicit env/plist opt-in for local app-hang tracking validation.

Fixes APPLE-MACOS-4 for future releases.

## Validation
- bash run-tests.sh — 846 passed
- bash build-deps.sh --force
- bash build.sh

### DIFF
--- a/Sources/Observability/CrashReporter.swift
+++ b/Sources/Observability/CrashReporter.swift
@@ -31,6 +31,7 @@ final class CrashReporter {
         options.sendDefaultPii = false
         options.enableCrashHandler = true
         options.enableUncaughtNSExceptionReporting = true
+        options.enableAppHangTracking = SentryRuntimeConfiguration.appHangTrackingEnabled()
         options.enableAutoSessionTracking = false
         options.enableNetworkBreadcrumbs = false
         options.maxBreadcrumbs = 0
--- a/Sources/Observability/SentryRuntimeConfiguration.swift
+++ b/Sources/Observability/SentryRuntimeConfiguration.swift
@@ -3,8 +3,10 @@ import Foundation
 enum SentryRuntimeConfiguration {
     static let dsnInfoKey = "TranscriptedSentryDSN"
     static let environmentInfoKey = "TranscriptedSentryEnvironment"
+    static let appHangTrackingInfoKey = "TranscriptedSentryAppHangTrackingEnabled"
     static let dsnEnvironmentKey = "SENTRY_DSN"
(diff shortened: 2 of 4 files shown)


r3dbars:
## Code Review — PR #515: Disable Sentry app hang noise

**Verdict**: APPROVE

**Summary**: This PR disables Sentry's automatic app-hang tracking by default because macOS modal dialogs (update, permission, confirmation) get misreported as hangs. It adds a clean opt-in mechanism via environment variable or Info.plist, defaults to off, and fails closed on invalid input. Well-scoped change with good test coverage and updated docs.

None.

- **Minor code reuse opportunity**: `parseBoolean` in `SentryRuntimeConfiguration` is nearly identical to `isEnabledFlag` in `Sources/TranscriptedCore/Logging/FileLogger.swift` (lines 119–125), which parses the same `"1"/"true"/"yes"/"on"` set. The new version is slightly richer (returns `Bool?` and recognizes explicit false values), but this is the second boolean-from-string parser in the codebase. Worth consolidating into a shared utility in a follow-up to avoid drift — not a blocker for this PR.

- **Test gap — plist string path**: Tests cover plist `Bool` (`true`) and env var strings, but the `configuredString` code path (plist value delivered as a `String` like `"true"`) has no dedicated test case. The code is straightforward and implicitly covered by `parseBoolean` tests, but an explicit test would prevent future regressions if that branch is refactored.

**Nits** (take or leave):

- The nested `if let` in `appHangTrackingEnabled()` (lines 48–55) could flatten slightly with a `switch` over `configured`'s type or an early-return guard, but current readability is fine and matches the rest of the file's style.

**What looks good**:

- **Fail-closed design**: Unrecognized env var values default to `false` and override a plist `true` — exactly the right call for a noisy diagnostic feature. The "not-a-bool" test case validates this explicitly.
- **Testability**: Dependency injection of `environment` and `infoDictionary` parameters follows the established pattern from `dsn()` and `environment()`, making every code path unit-testable without mocking.
- **Docs updated in lockstep**: Privacy contract, setup checklist, and the opt-in instructions in `privacy-first-observability.md` all updated alongside the code change.
- **Minimal blast radius**: One line in `CrashReporter.setup()`, one new config method, tests, and docs. No unnecessary refactoring or feature creep.

---

PR #521, opened 2026-04-25
Code review cleanups: TranscriptedCore boundary, onboarding poll, EOU lookup, stale TODO

r3dbars:
## Summary

Four low-risk fixes flagged in a code review of `main`:

- **TranscriptedCore boundary** — `Audio.swift` and `AudioFileManager.swift` were dragging `import AppKit` into the reusable library to play two `NSSound` cues on recording start/stop. Replaced with a `CaptureLifecycleCue` callback on `Audio`; the host (`MeetingCaptureBridge`) now plays the sounds. `AudioFileManager.swift` no longer imports AppKit. `Audio.swift` still imports AppKit because of `NSWorkspace` sleep/wake notifications — that's a real OS API and a separate concern from cosmetic UI cues.
- **Onboarding polling** — `PermissionsOnboardingView` was using `Timer.scheduledTimer` plus a nested `Task { @MainActor in checkAllPermissions() }`. Foundation timers run in `.default` mode and pause during menu tracking, so the loop could miss a permission flip while System Settings was open. Replaced with a single cancellable `Task` that awaits `Task.sleep` — same `onDisappear` cleanup, but polling continues regardless of UI run-loop state.
- **EOU model cache lookup** — `ParakeetEngine.initializeEouModel()` had `urls(for:.applicationSupportDirectory, in:.userDomainMask).first!`. Replaced with `guard let` + a structured `EventReporter` warning + clean return. Almost-impossible to hit in practice, but a hard force-unwrap on a deep init path was a hard crash if it ever did fail.
- **Stale Phase-2 TODO** — `MeetingSessionController` carried a comment claiming meetings shipped without failed-transcription retry UI. That stopped being true some time ago: the queue is drained by `refreshFailedMeetings()` (subscribed to `failedManager.$failedTranscriptions`), surfaced in Settings → Meetings → "Needs Attention" via `SettingsFailedMeetingRow`, and `retryFailedMeeting` / `dismissFailedMeeting` / `deleteFailedMeeting` wire the actions through. Comment-only change; rewritten to describe current reality.

No user-visible change for #1, #2, #4. #3 converts a rare crash into a logged warning.

## Test plan

- [x] `bash build.sh` — clean
- [x] `bash run-tests.sh` — 846 / 846 passed
- [x] `bash run-integration-smoke.sh` — passed (Audio + AudioFileManager touched)
- [x] `swift test` — 77 / 77 passed (TranscriptedCore package seam)
- [ ] Manual: meeting recording start/stop still plays Tink/Pop sounds
- [ ] Manual: permissions onboarding still flips green when permissions are granted while System Settings is in the foreground

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -14,6 +14,7 @@
 //   3. Keeping the bridge isolated from the pipeline lets Lane C swap in a mock
 //      for preview/testing without touching CoreAudio.
 
+import AppKit
 import AVFoundation
 import Combine
 import Foundation
@@ -228,6 +229,23 @@ final class MeetingCaptureBridge: ObservableObject {
                 self.completionTimeoutTask?.cancel()
                 self.completionTimeoutTask = nil
                 continuation?.resume(returning: (micURL, systemURL))
+            }
+        }
+
+        // Capture-lifecycle cues used to live inside Core (NSSound("Tink") on
+        // start, NSSound("Pop") on stop). Core no longer depends on AppKit for
+        // cosmetic UI; the host plays the sounds here. Audio fires the cue from
+        // the main queue (via DispatchQueue.main.async / MainActor.run inside
+        // the lifecycle helpers), but we still bounce through Task @MainActor
+        // to match the rest of the bridge's threading discipline.
+        audio.onCaptureLifecycleCue = { cue in
+            Task { @MainActor in
+                switch cue {
+                case .recordingStarted:
+                    NSSound(named: "Tink")?.play()
+                case .recordingStopped:
+                    NSSound(named: "Pop")?.play()
+                }
             }
         }
     }
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -208,11 +208,12 @@ final class MeetingSessionController: ObservableObject {
         self.statsDatabase = StatsDatabase(path: storagePaths.statsDB.path)
 
         // Failed-queue manager: takes CoreStoragePaths so its JSON file lives
-        // under app-owned state, not the capture library.
-        // TODO: Phase 2 ships without failed-transcription recovery for meeting
-        // mode — we construct the manager because TranscriptionTaskManager
-        // requires it, but nothing in the app currently drains the failed queue
-        // or exposes retry UI. Follow-up work lands in a later phase.
+        // under app-owned state, not the capture library. The queue is drained
+        // by `refreshFailedMeetings()` (subscribed to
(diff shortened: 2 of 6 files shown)


r3dbars (commented):

**Summary**: Three low-risk, well-scoped fixes: moving `NSSound` playback out of `TranscriptedCore` into the host layer via a callback, replacing a Foundation `Timer`-based permission poll with a `Task`-driven loop that survives run-loop mode changes, and converting a force-unwrap on Application Support directory lookup into a guarded early return with structured logging. All three changes are correct, cleanly implemented, and well-motivated.

**Blockers**: None.

**Should fix**: None.

**Nits**:

- `CaptureLifecycleCue` lives in `Audio.swift` alongside its producer, which is fine for a two-case enum. If more consumers or cases appear later, moving it to `TranscriptedCore/Models/` would improve discoverability — but not worth doing now.

- The PR description says "Audio.swift still imports AppKit because of `NSWorkspace` sleep/wake notifications" — confirmed at lines 363/373 (`NSWorkspace.willSleepNotification` / `didWakeNotification`). Worth noting that extracting those into `NotificationCenter`-based wrappers someday would let `Audio.swift` fully drop the AppKit import, but that's a separate concern and out of scope here.

**What looks good**:

- **Thread-safe callback capture**: Both `Audio.stop()` and `AudioFileManager.startAudioCapture()` correctly snapshot `onCaptureLifecycleCue` into a local `let cueHandler` before crossing async boundaries (`DispatchQueue.main.async` and `await MainActor.run`). This avoids a data race if the property is nilled out between dispatch and execution. Textbook pattern for `@unchecked Sendable` classes with mutable callback properties.

- **No retain cycle in the bridge closure**: `audio.onCaptureLifecycleCue` in `MeetingCaptureBridge.wireCallbacks()` doesn't capture `self` — only the `cue` parameter and `NSSound` — so there's no bridge → audio → closure → bridge cycle. The adjacent `onRecordingComplete` correctly uses `[weak self]` because it *does* reference `self`. Good consistency.

- **Polling loop correctness**: The `Task`-based loop in `PermissionsOnboardingView` checks `Task.isCancelled` both in the loop condition and after the sleep, so cancellation is responsive. The initial `checkAllPermissions()` call happens synchronously in `onAppear` before `startPolling()`, so there's zero initial delay — actually better than the old Timer which had a 1-second wait before the first fire.

- **EOU guard-let**: Converting `first!` to `guard let` + `EventReporter` + early return is the right call. Nearly impossible to hit, but a force-unwrap on a deep init path is a hard crash if it ever does fail. The structured event log makes it diagnosable without a crash report.

- **PR description quality**: Clear motivation for each change, explicit test plan with both automated and manual items, and honest about what's user-visible vs. not. Well done.

---

PR #522, opened 2026-04-25
Fix dictation clipboard and meeting menu state

r3dbars:
## Summary
- restore the previous clipboard quickly after dictation paste-back so immediate manual paste does not reuse dictated text
- make the meeting menu row switch to Stop Meeting while recording and toggle stop from the menu
- add a small HTML preview for the meeting menu row states

## Verification
- bash build-deps.sh --force
- bash build.sh
- bash run-tests.sh

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -40,6 +40,7 @@ final class MeetingSessionController: ObservableObject {
     enum StopReason: String {
         case hotkeyToggle = "hotkey_toggle"
         case overlayStopButton = "overlay_stop_button"
+        case menuBarStopButton = "menu_bar_stop_button"
         case audioInactivityPrompt = "audio_inactivity_prompt"
         case audioInactivityTimeout = "audio_inactivity_timeout"
         case unknown = "unknown"
--- a/Sources/Support/ClipboardRestoringTextPaster.swift
+++ b/Sources/Support/ClipboardRestoringTextPaster.swift
@@ -96,15 +96,7 @@ final class ClipboardRestoringTextPaster {
 
         let changeCountAfterSet = pasteboard.changeCount
         clipboardRestoreTask = Task { @MainActor [weak self] in
-            let startTime = CFAbsoluteTimeGetCurrent()
-            while CFAbsoluteTimeGetCurrent() - startTime < TranscriptedConstants.clipboardRestoreTimeout {
-                try? await Task.sleep(nanoseconds: TranscriptedConstants.clipboardPollInterval)
-                guard !Task.isCancelled else { return }
-                if pasteboard.changeCount != changeCountAfterSet {
-                    return
-                }
-            }
-
+            try? await Task.sleep(nanoseconds: TranscriptedConstants.clipboardRestoreDelay)
             guard !Task.isCancelled else { return }
             self?.restorePasteboardItems(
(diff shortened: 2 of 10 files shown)


r3dbars:
## Code Review — PR #522

**Verdict**: APPROVE

**Summary**: This PR fixes two distinct UX issues: (1) the clipboard is restored much faster after dictation paste-back so users don't accidentally re-paste dictated text, and (2) the meeting menu row now toggles to "Stop Meeting" while a recording is active instead of confusingly staying on "Start Meeting." Both changes are clean, well-scoped, and properly tested.

---

### Blockers
None.

---

### Should fix

**1. Meeting row tone stays `.accent` during recording** — `MenuBarPrimaryActionsView.swift:78` always passes `tone: .accent` regardless of whether the row says "Start Meeting" or "Stop Meeting." The HTML preview shows a red stop icon, which suggests the intent is for the row to feel visually distinct during recording. Consider adding a `tone` property to `MenuBarPrimaryActionState` (or a conditional at the call site) so the stop state uses `.warning` or a dedicated stop tone. This would reinforce the destructive-action feel of stopping a recording.

**2. MenuBarShortcutsView meeting row doesn't reflect recording state** — The shortcuts view's meeting row (`MenuBarShortcutsView.swift:192-201`) hardcodes `symbolName: "record.circle.fill"` and `title: "Record meeting"`, and tapping it triggers the same `onStartMeeting` callback. During an active recording, it would still say "Record meeting" but actually stop the meeting. Low priority since users are unlikely to be editing shortcuts mid-recording, but worth noting for consistency.

---

### Nits

**1. Clipboard restore timing** — The switch from a 2-second polling loop to a fixed 120ms delay is a meaningful behavioral change. 120ms is plenty for most apps to process a CGEvent-based Cmd+V, and the `restorePasteboardItems` guard (checking `changeCount` and content) prevents incorrect restoration if another clipboard write intervenes. The test asserting `clipboardRestoreDelay < dictationAutoEnterDelay` is a nice invariant. Just noting this is tight for very slow or heavily-loaded target apps — if paste failures are reported, this constant is the first place to look.

**2. Explicit memberwise init on `MenuBarPrimaryActionState`** — The explicit `init` at `FirstRunExperience.swift:38-48` matches the synthesized memberwise init exactly. It's harmless, but the struct would get the same init for free without it.

**3. HTML preview** — `docs/meeting-menu-row-preview.html` is a nice documentation artifact for design review. No concerns.

---

### What looks good

- **Reactive menu state**: Subscribing to `appState.meetingSession.$state` in `MenuBarPanelController` and refreshing the whole panel is the right approach — it's consistent with how every other publisher is handled in `setupSubscriptions()`.
- **`StopReason.menuBarStopButton`**: Clean enum extension. Analytics and diagnostics will correctly attribute stops from the menu bar.
- **Task capture pattern**: `Task { [meetingSession = appState.meetingSession] in ... }` in `startMeetingFromMenu()` avoids a retain cycle on `self` through `appState`. Good habit.
- **Guard safety**: Both `startRecording()` and `stopRecording()` on `MeetingSessionController` have guards that make the toggle in `startMeetingFromMenu()` safe against race conditions (double-start or double-stop are no-ops).
- **Test coverage**: Both the `meetingAction` recording-state test and the `clipboardRestoreDelay < dictationAutoEnterDelay` invariant test are targeted and meaningful. They verify behavior, not implementation.

---

PR #540, opened 2026-04-26

r3dbars:
## Summary
Nightly automated security audit for Transcripted. This PR applies minimal security-only fixes for confirmed issues and leaves non-exploitable findings unchanged.

### Medium
- Hardened `FailedTranscriptionManager` so failed-queue entries are only persisted when their audio paths stay inside Transcripted-owned audio roots.
- Re-validated queued audio paths immediately before deletion to prevent cleanup flows from acting on tampered or out-of-sandbox filesystem targets.

### Low
- Re-applied owner-only permissions after `FileLogger` trims/atomically rewrites `app.jsonl`, preventing sensitive logs from drifting back to default broader permissions after rotation-like rewrites.

### Reviewed, no change needed
- `SpeakerDatabase` and `StatsDatabase` queries are parameterized/serialized; no exploitable SQL injection found.
- Transcript save path handling already validates custom destinations and falls back to safe defaults.
- Sparkle update packaging/signing remained on secure channels in the audited paths.
- HuggingFace/model download handling already uses HTTPS and existing integrity-related plumbing in the reviewed code paths.
- Repository scans did not surface live hardcoded secrets.
- Remaining `try!`, force unwrap, and unsafe pointer sites reviewed in production sources did not present a confirmed security exploit path during this audit.

## Verification
- `bash build.sh`
- `bash run-tests.sh`


r3dbars:
## Code Review: PR #540 — security: nightly security audit findings [automated]

**Verdict**: APPROVE

**Summary**: This PR adds three targeted security hardening measures: (1) path-sandboxing guards in `FailedTranscriptionManager` at the enqueue and pre-deletion entry points, closing gaps where only the deserialization path was previously validated, and (2) re-applying owner-only file permissions in `FileLogger` after atomic rewrites and file recreation, which reset permissions to umask defaults via new-inode creation. The changes are minimal, correct, and well-placed.

**Blockers**: None.

**Should fix**:

- **Missing test coverage for new code paths.** The existing `testInitRejectsOutOfHomeAudioPathsAndRewritesQueue` validates the `loadFailedTranscriptions` deserialization guard, but no tests exercise the two new guards introduced by this PR:
  - `addFailedTranscription()` rejecting an out-of-sandbox `micAudioURL` or `systemAudioURL`
  - `removeAudioFile()` refusing to delete an out-of-sandbox path
  
  While the underlying `isSafeAudioURL` logic is the same, explicit tests for these entry points would confirm the guards are wired correctly and prevent regressions if the method signatures change.

- **No test verifying file permissions after `FileLogger` trim.** The existing `testLoggerTrimsOversizedExistingLogOnInit` exercises the trim path but doesn't assert that the resulting file has 0o600 permissions. A one-line `XCTAssertEqual(fileAttributes[.posixPermissions], 0o600)` after trim would lock down the fix.

**Nits**:

- The security comments (3-line blocks in FileLogger, 2-line blocks in FailedTranscriptionManager) are justified — they explain *why* atomic writes reset permissions and *why* defense-in-depth re-validation is needed. These are exactly the kind of non-obvious constraints that warrant comments. No change needed.

- The `removeAudioFile` label strings across callers are inconsistent (`"mic audio"`, `"cleanup mic audio"`, `"old failure mic audio"`). Not introduced by this PR, but worth standardizing in a follow-up for easier log filtering.

**What looks good**:

- **Defense in depth done right.** Validation at three layers — enqueue, deserialization, and pre-deletion — each serving a distinct purpose in the threat model. The enqueue guard prevents bad data from entering; the deserialization guard catches tampered JSON on disk; the pre-deletion guard is the last line of defense before the destructive `removeItem` call.
- **Correct understanding of atomic write semantics.** `write(to:atomically:)` creates a temp file and renames it, so the new inode inherits umask-based permissions (often 0644). The `restrictToOwnerOnly` calls are placed precisely where needed — not redundant.
- **Symlink-aware path canonicalization.** `isSafeAudioURL` uses `standardizedFileURL.resolvingSymlinksInPath()` to prevent path traversal and symlink escape attacks.
- **Minimal diff, no scope creep.** 26 lines added across 2 files, all security-focused. No unrelated refactoring or feature work mixed in.

---

PR #541, opened 2026-04-26

r3dbars:
## Summary
- keep meeting transcription running when the microphone capture is shorter than Parakeet's minimum but the system track is still long enough to transcribe
- tighten short-audio logging so both mic and system durations are visible when a capture is rejected

### Transcription pipeline
- `TranscriptionTaskManager.startTranscription` rejected the entire meeting whenever the mic file was under 2 seconds, even if the system audio capture was valid.
- This could drop otherwise recoverable meeting transcripts after mic recovery edge cases, because the short mic stub caused both recordings to be deleted before transcription started.
- The gate now only rejects captures when every available track is too short, and it explicitly logs when we continue with usable system audio.

## Verification
- `bash build.sh`
- `bash run-tests.sh`

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
@@ -83,10 +83,21 @@ public class TranscriptionTaskManager: ObservableObject {
         }
 
 
-        // Gate: reject recordings shorter than 2 seconds (they'll fail in Parakeet anyway)
+        // Gate: reject only when every available capture track is too short.
+        // Meeting recovery can produce a very short mic stub while system audio is still
+        // intact and fully transcribable, so don't throw away the whole recording just
+        // because the mic side is below Parakeet's minimum length.
         let minDuration: TimeInterval = 2.0
-        if let micDuration = audioDuration(url: micURL), micDuration < minDuration {
-            AppLogger.pipeline.info("Recording too short, skipping transcription", ["duration": String(format: "%.1fs", micDuration)])
+        let micDuration = audioDuration(url: micURL)
+        let systemDuration = systemURL.flatMap { audioDuration(url: $0) }
+        let hasUsableMicAudio = (micDuration ?? 0) >= minDuration
+        let hasUsableSystemAudio = (systemDuration ?? 0) >= minDuration
+
+        guard hasUsableMicAudio || hasUsableSystemAudio else {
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review — PR #541

**Verdict**: APPROVE

**Summary**: This PR fixes a real bug where `startTranscription` would reject an entire meeting recording when only the mic capture was under 2 seconds, even if the system audio track was long enough to transcribe. The fix is correct, minimal, and well-scoped — it changes the gate to reject only when *every* available track is too short, and adds a warning log when proceeding with short mic audio.

### Blockers
None.

### Should fix
Nothing specific to this PR's changes. Two pre-existing items surfaced during review that are worth addressing separately:
- **Magic number duplication**: `let minDuration: TimeInterval = 2.0` appears independently in both `startTranscription` (line 90) and `startImportedTranscription` (line 203). If one is updated without the other, they'll drift silently. Consider extracting to a shared constant (e.g. in `TranscriptedConstants`). Not caused by this PR.
- **Repeated format pattern**: `String(format: "%.1fs", $0)` appears 7+ times across the file. A small helper would reduce the copy-paste surface. Also pre-existing.

### Nits
- The nil-fallback string for `systemDuration` differs between the two log sites: `"none"` in the guard block (line 99) vs `"unknown"` in the warning (line 115). In practice the `"unknown"` fallback is **unreachable** — `hasUsableSystemAudio` being true guarantees `systemDuration` is non-nil — so this is purely cosmetic. Using `"none"` consistently would read slightly better if someone grepping logs doesn't know the invariant, but it's take-or-leave.

### What looks good
- **Correct downstream behavior confirmed**: The pipeline handles short mic files gracefully through `prepareMicSegmentForTranscription` (drops silent segments), `AudioSignalRecovery.padForParakeet` (pads segments with detected speech), and `ParakeetShortAudioGate` (returns empty string instead of throwing). A sub-2-second mic file passed through will never crash the pipeline.
- **Improved nil-duration handling**: The old code silently bypassed the duration check when `audioDuration(url:)` returned nil (the `if let` binding just failed). The new code treats nil as 0, which correctly routes unreadable mic files through the same logic.
- **Clean control flow**: The `guard` for rejection + separate `if` for the warning log keeps the two concerns distinct without nesting.
- **Well-named intermediates**: `hasUsableMicAudio` / `hasUsableSystemAudio` make the boolean logic self-documenting.
- **Justified comment**: The 4-line comment explaining the recovery scenario is appropriate — domain knowledge a future reader wouldn't infer from the code alone.

---

PR #543, opened 2026-04-26

r3dbars:

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -83,24 +83,13 @@ final class MeetingCaptureBridge: ObservableObject {
     /// Start a new recording session. Returns immediately; the session remains
     /// active until `stopAndAwaitFiles()` is called.
     func startRecording() async -> Bool {
-        if let continuation = completionContinuation {
-            completionContinuation = nil
-            completionAttemptID = nil
-            completionTimeoutTask?.cancel()
-            completionTimeoutTask = nil
-            continuation.resume(returning: CaptureStopResult(
-                micURL: audio.micAudioFileURL,
-                systemURL: audio.systemAudioFileURL,
-                didTimeOut: false
-            ))
-        }
+        resetCompletionAttempt()?.resume(returning: currentStopResult())
         if audio.isRecording { return true }
 
         errorMessage = nil
 
         return await withCheckedContinuation { continuation in
-            startTimeoutTask?.cancel()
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: Extracts four private helpers from repeated inline patterns in `MeetingCaptureBridge` — `resetStartAttempt()`, `resetCompletionAttempt()`, `currentStopResult(didTimeOut:)`, and `sinkStartAttemptTriggers(from:)`. Pure refactoring, no behavioral change. The continuation resume-exactly-once invariant is preserved at every call site, and the net result is 9 fewer lines that are significantly easier to audit.

**Blockers**: None.

**Should fix**: None.

**Nits**:

- In the timeout handlers (lines 108, 141), the pattern of guard-let capturing the continuation *then* calling `resetXxxAttempt()` (which also captures and returns it) *then* discarding the return with `_ =` reads a little odd on first pass — a future reader might wonder who resumes the continuation. This is correct because the guard-let binding happens before `resetXxxAttempt()` nils the property, but it's worth being aware of if these handlers are ever touched again. Not worth changing now since the `startAttemptID == attemptID` guard *must* run before the reset nils the ID.

- `deinit` (lines 69–78) still uses the inline `CaptureStopResult(...)` constructor rather than `currentStopResult()`. Totally fine — calling instance methods in `deinit` is valid for `final class`, and it's only one site, but noting for consistency awareness.

**What looks good**:

- `resetStartAttempt()` and `resetCompletionAttempt()` returning the captured continuation as an `Optional` is a nice pattern — it lets callers chain `?.resume(returning:)` for the simple cases (line 86, 92) and use `_ =` + local variable for the guarded timeout cases. Clean API for an internal helper.
- `currentStopResult(didTimeOut:)` with a `false` default parameter is a good fit — eliminates the most repeated boilerplate without hiding the timeout distinction.
- `sinkStartAttemptTriggers(from:)` correctly preserves `[weak self]` and `.receive(on: RunLoop.main)`, and the generic `Published<Output>.Publisher` signature is appropriately scoped for the two call sites.
- The `self?.finishPendingStartAttemptIfPossible()` simplification from the longer `guard let self` pattern is a nice touch — equivalent behavior, less noise.

Solid cleanup. Ship it.

---
*Note: Could not submit as a formal approval since the PR and review come from the same account. Build/test verification requires macOS toolchain not available in this environment — recommend confirming `bash build.sh && bash run-tests.sh && bash run-integration-smoke.sh` passes before merge.*

---

PR #546, opened 2026-04-27

r3dbars:
## Summary
- fixed a low-severity log-permissions issue in `FileLogger`
- added a regression test covering the atomic rewrite path
- audited the requested security areas and found no additional vulnerabilities requiring code changes

### Critical
- none

### High
- none

### Medium
- none

### Low
- `Sources/TranscriptedCore/Logging/FileLogger.swift`: when oversized logs were trimmed, the atomic rewrite could replace `app.jsonl` with a new inode using default umask permissions. On multi-user systems that could make sensitive application logs broader than the intended owner-only `0600` mode. The fix reapplies owner-only permissions after the atomic rewrite and also tightens custom log directories to `0700`.

## Verification
- `bash build.sh`
- `bash run-tests.sh`


r3dbars:
## PR Review: security: nightly security audit findings [automated]

**Verdict**: APPROVE (with one suggested fix)

**Summary**: This PR fixes a low-severity permissions issue where `FileLogger.trimIfNeeded()` performs an atomic rewrite of `app.jsonl`, replacing the inode with one that inherits the process umask (typically `0o644`) instead of the intended owner-only `0o600`. The fix reapplies `restrictToOwnerOnly` after the atomic write and tightens the logs directory to `0o700`. A well-structured regression test covers the trim path. Clean, minimal, correct.

### Blockers
- None.

### Should fix
- **Fallback `createFile` on line 143 missing permission hardening.** In `trimIfNeeded()`, when the atomic write somehow deletes the file and the fallback `createFile` recreates it, `restrictToOwnerOnly` is not called on the new file. The `init` path (lines 48–51) correctly pairs `createFile` with `restrictToOwnerOnly`, but this fallback doesn't. The parent directory is `0o700` so practical risk is low, but for consistency and defense-in-depth, add:
  ```swift
  FileManager.default.createFile(atPath: logFileURL.path, contents: nil)
  FileManager.default.restrictToOwnerOnly(atPath: logFileURL.path)  // <- add this
  ```

### Nits
- **Inline `0o700` could be a helper.** The raw `setAttributes([.posixPermissions: 0o700], ...)` pattern appears in 4 places across the codebase (`FileLogger`, `RecordingAudioArchiver`, `SpeakerClipExtractor`, `TranscriptedStoragePaths`). A `restrictDirectoryToOwnerOnly(atPath:)` method in `FilePermissions.swift` alongside the existing `restrictToOwnerOnly` would make the intent self-documenting and eliminate duplicated octal literals. Not required for this PR, but worth a follow-up.
- **No test for the directory `0o700` change.** The new test covers the file permission after trim, but the directory tightening in `init` is untested. Minor — the existing `0o700` pattern works fine elsewhere.

### What looks good
- **Correct diagnosis.** Atomic writes replacing the inode is a real and subtle permissions pitfall. Good catch.
- **Uses established patterns.** The fix calls `restrictToOwnerOnly` — the same helper used in 14+ sites across TranscriptedCore — rather than introducing a one-off inline `setAttributes` call.
- **TOCTOU is mitigated.** The `flock(LOCK_EX)` held across the entire trim cycle plus the `0o700` parent directory means the brief window between atomic write and chmod is not exploitable in practice.
- **Test is well-crafted.** Sets up deliberately loose permissions (`0o644`), triggers the exact trim path (2105 entries > `maxEntries`), and verifies the fix. Follows the existing test patterns with proper `defer` cleanup.

---

PR #547, opened 2026-04-27

r3dbars:
## Summary
- fix retained-audio archiving to use the live capture-library archive path for successful transcriptions, matching the failed-transcription path
- add a regression test covering provider-backed retained-audio directory resolution

### Logic bug
- successful recording archive copies previously used the task manager's initial `retainedAudioDirectory` snapshot, so changing the capture library at runtime could keep saving retained audio into the old archive location
- failed-transcription archiving already used the live `retainedAudioDirectoryProvider`, so the behavior was inconsistent across success vs failure paths

### Tests
- added coverage proving the task manager prefers the live provider value when it changes during runtime

## Verification
- `bash build.sh`
- `bash run-tests.sh`


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR fixes a real logic inconsistency where successful-transcription audio archiving used a stale `retainedAudioDirectory` snapshot from init time, while failed-transcription archiving already used the live `retainedAudioDirectoryProvider` closure. The fix extracts a single `currentRetainedAudioDirectory()` method that both paths now share, and adds a focused regression test. Clean, well-scoped change.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- `retainedAudioDirectory` (line 26 of `TranscriptionTaskManager.swift`) is `let` with `internal` access. Now that `currentRetainedAudioDirectory()` exists as the correct way to get the live value, a future caller within the module could still accidentally read the stale stored property directly. Consider marking it `private` in a follow-up if nothing outside the class reads it. Not a regression from this PR — just a guardrail opportunity.

**What looks good**:
- The bug analysis is spot-on: `retainedAudioDirectory` is a `let` set at init, so it never reflects runtime capture-library changes. The provider closure is the live path. Unifying them is the right call.
- Thread safety is correct: the pipeline runner's `nonisolated` method properly hops to `@MainActor` via `await MainActor.run { ... }` before calling the method on the `@MainActor`-isolated class.
- The test is well-structured — it verifies both the initial value and the updated value after mutating the captured variable, which is exactly the scenario that triggered the bug.
- `currentRetainedAudioDirectory()` is `internal` (not `public`), keeping it out of the library's public API surface while still accessible from the extension file and `@testable` tests. Right access level.
- The `makeManager` helper update with defaulted optional parameters is clean and doesn't break the existing test.

---

PR #549, opened 2026-04-27

r3dbars:

### DIFF
--- a/Sources/Meeting/MeetingFailureCopy.swift
+++ b/Sources/Meeting/MeetingFailureCopy.swift
@@ -11,16 +11,6 @@ struct MeetingFailureCopy: Equatable {
     ) -> MeetingFailureCopy {
         let message = errorMessage.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
 
-        // Keep the broader "microphone access" wording that some meeting-start
-        // errors use even though the analytics classifier keys off more specific
-        // permission phrases.
-        if message.contains("microphone access") {
-            return MeetingFailureCopy(
-                title: "Turn on Microphone",
-                detail: "Turn on Microphone access in System Settings, then retry the meeting."
-            )
-        }
-
         switch MeetingFailureKind.classify(message: message) {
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: APPROVE (posted as comment — can't formally approve own PR)

**Summary**: This PR consolidates the broad `"microphone access"` error matching from a special-case early-return in `MeetingFailureCopy` into the canonical `MeetingFailureKind.classify(...)` classifier, and refactors repeated `||`-chained `contains` calls into a small `contains(anyOf:)` helper. Clean, well-motivated change that aligns with the documented invariant that `MeetingFailureKind` is the single source of truth for failure classification.

**Blockers**: None.

**Should fix**: None.

**Nits** (take or leave):

- `"microphone access denied"` in the `.microphonePermission` fragment list is now dead — `"microphone access"` is a substring of it, so the broader pattern always matches first. Removing the redundant entry would make the list honest about what's actually doing the matching. Not a correctness issue, just readability.

```swift
if normalized.contains(anyOf: [
    "microphone access",
    // "microphone access denied" is already covered by the line above
    "microphone permission",
    "mic_not_authorized",
]) {
```

**What looks good**:

- Moving the classification into `MeetingFailureKind` benefits all four call sites (`MeetingFailureCopy`, `FailedMeetingPresentation`, `MeetingSessionController` analytics, `MeetingOverlayController`), not just the one that had the workaround. Good consolidation.
- The `contains(anyOf:)` helper is a nice minimal abstraction — point-free `fragments.contains(where: contains)` is idiomatic Swift, scoped `private` so it doesn't leak, and the compound `isRecordingTooShortMessage` was correctly left alone since it doesn't fit the pattern.
- Existing `FailedMeetingPresentationTests` at line 45 already exercises the exact "Turn on Microphone access..." message through `MeetingFailureCopy.make(...)`, so behavioral equivalence is covered by both the new and existing test suites.
- Test added for the new classifier behavior. Commit message is clear.

---

Issue #500, opened 2026-04-24
Mic audio recording is a lot quiter


LeonStaufer:
**Describe the bug**
If I'm in a meeting and the meeting software uses the device microphone, the recording in Transcripted using my internal microphone is a lot quieter and thus it is not picked up in the transcription.

The meeting can be empty and no one else speaking. Just being in a meeting means that my microphone audio is picked up a lot quieter.

**Steps to reproduce**
1. Enter a meeting that uses microphone, e.g. Google Meet or even WhatsApp
2. Start Transcripted meeting recording
3. The mic audio is a lot quiter
4. Thus my speaking is not included in transcription

**Expected behavior**
My microphone should still be equal loudness and included in transcription.

**Actual behavior**
My microphone is only picked up very quietly and not included in the transcript.

**Area**
- Meetings

**Environment**
- macOS version: 26.3.1 (a) (25D771280a)
- Transcripted version/commit: 1.1.18
- Audio source (mic, Zoom, Meet, etc.): mic and any meeting software. Both when using internal speakers and bluetooth devices to listen to the meeting.

**Diagnostics**
If you have logs, screenshots, or exported diagnostics, attach them here.

Normal loudness when not in meeting:
<img width="301" height="48" alt="Image" src="https://github.com/user-attachments/assets/04f58c9a-6a8e-481b-a334-9529be21b03d" />

Very quite at the same volume of speaking in an empty Google Meet:
<img width="299" height="65" alt="Image" src="https://github.com/user-attachments/assets/4387a4a4-6b06-4dca-972e-89aea56f82ad" />

justinbetker:
Thanks for flagging this, I will look into this and get this fixed!

r3dbars:
Got a fix for this in! I am going to push a new release today sometime. Let me know if this helps

LeonStaufer:
Thanks! Unfortunately, it still doesnt work in 1.1.22. Once the microphone is in use, the recorded volume by Transcripted is very quiet.

r3dbars:
Is this through the browser? And what web browser are you using? I know there are a lot of issues with browsers and the web rtc and mic sharing on Mac OS. I am going to try to recreate this issue and see if we can make this work for ya!

r3dbars:
Yeah I just try to re-create this issue and chromium based browsers work fine like, chrome, atlas and firefox and safari have the web rtc issue, where it lowers the mic volume tremendously. I am not sure the fix here, but I can look into it a bit more.

[...]

PR #553, opened 2026-04-27
Default VPIO off, swap in real-time software AGC for meeting mic

r3dbars:
## Summary

- Replace unconditional `setVoiceProcessingEnabled(true)` (Apple VPIO) with an opt-in Settings toggle, defaulting **off**.
- Add a real-time software AGC (`RealtimeAGC`, vDSP-based) in the mic tap callback as the new default path. It boosts attenuated mic streams in software without engaging macOS voice-comms ducking.
- Settings → Meetings now exposes a "Microphone Processing" toggle for the rare user who specifically needs VPIO (Safari/Firefox WebRTC meetings).

## Why

PR #523 / v1.1.24 enabled VPIO unconditionally to fix issue #500 (Safari/Firefox WebRTC attenuates the shared mic device, making meeting recordings come out quiet). Real-world side effect, reported on Slack: macOS treats any VPIO holder as a voice-comms app and ducks output from other apps — Zoom playback gets noticeably quieter while Transcripted records, and quitting Transcripted is the only way to restore Zoom volume.

Apple's [WWDC23 session](https://developer.apple.com/videos/play/wwdc2023/10235/) and the [`voiceProcessingOtherAudioDuckingConfiguration`](https://developer.apple.com/documentation/avfaudio/avaudioinputnode/voiceprocessingotheraudioduckingconfiguration) docs confirm `.min` ducking is "the smallest amount of ducking" — not zero. PR #535's tuning helps but cannot eliminate the regression on its own.

Comparable Mac apps (MacWhisper, SuperWhisper, Granola, Wispr Flow) all avoid VPIO continuously for record-only paths and do gain cleanup elsewhere. This PR brings Transcripted in line with that pattern.

### New files
- `Sources/TranscriptedCore/Audio/RealtimeAGC.swift` — peak-tracking AGC, rolling 16-buffer window, vDSP `vDSP_maxmgv` / `vDSP_vsmul` / `vDSP_vclip`, fast-attack / slow-release smoothing, hard-clip to `[-1, 1]`. Real-time safe: zero allocations and no locks in `process(buffer:)`.
- `Sources/Support/MicrophoneProcessingPreferences.swift` — UserDefaults bool, default off, injectable for tests.
- `Tests/TranscriptedCoreTests/RealtimeAGCTests.swift` — 9 tests covering ramp-up to target, gain cap, normal pass-through, silence-hold, reset, empty buffer, hard-clip safety, interleaved stereo.
- `Tests/MicrophoneProcessingPreferencesTests.swift` — 4 tests covering default, persistence, round-trip, stable storage key.

### Modified files
- `Sources/TranscriptedCore/Audio/Audio.swift` — adds public `enableVoiceProcessing` flag (default false), gates `armVoiceProcessing` on it, drops `realtimeAGC` reference at `stop()` and `stopMonitoring()`.
- `Sources/TranscriptedCore/Audio/AudioFileManager.swift` — instantiates `RealtimeAGC` at start when VPIO is off, applies it to the deep-copied buffer before file write and consumer callback. Level meter intentionally still reads the raw (pre-AGC) buffer so silence/inactivity detection keeps firing in a quiet room.
- `Sources/TranscriptedCore/Audio/AudioDeviceRecovery.swift` — re-evaluates AGC on device change (drop if VPIO armed, reset if existing, create new if missing).
- `Sources/Meeting/MeetingCaptureBridge.swift` — reads the preference and applies it before each `audio.start()`.
- `Sources/UI/Settings/TranscriptedSettingsView.swift` — adds "Microphone Processing" section to the Meetings page with a toggle that explicitly names Zoom and Safari/Firefox in the explanation.
- `Tests/TranscriptedCoreTests/AudioInitializationTests.swift` — adds a default-off assertion on `enableVoiceProcessing`.
- `Tests/FastTests.manifest`, `scripts/entrypoints/run-tests.sh` — register the new preference test and source.

### What stayed the same
- `AudioSignalRecovery.normalizeForSpeech` and the post-processing pipeline are untouched. The new real-time AGC produces files at roughly the same loudness target, so the post-processing path is now usually a no-op (gain stays near 1.0) but remains as a safety net.
- VPIO with `.min` ducking config (PR #535's tuning) is preserved on the opt-in path; users who turn the toggle on get exactly the v1.1.24+#534+#535 behavior.
- Issue #500 reporters running Safari/Firefox can still get the original fix by toggling the new setting on.

## Test plan

- [x] `bash build.sh` (build green, app signed and verified)
- [x] `bash run-tests.sh` — 899/899 pass
- [x] `bash run-integration-smoke.sh` — 2/2 pass
- [x] `swift test` — 92/92 pass (including 9 new RealtimeAGC tests)
- [ ] Manual repro: confirm Zoom-ducking regression is gone with the toggle off — record a meeting while Zoom is playing audio and verify Zoom volume stays normal.
- [ ] Manual repro: enable the toggle, verify VPIO arms and Safari/Firefox WebRTC attenuation case is recovered (compare saved `meeting_*_mic.wav` loudness to off-toggle baseline in same condition).
- [ ] Manual repro: with toggle off, record a meeting in a quiet room and verify the "Still recording?" inactivity prompt still fires.
- [ ] Bluetooth/AirPods sanity: with toggle off, verify recording starts and stays clean.
- [ ] Device-switch sanity: with toggle off, switch input mid-recording and verify AGC resets cleanly across the new tap.

## Out of scope (intentionally)

- The Force-Quit dialog visibility ask from Taylor and the meeting-widget freeze report — those are tracked as separate PRs.


r3dbars (commented):

**Summary**: This PR cleanly replaces unconditional VPIO (which causes macOS to duck other apps' audio) with a software AGC default path, while preserving VPIO as an opt-in for the Safari/Firefox WebRTC case. The `RealtimeAGC` implementation is sound, the lifecycle integration across start/stop/monitoring/device-recovery is thorough, and the 13 new tests cover the important invariants.

---

**Blockers**: None.

---

**Should fix**:

- `Notification.Name.microphoneProcessingPrefsDidChange` is posted in `setVoiceProcessingEnabled` but no observer is registered anywhere in this PR or (from what I can see) the existing codebase. If it's a forward-looking hook, that's fine, but if the intent was to trigger a runtime behavior change (e.g. updating the bridge mid-session), the observer is missing. Worth a one-line comment clarifying intent, or removing the post if unneeded. No functional impact either way — the preference is read at recording start.

---

**Nits**:

- The comments in this PR are more verbose than the codebase norm (CLAUDE.md says "default to writing no comments"), but for audio DSP code where the "why" involves IIR coefficients, attack/release semantics, and CoreAudio threading invariants, the verbosity is justified. I'd just note that a few of the inline comments in `handleMicBuffer` restate what the doc comment already covers — could trim those if you want, but no strong feelings.

- `RealtimeAGC.init` clamps `bufferTimeMs` to `max(0.001, bufferTimeMs)` but the parameter doc says "≈ 85ms". The 0.001 floor means a caller passing 0 gets `α ≈ 1.0` (snap to target instantly) rather than a crash. Defensive, correct, but the asymmetry with `attackTimeMs`/`releaseTimeMs` (which use `max(0.1, ...)`) is worth being aware of — a truly degenerate `bufferTimeMs` wouldn't crash but would make the smoothing behave oddly.

---

**What looks good**:

- **Lifecycle completeness**: AGC is created at start, nil'd at stop and stopMonitoring, reset on device change, and conditionally dropped if VPIO arms during recovery. Every path is covered — no stale gain leaking across sessions.

- **Level meter on raw signal**: Keeping `calculateLevel(buffer:)` on the pre-AGC buffer so silence detection still fires in a quiet room is the right call. AGC would amplify ambient noise to speech-looking levels and defeat the inactivity prompt. This is a subtle correctness detail that's easy to get wrong.

- **Thread safety via tap ordering**: `realtimeAGC` is written only between tap-remove + engine-stop and tap-reinstall, which is the same pattern the codebase uses for `voiceProcessingEnabled` and other audio state. No new concurrency hazard introduced.

- **Consumer callback now gets a deep copy**: Before this PR, `onMicPCMBuffer` received the CoreAudio-owned buffer that could be overwritten in the next callback. Now it gets a stable deep copy (with AGC applied). This is a correctness improvement independent of the AGC feature.

- **Test quality**: The AGC tests cover ramp-up convergence, gain cap, normal pass-through, no-below-unity invariant, silence hold (no pumping), reset, empty buffer, hard-clip safety, and interleaved stereo. The silence-hold test correctly drains the full rolling window before asserting gain stability. The preferences tests lock the storage key to prevent silent invalidation on refactor.

- **Settings UX**: The toggle copy clearly names the tradeoff ("quiets audio from other apps like Zoom") and the "takes effect on next recording" note prevents confusion about mid-session behavior.

- **Security**: No concerns. No user-controlled input beyond a boolean toggle, no file path manipulation beyond existing patterns, no network access, static UserDefaults key.

---

PR #554, opened 2026-04-27
Show recording sessions in the macOS force-quit dialog

r3dbars:
## Summary

- Add `ActivationPolicyController` that switches `NSApp` between `.accessory` (idle) and `.regular` (recording) so a frozen recording session shows up in Cmd+Option+Esc.
- Subscribe to `STTRouter.$isRecording` (dictation) and `MeetingSessionController.$isRecording` (meeting); the OR of the two drives policy.
- Tests: 9 unit cases covering the pure decision matrix and the controller's coalescing/init behavior.

## Why

Transcripted runs as `LSUIElement = true` with `setActivationPolicy(.accessory)`, so it's deliberately hidden from the macOS force-quit dialog. Taylor reported on Slack that the meeting widget freezes occasionally during a session, and his only recovery path was opening Activity Monitor — that's a bad UX for a stuck recording. macOS only shows `.regular` apps in Cmd+Opt+Esc, so the cleanest way to make Transcripted force-quittable is to flip activation policy while a session is running and revert when it stops.

## What stayed the same

- Idle state is unchanged: `.accessory` policy, `LSUIElement = true` in Info.plist, no Dock icon, no force-quit visibility.
- The menubar `NSStatusItem` is independent of activation policy and stays visible the whole time — users won't see anything different in the menubar.

## What changes during a recording

- Dock icon flickers in for the duration of the session. (Trade-off accepted: this is the only way to get force-quit visibility, and it's only visible while the session is live.)
- App shows up in Cmd+Opt+Esc; users can force-quit it without opening Activity Monitor.
- Reverts to `.accessory` (Dock icon disappears, force-quit hidden again) when the session stops.

## Files

- New `Sources/Support/ActivationPolicyController.swift` — pure decision enum + `@MainActor` controller with injectable apply closure for tests.
- `Sources/TranscriptedApp.swift` — replaces the static `setActivationPolicy(.accessory)` call with the controller; subscribes to both recording flags via Combine.
- New `Tests/ActivationPolicyControllerTests.swift` — 9 cases:
  - decision returns `.accessory` when both off
  - decision returns `.regular` for meeting alone, dictation alone, both
  - init applies initial policy exactly once
  - meeting on flips policy
  - meeting stop with no other session restores `.accessory`
  - overlapping session keeps `.regular` while either is active (no Dock bouncing)
  - redundant flag updates don't re-apply policy

## Test plan

- [x] `bash build.sh` (signed bundle)
- [x] `bash run-tests.sh` — 905/905 (was 899; 6 new tests)
- [x] `bash run-integration-smoke.sh` — 2/2
- [x] `swift test` — 82/82
- [ ] Manual: idle app should stay menubar-only (no Dock icon).
- [ ] Manual: start a meeting → confirm Dock icon appears AND app shows up in Cmd+Opt+Esc.
- [ ] Manual: stop meeting → confirm Dock icon disappears AND app drops out of force-quit dialog.
- [ ] Manual: same flow but with dictation instead of meeting.
- [ ] Manual: start meeting + dictation overlap (if reachable) → verify policy stays `.regular` until both stop.

## Out of scope

- The Zoom-ducking fix (#553) and the meeting widget freeze fix (separate PR) are not part of this change. This PR only addresses the recovery path; it does not stop the freeze from happening in the first place. The widget-freeze PR will land alongside.

### DIFF
--- a/Sources/Support/ActivationPolicyController.swift
+++ b/Sources/Support/ActivationPolicyController.swift
@@ -1,0 +1,97 @@
+// Support/ActivationPolicyController.swift
+// Dynamically switches NSApp's activation policy so an active recording
+// session shows up in the macOS force-quit dialog (Cmd+Option+Esc).
+//
+// Background: Transcripted is a menubar app, so it normally runs with
+// activation policy `.accessory` and `LSUIElement = true`. macOS hides
+// `.accessory` apps from the force-quit dialog; if the app freezes during
+// a recording, the user has to open Activity Monitor to kill it. Reported
+// pain: Taylor's meeting widget froze and the only recovery was Activity
+// Monitor.
+//
+// Fix shape: while a recording is active (meeting OR dictation), switch
+// to `.regular` so the app appears in the force-quit dialog. When all
+// sessions stop, switch back to `.accessory`. The Dock icon flickers in
+// for the duration of the recording — acceptable trade for a recovery
+// path. The menubar `NSStatusItem` is independent of activation policy
+// and stays visible the whole time.
+//
+// Thread-safety: this type is `@MainActor` because `NSApplication`
+// state is main-thread-only. Callers update the recording flags from
+// the main thread; subscribers running on other threads should hop
+// before calling `update(...)`.
+
+import AppKit
+import Foundation
+
+/// Pure policy decision so we can unit-test it without an `NSApp`
+/// instance. Given the current recording flags, returns the activation
+/// policy the app should have.
+enum ActivationPolicyDecision {
+
+    /// Returns `.regular` if any session is active so the app shows in
+    /// the force-quit dialog. Otherwise returns `.accessory` to keep
+    /// the Dock clean.
+    static func desiredPolicy(
+        isMeetingRecording: Bool,
+        isDictationRecording: Bool
+    ) -> NSApplication.ActivationPolicy {
+        if isMeetingRecording || isDictationRecording {
+            return .regular
+        }
+        return .accessory
+    }
+}
+
+@MainActor
+final class ActivationPolicyController {
+
+    /// Most-recently applied policy. Used so we don't redundantly call
+    /// `setActivationPolicy(_:)` (which has side effects on the Dock
+    /// tile and key-window state).
+    private(set) var currentPolicy: NSApplication.ActivationPolicy
+
+    private var isMeetingRecording: Bool = false
+    private var isDictationRecording: Bool = false
+
+    /// Indirection so tests can capture policy changes without touching
+    /// NSApp.
+    private let applyPolicy: (NSApplication.ActivationPolicy) -> Void
+
+    init(
+        initialPolicy: NSApplication.ActivationPolicy = .accessory,
+        applyPolicy: @escaping (NSApplication.ActivationPolicy) -> Void = { policy in
+            NSApp.setActivationPolicy(policy)
+        }
+    ) {
+        self.currentPolicy = initialPolicy
+        self.applyPolicy = applyPolicy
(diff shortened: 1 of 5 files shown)


r3dbars:
## Code Review — PR #554

**Verdict**: APPROVE

**Summary**: Clean, well-scoped fix for a real user pain point. The `ActivationPolicyController` correctly toggles `NSApp` between `.accessory` and `.regular` based on the OR of meeting/dictation recording flags, giving users a force-quit path during frozen sessions. Architecture is sound: pure decision logic separated from side effects, closure injection for testability, proper coalescing to avoid Dock bouncing, and thorough test coverage of the full state matrix. No blockers.

---

### Blockers
None.

---

### Should fix

**1. File-header comment is PR-description material, not source code (lines 1–22)**
The 22-line header recounts the bug report origin ("Taylor's meeting widget froze"), trade-off rationale, and thread-safety notes. This is excellent context — for the PR description, which already has it. In source code it will rot. Suggest trimming to 2–3 lines (what it does and the key invariant). The `@MainActor` annotation already communicates thread-safety; the comment on line 22 references a nonexistent `update(...)` method.

**2. `ActivationPolicyDecision` enum is module-scoped but only used by the controller**
This exists "so we can unit-test it without an `NSApp` instance," but the controller already accepts an injectable `applyPolicy` closure for exactly that purpose — and the tests use that injection, not the enum directly (except the 4 decision-matrix cases). The decision logic is one OR expression. Suggest either making it a `private` nested type / static func on the controller, or inlining it in `reconcile()`. The 4 decision tests can exercise the same logic through the controller's `PolicyRecorder` pattern. As-is, it adds an internal type to the module namespace that no other code will reference.

---

### Nits

**3. `DispatchQueue.main` vs codebase convention `RunLoop.main`**
The existing Combine subscriptions in this codebase (`FloatingOverlayController`, `MenuBarPanelController`, `bindStatusItemUpdateBadge`) consistently use `.receive(on: RunLoop.main)`. The new subscriptions use `.receive(on: DispatchQueue.main)`. Both work, but matching the local convention keeps things consistent.

**4. `.removeDuplicates()` before the sinks**
`@Published` emits on every `willSet`, including redundant `isRecording = false` when already `false`. The controller's internal guards handle this correctly, but adding `.removeDuplicates()` in the Combine pipeline filters at the publisher layer before even entering the closure — strictly cheaper and more idiomatic.

**5. `activationPolicyController` is `Optional` but never nil after launch**
It's assigned unconditionally in `applicationDidFinishLaunching`. If the app delegate's init timing permits, a non-optional `lazy var` would remove the optionality. Minor — the current shape is fine and consistent with other late-init properties in the delegate.

**6. Test count discrepancy in PR description**
PR body says "905/905 (was 899; 6 new tests)" but the test file has 9 `runSuite` calls. Might be a manifest-vs-harness counting difference — worth a quick check that all 9 cases actually run.

---

### What looks good

- **Closure injection for testability** — the `applyPolicy` parameter is the right pattern for testing AppKit side effects. Clean and idiomatic.
- **Double-guard coalescing** — `setMeetingRecording`/`setDictationRecording` guard on flag change, then `reconcile()` guards on policy change. This correctly prevents redundant `setActivationPolicy` calls that would bounce the Dock tile.
- **Overlapping session test** — the test that verifies `[.accessory, .regular, .accessory]` across a meeting+dictation overlap is exactly the edge case that matters. Good that it's explicitly covered.
- **`#available(macOS 14.0, *)` gate** — correctly matches the availability of `MeetingSessionController` so pre-Sonoma builds compile cleanly.
- **Scope discipline** — this PR only adds the recovery path; it doesn't try to fix the freeze itself. That separation is the right call.

---

PR #555, opened 2026-04-27
Move audio stop teardown off main thread + reorder health snapshot

r3dbars:
## Summary

- `Audio.stop()` now dispatches the synchronous AVFoundation teardown (engine.stop, removeTap, disarmVoiceProcessing) to a background queue so the main run loop stays responsive while the audio thread drains its current buffer.
- `RecordingHealthInfo.from(...)` and `Audio.createHealthInfo(...)` gain an `overrideSystemAudioStatus` parameter so callers can snapshot the live status BEFORE stop (where stop resets it to `.unknown`) and read the lock-protected health fields AFTER stop (where the audio thread is no longer mutating them).
- `MeetingSessionController.stopRecording()` follows the new order.

## Why

Taylor reported on Slack that the meeting widget freezes "sometimes when I try to engage with it" during meetings — and his only recovery was Activity Monitor. (#554 is the recovery-path fix; this PR is the actual freeze fix.)

A focused investigation traced two compounding main-thread blocks in the stop path:

1. **`engine.stop()` blocks for as long as the audio thread takes to finish its current buffer.** With voice processing active, `setVoiceProcessingEnabled(false)` adds another 5–30ms. Combined and serialized on main, this can stall the meeting widget for hundreds of ms.
2. **`capture.healthInfo()` ran BEFORE `stopAndAwaitFiles()`** at `MeetingSessionController.swift:498`. The factory reads `recordingGaps`, `deviceSwitchCount`, and `recoveryAttemptCount` — all `NSLock`-protected and concurrently written by the audio/recovery threads while buffers drain. Lock contention adds a few more ms per read.

### `Sources/TranscriptedCore/Audio/Audio.swift::stop()`
Rewritten so the only synchronous main-thread work is:
- `recordingSessionGeneration &+= 1` (so concurrent recovery sees the new boundary immediately)
- snapshotting refs (`engineRef`, `inputNodeRef`, `systemCaptureRef`, URLs, segments, callback)
- `DispatchQueue.main.async` to flip `isRecording`/UI state

Then `DispatchQueue.global(qos: .userInitiated).async` runs:
- `inputNode.removeTap` + `engine.stop` + `disarmVoiceProcessing`
- `systemAudioCapture?.stop()` (already non-blocking, kept ordered)
- The original `DispatchGroup` mic+system file-close coordination
- `cleanupGroup.notify` → `finalizeMicRecording` → `DispatchQueue.main.async { onRecordingComplete }`

The `MeetingCaptureBridge.stopAndAwaitFiles()` continuation still resumes when `onRecordingComplete` fires, so all callers' `await` semantics are unchanged — but the main run loop is no longer blocked while the audio engine shuts down.

### `Sources/TranscriptedCore/Models/RecordingHealthInfo.swift`
Added `overrideSystemAudioStatus: SystemAudioStatus? = nil` to `from(...)`. When supplied, it replaces `audio.systemAudioStatus` for the `.failed` check that drives `successRate`. Default-nil so existing callers (`TranscriptMetadataBuilderTests`, the `AudioCaptureEngine` protocol no-arg `createHealthInfo()`) keep working unchanged.

### `Sources/TranscriptedCore/Audio/Audio.swift::createHealthInfo()`
Two variants now:
- `createHealthInfo()` — preserves the protocol-required no-arg shape used by the `AudioCaptureEngine` conformance.
- `createHealthInfo(overrideSystemAudioStatus:)` — the new path, used by callers that snapshot health AFTER stop.

### `Sources/Meeting/MeetingCaptureBridge.swift::healthInfo(...)`
Forwards the optional override to `audio.createHealthInfo(overrideSystemAudioStatus:)`.

### `Sources/Meeting/MeetingSessionController.swift::stopRecording(...)`
Reordered:
```swift
let finalSystemAudioStatus = capture.systemAudioStatus     // pre-stop snapshot
let stopResult = await capture.stopAndAwaitFiles()
let healthInfo = capture.healthInfo(overrideSystemAudioStatus: finalSystemAudioStatus)
```

## What stays the same

- The await contract on `stopAndAwaitFiles()` — callers still receive control after the full file-close pipeline.
- `Audio.deinit`'s synchronous engine teardown — unchanged (deinit is meant to be sync).
- `RecordingHealthInfo.from(...)` callers that don't pass the override — unchanged.

## Tests

- `AudioInitializationTests.testStopOnIdleAudioDoesNotCrashAndBumpsGeneration` (new) — confirms `stop()` synchronously bumps the recording-session generation and returns immediately when there's no real engine.
- `RecordingHealthInfoOverrideTests` (new, 5 cases):
  - live status used when no override
  - override `.failed` overrides live `.unknown` post-stop
  - healthy override doesn't degrade an otherwise-clean recording
  - persistent `systemAudioFailed` flag still dominates over a healthy override
  - `Audio.createHealthInfo(overrideSystemAudioStatus:)` forwards to the factory

## Test plan

- [x] `bash build.sh` — green, signed bundle
- [x] `bash run-tests.sh` — 895/895
- [x] `bash run-integration-smoke.sh` — 2/2
- [x] `swift test` — 88/88 (6 new tests)
- [ ] Manual: start a meeting, click Stop, confirm the widget animates promptly to "Saving…" without a multi-second freeze.
- [ ] Manual: stop a meeting that had a `systemAudioStatus = .failed` event during capture; verify the saved transcript's `capture_quality` reflects the failure (not a false-clean reading).
- [ ] Manual: stress test by stopping back-to-back with rapid start clicks; the engine teardown is now async, so confirm no Bad State if a new recording starts before the previous teardown's `onRecordingComplete` fires.

## Out of scope

- The Zoom-ducking fix (#553) and the force-quit visibility fix (#554) are independent PRs in this trio. This PR is the third and final.
- This is a surgical fix targeting the two specific blocking sites identified by sub-agent investigation. There may be additional hidden stalls (e.g., backed-up file write queues, slow disk I/O) that this PR doesn't address. If freezes persist after merging, those become the next investigation target.


r3dbars (commented):
**Verdict**: APPROVE (with one recommended fix)

**Summary**: This PR moves `AVAudioEngine` teardown off the main thread during `Audio.stop()` to eliminate the multi-hundred-ms widget freeze Taylor reported, and reorders the health snapshot in `MeetingSessionController.stopRecording()` so `systemAudioStatus` is captured before stop resets it while lock-protected fields are read after the audio thread drains. The approach is well-reasoned, the `overrideSystemAudioStatus` parameter design is clean, and test coverage is thorough. One thread-safety concern worth addressing before merge.

---

**Rapid start/stop race on `AVAudioEngine`** — `stop()` snapshots `engineRef`/`inputNodeRef` and dispatches teardown to `DispatchQueue.global`, but never nils out `self.engine` / `self.inputNode`. If the user taps Start before the background `engineRef.stop()` finishes (the window is small but real — the PR description itself flags this as a manual test concern), `ensureEngineInitialized()` at `Audio.swift:438` sees `engine != nil` and returns the *same instance* that's concurrently being stopped on the background queue. `AVAudioEngine` is not documented as thread-safe for concurrent `stop()` / `start()` / `installTap()` calls.

Suggested fix: nil out `self.engine` and `self.inputNode` synchronously in `stop()` right after snapshotting, so the next `start()` path through `ensureEngineInitialized()` creates a fresh engine:

```swift
let engineRef = self.engine
let inputNodeRef = self.inputNode
+ self.engine = nil
+ self.inputNode = nil
let systemCaptureRef = self.systemAudioCapture
```

This also makes `deinit`'s guard (`if let engine, let inputNode, engine.isRunning`) a harmless no-op after a normal stop, which is already the desired behavior. Recovery code on background threads already checks `recordingSessionGeneration` before touching the engine, so they'll correctly bail out.

---

**Nits** (take or leave):

1. **Comment density**: 55 of 124 added source lines are comments (~44%). The comments explain real subtlety (the "why" of each dispatch boundary, the ordering contract), so they earn their keep. But a few could be trimmed — e.g. the 6-line block at `Audio.swift:679–683` explaining the `DispatchQueue.main.async` could be shortened since the method-level context already establishes the "freeze fix" framing.

2. **Two `createHealthInfo` overloads vs. one with default nil**: The no-arg `createHealthInfo()` and `createHealthInfo(overrideSystemAudioStatus:)` could be a single method with `overrideSystemAudioStatus: SystemAudioStatus? = nil`. The PR description says the split preserves the `AudioCaptureEngine` protocol conformance — if the protocol requires the exact no-arg signature, the split is correct and this is a non-issue.

---

**What looks good**:

- **Root cause analysis is solid.** Two distinct main-thread stalls identified (engine drain + lock contention on health reads) and each addressed independently with clear before/after ordering.
- **Snapshot-then-override pattern for `RecordingHealthInfo`** is the right design. Default-nil preserves all existing callers, the `systemAudioFailed` flag correctly dominates over the transient override, and the test matrix at `RecordingHealthInfoOverrideTests` covers the important edges (absent override, override wins, healthy override, persistent flag dominance, forwarding through `Audio.createHealthInfo`).
- **`recordingSessionGeneration` bump stays synchronous** — critical for invalidating concurrent recovery work immediately.
- **`[weak self]` throughout nested closures** is correct and prevents retain cycles during the multi-hop async teardown.
- **`deinit` is unaffected** — its `engine.isRunning` guard correctly short-circuits if background teardown already stopped the engine.
- **Well-scoped PR** — fixes the freeze, doesn't over-reach into adjacent issues, explicitly calls out what's out of scope.

---

PR #558, opened 2026-04-28
Give Transcripted a permanent Dock icon

r3dbars:
## Summary
- Flip `LSUIElement` to `false` so Transcripted shows a permanent Dock icon instead of being a menubar-only `.accessory` app that only briefly appeared in the Dock during recordings.
- Route Dock-icon clicks (via `applicationShouldHandleReopen`) to the Settings window on the Home page when no window is already visible; otherwise let AppKit bring the existing window to the front.
- Simplify `ActivationPolicyController` to always report `.regular`. Its original purpose — making the app appear in the force-quit dialog during a recording — is now satisfied for free by the always-regular policy. The recording-flag wiring is kept as a no-op to avoid churning callers, and the unit tests are updated to assert the new always-`.regular` behavior.

The menubar status item and its existing quick-action popover are unchanged — left-click still opens the popover, exactly as before.

## Test plan
- [x] `bash build.sh` passes.
- [x] `bash run-tests.sh` — 909/909 pass.
- [ ] Manually verify: Transcripted appears in the Dock at launch (no longer hides as `.accessory`).
- [ ] Click the Dock icon with no window open → Settings window opens on Home.
- [ ] Click the Dock icon with the Settings window already visible → existing window comes to the front.
- [ ] Left-click the menubar icon → quick-action popover appears (unchanged behavior).
- [ ] First-run: clicking the Dock icon before onboarding completion routes to the onboarding window.
- [ ] Start and stop a meeting / dictation → Dock icon stays put (no flicker between accessory and regular).

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -29,7 +29,7 @@
 	<key>LSMinimumSystemVersion</key>
 	<string>26.0</string>
 	<key>LSUIElement</key>
-	<true/>
+	<false/>
 	<key>SUEnableAutomaticChecks</key>
 	<true/>
 	<key>SUAllowsAutomaticUpdates</key>
--- a/Sources/Support/ActivationPolicyController.swift
+++ b/Sources/Support/ActivationPolicyController.swift
@@ -1,20 +1,14 @@
 // Support/ActivationPolicyController.swift
-// Dynamically switches NSApp's activation policy so an active recording
-// session shows up in the macOS force-quit dialog (Cmd+Option+Esc).
+// Locks NSApp's activation policy to `.regular` so the app has a
+// permanent Dock presence (and shows up in Cmd+Option+Esc).
 //
-// Background: Transcripted is a menubar app, so it normally runs with
-// activation policy `.accessory` and `LSUIElement = true`. macOS hides
-// `.accessory` apps from the force-quit dialog; if the app freezes during
-// a recording, the user has to open Activity Monitor to kill it. Reported
-// pain: Taylor's meeting widget froze and the only recovery was Activity
-// Monitor.
-//
-// Fix shape: while a recording is active (meeting OR dictation), switch
-// to `.regular` so the app appears in the force-quit dialog. When all
-// sessions stop, switch back to `.accessory`. The Dock icon flickers in
-// for the duration of the recording — acceptable trade for a recovery
-// path. The menubar `NSStatusItem` is independent of activation policy
-// and stays visible the whole time.
+// History: this controller used to flip between `.accessory` (menubar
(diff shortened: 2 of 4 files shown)


r3dbars (commented):
**Verdict**: APPROVE (with one minor fix suggested)

**Summary**: Clean, well-scoped PR that gives Transcripted a permanent Dock icon by flipping `LSUIElement` to `false`, simplifying `ActivationPolicyController` to always return `.regular`, and routing Dock-icon clicks to the Settings/Home window (or onboarding if not yet completed). The change is small, correct, and the tests are properly updated to match the new always-regular semantics.

**Blockers**: None.

**Should fix**:
- Three doc comments in `TranscriptedApp.swift` still describe the old `.accessory` ↔ `.regular` switching behavior, which is now stale and misleading:
  - **Lines 69–72**: `"Switches NSApp's activation policy between .accessory (idle) and .regular (during a meeting/dictation recording)…"` — should say something like "Keeps NSApp's activation policy at `.regular` for permanent Dock presence."
  - **Lines 80–82**: `"Dock icon + menubar — start .accessory for the menubar-only feel; the controller will swap to .regular while a recording session runs…"` — should say the app starts as `.regular` for its permanent Dock icon.
  - **Lines 453–457**: `"activation-policy controller can swap the app between .accessory (idle) and .regular (recording)…"` — same issue; the controller no longer swaps.

  These are in the same file that was edited but outside the diff. Since the PR specifically changed the behavior they describe, updating them would be consistent. Not a blocker, but readers of this code will be confused by the contradictory comments.

**Nits**:
- In `ActivationPolicyDecision.desiredPolicy`, the `_ = isMeetingRecording` / `_ = isDictationRecording` lines are fine for suppressing unused-parameter warnings, but an alternative would be to use `_` parameter labels in the signature. That said, keeping the named parameters preserves API compatibility with callers, so the current approach is defensible. Take or leave.

**What looks good**:
- The `applicationShouldHandleReopen` handler is well-structured: it correctly returns `true` when windows exist (letting AppKit handle it), checks onboarding state before showing settings, and includes the `resolvedSourceApp()` call to preserve the external-app tracking that `togglePopover` also uses.
- Keeping `ActivationPolicyController` and its recording-flag wiring as a no-op rather than ripping it out is a pragmatic choice — avoids churning callers while the behavior converges.
- Tests are thorough and correctly updated. Every test case now asserts `[.regular]` as the only policy in the recorder history, which validates that no unnecessary `setActivationPolicy` calls are made.
- The PR description and test plan are excellent — clear intent, specific manual verification steps.

---

PR #567, opened 2026-04-28
Redesign settings home dashboard

r3dbars:
## Summary
- redesign Settings Home around today counts, overall stats, and recent activity
- keep Home fast by loading a small recent slice with load-more controls
- add total dictation and dictated-word counts for the overview rail

## Verification
- bash build.sh
- bash run-tests.sh
- verified Settings Home visually with Computer Use

### DIFF
--- a/Sources/Dictation/DictationTranscriptStore.swift
+++ b/Sources/Dictation/DictationTranscriptStore.swift
@@ -23,6 +23,12 @@ struct SavedDictationEntry: Identifiable, Sendable {
     }
 }
 
+struct DictationTranscriptCounts: Sendable {
+    let total: Int
+    let today: Int
+    let totalWords: Int
+}
+
 enum DictationTranscriptStore {
     private static let dictationDayPrefix = "Dictations_"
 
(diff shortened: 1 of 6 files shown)


r3dbars (commented):
## Verdict: APPROVE

**Summary**: This PR replaces the Settings Home action-tile grid and always-on readiness checklist with a Wispr Flow-style activity dashboard — welcome header, hero CTA, tabbed day-grouped timeline of recent dictations/meetings, lifetime stats rail, and a needs-attention card that only surfaces when something is actually wrong. It also adds `DictationTranscriptStore.deleteEntry()` for single-entry deletion and `savedDictationCounts()` for aggregate stats. Well-structured, no security concerns, good test coverage for the new store logic.

---

### Blockers
None.

---

### Should fix

1. **`todayMeetingCount` only counts the visible slice** (`HomeView.swift:77`) — `todayMeetingCount` is derived from `visibleMeetings` (the loaded page), not from a full scan the way dictation counts use `savedDictationCounts()`. If a user has >10 meetings today, the welcome summary will show a truncated count. Consider adding a meeting-side equivalent of `savedDictationCounts()`, or at minimum pull today's count from `statsService.todayRecordings` (which you already have).

2. **`deleteMeeting` silently swallows errors** (`TranscriptedSettingsView.swift`, the `deleteMeeting` helper) — Both `removeItem` calls use `try?`. If deletion fails (file locked, permissions), the user gets no feedback but `homeViewModel.refresh()` runs, making the meeting reappear on the next load with no explanation. A `do/catch` with `NSSound.beep()` like the dictation path would be more consistent.

3. **`deleteEntry` matches on `createdAt` alone** (`DictationTranscriptStore.swift:151-153`) — Two entries saved in rapid succession could share a timestamp. Since `SavedDictationEntry.id` already incorporates `title`, matching on both `createdAt` and `title` (or the raw heading text) would be a safer discriminator.

4. **`savedDictationCounts()` runs on every paginated load** (`RecentCaptureScanners.swift:28`) — The full file scan runs inside `RecentCaptureLoader.load(dictationLimit:meetingLimit:)`, which fires on "Load more" too. The counts don't change between paginated requests. Consider computing counts only on initial refresh or caching the result in `HomeViewModel`.

---

### Nits

1. **`homeShouldUseWideLayout` is hardcoded to `true`** — `HomeStatsStrip` (the narrow-layout fallback) is unreachable dead code. Either wire it to a `GeometryReader` width check or remove the strip view for now.

2. **`activitySubtitle` hardcodes "10"** — `"Showing the latest 10 dictations"` won't track if `initialDictationLimit` changes. Could interpolate the actual limit.

3. **Duplicate `timeFormatter`** — `HomeDictationRow` and `HomeMeetingRow` each define an identical static `DateFormatter`. A single shared formatter would trim the duplication.

---

### What looks good

- Clean `HomeViewModel` separation with proper `@MainActor` and task cancellation throughout.
- The "request N+1 to detect has-more" pagination pattern is elegant and avoids a separate count query.
- `fileStats()` scanning metadata lines without full entry parsing is a smart efficiency win.
- `contentShape(Rectangle())` on transparent button areas ensures correct hit testing.
- Delete confirmation via `HomeDeleteConfirmation` is user-safe — destructive actions require explicit confirmation.
- `NotificationCenter` post after `deleteEntry` keeps other listeners (e.g., menubar) in sync.
- Backward-compatible `load(limit:)` wrapper preserves all existing call sites.
- New test for `savedDictationCounts` covers totals, today filtering, and word counting.
- No security concerns: all changes are local file I/O and SwiftUI — no network calls, no shell execution, no unsafe memory.

Solid redesign. The should-fix items are all incremental improvements, nothing that would block shipping.

---

PR #573, opened 2026-04-29

r3dbars:
## Summary
Nightly automated security audit for Transcripted found one confirmed low-severity hardening issue in updater configuration.

### Critical
- None.

### High
- None.

### Medium
- None.

### Low
- Hardened Sparkle update configuration handling in `SparkleUpdaterController`.
  - Vulnerability: the app previously treated any non-empty `SUFeedURL` as configured, without locally requiring HTTPS or verifying that `SUPublicEDKey` was present before enabling update checks.
  - Impact: if a bundled configuration were tampered with, update checks could be pointed at an insecure or unsigned channel instead of failing closed.
  - Fix: require a valid HTTPS `SUFeedURL` and a non-empty `SUPublicEDKey` before considering Sparkle configured.

## Verification
- `bash build.sh`
- `bash run-tests.sh`

### DIFF
--- a/Sources/Observability/SparkleUpdaterController.swift
+++ b/Sources/Observability/SparkleUpdaterController.swift
@@ -319,10 +319,31 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
     }
 
     private var hasConfiguredFeedURL: Bool {
-        guard let value = Bundle.main.object(forInfoDictionaryKey: "SUFeedURL") as? String else {
+        guard let value = Bundle.main.object(forInfoDictionaryKey: "SUFeedURL") as? String,
+              let feedURL = normalizedHTTPSURL(value) else {
             return false
         }
-        return !value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+
+        // Security: require an HTTPS Sparkle feed with a non-empty EdDSA public key.
+        // If a tampered Info.plist swaps in HTTP or removes SUPublicEDKey, fail closed
+        // instead of letting update checks proceed against an unsigned/plaintext channel.
+        guard let publicKey = Bundle.main.object(forInfoDictionaryKey: "SUPublicEDKey") as? String,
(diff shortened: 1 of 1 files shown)


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR hardens `hasConfiguredFeedURL` to require a valid HTTPS scheme and a non-empty `SUPublicEDKey` before allowing Sparkle update checks to proceed. It's a clean, focused defense-in-depth measure that fails closed if the bundled configuration is tampered with. Single file, +23/−2 lines, no behavioral change for properly configured builds.

**Blockers**:
- None.

**Should fix**:
- `feedURL.host != nil` on line 335 — Foundation's `URL(string: "https://")?.host` returns `nil` on macOS 26's new URL parser, so this works correctly today. However, on older Foundation versions the same input could yield an empty string `""` rather than `nil`, which would pass this check. Since the app targets macOS 26+ this is fine in practice, but `!(feedURL.host?.isEmpty ?? true)` would be slightly more defensive if you ever want to lower the deployment target.

**Nits**:
- `normalizedHTTPSURL(_:)` doesn't reference `self` — it could be `static` or `nonisolated`. Not a functional issue since the class is `@MainActor` and this is a pure function only called from the same actor, but it signals intent more clearly.
- No unit tests added for the new validation logic. Understood that `hasConfiguredFeedURL` reads from `Bundle.main` which makes it hard to unit-test without injection, but `normalizedHTTPSURL` is a pure function that could easily get a few test cases (empty string, HTTP URL, valid HTTPS, whitespace-padded input).

**What looks good**:
- Correct fail-closed behavior — if either check is violated, updates simply don't activate rather than falling through to an insecure path.
- The HTTPS scheme comparison is case-insensitive, which handles unlikely but possible plist casing variations.
- Comment in the security guard block is warranted and well-written — explains the *why* without over-explaining.
- Consistent with the existing code style in this file and aligns with the repo's existing pattern of rejecting non-HTTPS in `SentryRuntimeConfiguration`.
- Scoped correctly — no drive-by refactoring, no unrelated changes.

---

PR #574, opened 2026-04-29

r3dbars:
## Summary
- fail fast when a meeting reaches transcription without a system-audio file
- preserve the failure in the failed-transcription queue instead of starting a doomed background task
- cover the regression with a focused task-manager test

### Transcription pipeline
- `TranscriptionTaskManager.startTranscription` could start a background transcription when the mic recording was long enough but the required system-audio file was missing.
- The pipeline would then fail later with `missingSystemAudio`, after moving through active-task state and notification paths that made the failure slower and less deterministic.
- The manager now rejects that case up front, archives the failed capture when configured, records the failure in the retry queue, and surfaces a clear user-visible failure immediately.

### Tests
- Added coverage proving a mic-only recording is rejected before any background task starts and is stored as a failed transcription with the expected error.

## Verification
- `bash build.sh`
- `bash run-tests.sh`

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
@@ -82,6 +82,24 @@ public class TranscriptionTaskManager: ObservableObject {
             return
         }
 
+        guard systemURL != nil else {
+            let errorMessage = PipelineError.missingSystemAudio.localizedDescription
+            AppLogger.pipeline.warning("Rejecting transcription — system audio capture is missing")
+            archiveFailedRecordingAudioIfConfigured(
+                micURL: micURL,
+                systemURL: nil,
+                taskId: UUID()
+            )
+            failedTranscriptionManager.addFailedTranscription(
+                micAudioURL: micURL,
+                systemAudioURL: nil,
+                errorMessage: errorMessage
+            )
+            displayStatus = .failed(message: "System audio required")
+            sendFailureNotification(errorMessage: errorMessage)
+            scheduleStatusReset(delay: 4)
+            return
+        }
 
         // Gate: reject only when every available capture track is too short.
         // Meeting recovery can produce a very short mic stub while system audio is still
--- a/Tests/TranscriptedCoreTests/TranscriptionTaskManagerMetadataTests.swift
+++ b/Tests/TranscriptedCoreTests/TranscriptionTaskManagerMetadataTests.swift
@@ -1,5 +1,6 @@
 import XCTest
 import Combine
+import AVFoundation
 import FluidAudio
 @testable import TranscriptedCore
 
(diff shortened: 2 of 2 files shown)


r3dbars:
## Code Review — PR #574

**Verdict**: APPROVE with suggestions

**Summary**: This PR adds an early guard in `startTranscription` that rejects calls with a nil `systemURL` before any background task state is mutated. Previously, the nil would pass through to the pipeline runner where it would throw `missingSystemAudio` — but only after incrementing `activeCount`/`backgroundTaskCount`, registering in `activeTasks`, and posting status updates, making the failure slower and harder to reason about. The fix is correct, well-placed, and covered by a focused test.

---

None.

---

**1. `guard systemURL != nil` → `guard let systemURL`** (`TranscriptionTaskManager.swift:85`)

`guard systemURL != nil else` is not idiomatic Swift. It checks for nil but discards the non-optional binding. After this guard, the compiler still sees `systemURL` as `URL?`, so downstream code at line 110 (`systemURL.flatMap { ... }`) and line 121 (`if let systemURL`) continues to treat it as optional even though it's guaranteed non-nil.

The idiomatic form is:
```swift
guard let systemURL else {
    // ... rejection logic ...
}
```

This shadows the parameter with a non-optional `URL`, giving the compiler full type information and preventing future edits from accidentally using the optional form inside the now-guaranteed-non-nil path.

---

### Nits (take or leave)

**2. Two sources of truth for the error string** (`TranscriptionTaskManager.swift:86,98`)

`errorMessage` is derived from `PipelineError.missingSystemAudio.localizedDescription` (the canonical string), but `displayStatus` uses a hand-rolled literal `"System audio required"`. If the `PipelineError` description is ever updated, these silently diverge. Consider deriving both from the same source, or adding a short `displayMessage` property to `PipelineError` if you want a shorter UI-facing string.

**3. Throwaway `UUID()` for archive** (`TranscriptionTaskManager.swift:91`)

The archive call gets a fresh `UUID()` that isn't correlated with the `FailedTranscription` entry's own `id`. If a user or support engineer later tries to match an archived recording directory to a failure record, the IDs won't line up. The concurrent-pipeline guard above avoids this by not archiving at all. If archiving is intentional here, consider creating the UUID once and passing it to both the archive and the failure record — or adding a brief comment explaining the intentional disconnect.

**4. `writeMonoWAV` test helper duplication** (`TranscriptionTaskManagerMetadataTests.swift:130`)

A near-identical `writeMonoWAV` private helper exists in `MicRecordingFileMergerTests.swift` (different signature but same AVFoundation boilerplate). Not worth blocking over, but if a shared `AudioTestFixtures` emerges later, these are natural candidates.

---

### What looks good

- **Guard placement is correct**: after the concurrent-pipeline check, before the duration check. A nil `systemURL` is rejected before any state is mutated.
- **Test is well-designed**: the 2.5s WAV duration is above the 2.0s minimum, so it exercises the system-audio guard specifically rather than being caught by the duration gate. Assertions cover all the important state: no active tasks, failure recorded, correct error message, correct display status.
- **Error taxonomy is consistent**: `missingSystemAudio` is already classified as non-retryable in `PipelineError.isRetryable`, and the `FailedTranscription` model correctly propagates that — so the queue entry shows up in the UI as a permanent failure the user can see and dismiss, not as a false retry candidate.
- **The notification and archive paths follow the existing async-catch pattern** (lines 184–197), which is the right template for "something failed and the user should know."
- **The pipeline runner's own `guard let systemURL` at `TranscriptionPipelineRunner.swift:22`** is preserved as a belt-and-suspenders check, which is the right call — defense in depth costs nothing.

---

PR #576, opened 2026-04-29

r3dbars:

### DIFF
--- a/Sources/Meeting/CLAUDE.md
+++ b/Sources/Meeting/CLAUDE.md
@@ -8,7 +8,9 @@
 
 - `FailedMeetingPresentation.swift` — maps `FailedTranscription` into `FailedMeetingItem` view-models with human-readable titles and retry metadata
 - `MeetingAudioInactivityDetector.swift` — detects prolonged audio silence during meetings and emits warning/cleared events so the UI can prompt the user to confirm the recording is still needed
-- `MeetingCaptureBridge.swift` — `@MainActor` wrapper around core `Audio` that converts start/stop into async flows, waits for both live capture and system-audio-file readiness, mirrors live levels, and exposes the meeting-mic processing preference to core capture
+- `MeetingCaptureBridge.swift` — `@MainActor` wrapper around core `Audio` that converts start/stop into async flows, waits for both live capture and system-audio-file readiness, and mirrors live levels for the UI
+- `MeetingCaptureBridge+LivePreview.swift` — bridge extension for recording health snapshots plus mic/system live-preview buffer forwarding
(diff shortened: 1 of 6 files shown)


r3dbars:
## Code Review: PR #576

**Verdict**: APPROVE

**Summary**: This PR extracts duplicated continuation/timeout/attemptID state management from `MeetingCaptureBridge` into a generic `MeetingCaptureAttempt<Output>` helper, moves `CaptureStopResult` to its own file, splits live-preview and health-info methods into a focused extension, and syncs three CLAUDE.md files with recent codebase additions. The refactoring is clean, behavior-preserving, and reduces real duplication without introducing unnecessary abstraction.

**Blockers**: None.

**Should fix**:

- **`MeetingCaptureAttempt` lacks `@MainActor` annotation** (`MeetingCaptureSupport.swift`): The class manages `CheckedContinuation` state with no internal synchronization. This is safe *today* because it's only used as a private stored property of the `@MainActor` `MeetingCaptureBridge`, so all access is serialized on the main actor. However, the type itself has no isolation annotation — if someone reuses it from a non-main-actor context, it becomes a data race. Adding `@MainActor` to the class declaration would make the thread-safety invariant compiler-enforced rather than convention-enforced.

**Nits**:

- `setTimeoutTask` being a separate call after `begin` creates a two-step protocol callers must remember. A combined `begin(_:timeout:)` would be cleaner in theory, but given the timeout closures capture `[weak self]` from the bridge, the split is a reasonable pragmatic choice for just two call sites. Take or leave.

**What looks good**:

- The generic `MeetingCaptureAttempt` is well-shaped: four methods (`begin`, `setTimeoutTask`, `reset`, `resetIfCurrent`), no extras, no over-abstraction. It collapses six stored properties and two identical `resetXxxAttempt()` methods into a reusable pattern.
- Double-resume prevention is correct: `reset()` atomically extracts and nils the continuation, and `resetIfCurrent` gates on `attemptID`, so whichever of normal-completion vs. timeout fires second gets `nil` back and noops.
- The `deinit` in both `MeetingCaptureAttempt` (cancels timeout) and `MeetingCaptureBridge` (resumes outstanding continuations) properly cleans up, preventing leaked tasks or hanging awaits.
- File split is logical: support types, live-preview extension, and core lifecycle each have a single concern.
- All files referenced in the CLAUDE.md updates (`LockedFileAppender.swift`, `UpdateFailureKind.swift`, `HomeView.swift`, test files) exist on disk. Documentation is accurate.
- `[weak self]` usage in all closures and `Task` blocks is consistent and correct. No retain cycles.

---

PR #592, opened 2026-04-29
Optimize settings capture refresh

r3dbars:
## Summary
- removes the duplicate Settings first-open refresh
- limits recent capture scans to Home, Meetings, and Dictation pages
- adds a small policy test so the routing stays explicit

## Measurements
- live STT timing looked stable in 1.1.25 vs 1.1.24
- Settings Home refresh on current local data dropped from about 25.84 ms median to about 12.8 ms median
- scaled synthetic dataset dropped from about 467.34 ms median to about 301.85 ms median

## Verification
- bash build-deps.sh --force
- bash run-tests.sh: 989 passed
- bash build.sh


r3dbars:
## Code Review — PR #592: Optimize settings capture refresh

**Verdict**: APPROVE

**Summary**: This PR removes a duplicate `.onAppear` that was calling `refreshState()` twice on initial settings open, and introduces a routing policy so `refreshRecentCaptures()` only does work on pages that actually display capture data (Home, Meetings, Dictation). The approach is clean, the performance numbers are solid (~50% reduction in refresh time), and the test coverage is thorough.

---

### Blockers
None.

### Should fix

**`SettingsRecentCaptureRefreshMode.none` shadows `Optional.none`** — This is a well-known Swift footgun. If someone later wraps the return type in an optional (e.g. during a refactor), `.none` in a switch will silently match `Optional.none` instead of the enum case, causing a subtle bug with no compiler warning. Consider renaming to `.skip` or `.noRefresh`.

### Nits

- **Unanchored `Task` in `.homeDashboard`** (line 1496): The `Task { @MainActor in await statsService.refreshStats() }` isn't assigned to `recentCaptureRefreshTask`, so it won't be cancelled by subsequent `refreshRecentCaptures()` calls or by `.onDisappear`. This is pre-existing behavior (not a regression), but the refactor was a natural opportunity to store it for cancellation. Low priority since `refreshStats()` is presumably idempotent and fast.

### What looks good

- **Removing the duplicate `.onAppear`** is the right call. `.task(id: navigation.presentationID)` already fires on first appearance, so the old code was running `refreshState()` (and therefore `refreshRecentCaptures()`) twice on every settings open. Clean fix, directly explains the timing improvement.
- **The policy enum with exhaustive switch** is a nice pattern — if a new `TranscriptedSettingsPage` case is added, the compiler will force updating the policy. Much better than an `if` chain that silently falls through.
- **Test coverage** hits all three branches and enumerates every page in the `.none` case, so nothing can slip through unnoticed.
- **`StatsService.shared` → `statsService`** is a good consistency fix — uses the already-injected `@ObservedObject` property instead of reaching for the singleton directly.
- **Test manifest and `run-tests.sh` updates** are included — no loose ends.

---

PR #601, opened 2026-04-29
Fix release blocker QA findings

r3dbars:
## Summary
- cancel stale meeting audio starts so timeout paths cannot publish a hidden recording later
- add an update-check watchdog so the About UI returns to an actionable state if Sparkle never finishes a cycle
- mark zero-buffer system audio captures as degraded, and clean unauthorized agent todo labels instead of leaving stale queue items

## Verification
- bash build-deps.sh --force
- bash build.sh
- bash run-tests.sh
- swift test
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift run transcripted-qa validate-all --format json
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift run transcripted-qa round-trip
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift run transcripted-qa check-health
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift run transcripted-qa validate-logs
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift run transcripted-qa validate-database
- Computer Use: relaunched patched app, checked About update fallback, Meetings, Models, and Agent screens

## Notes
`validate-all` still reports the known local stale artifact drift for `Call_2026-04-18_14-43-40.json` / `transcripted.json`; fresh round-trip fixtures pass.

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -88,9 +88,7 @@ final class MeetingCaptureBridge: ObservableObject {
                 self.errorMessage = AudioCaptureStartState.timeoutFailureMessage(
                     existingErrorMessage: self.errorMessage
                 )
-                if self.audio.isRecording {
-                    self.audio.stop()
-                }
+                self.audio.stop()
                 continuation.resume(returning: false)
             })
         }
--- a/Sources/Observability/SparkleUpdaterController.swift
+++ b/Sources/Observability/SparkleUpdaterController.swift
@@ -71,6 +71,8 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
     private var pendingImmediateInstallHandler: (() -> Void)?
     private var pendingImmediateInstallVersion: String?
     private var didTrackCurrentUpdateCycleFailure = false
+    private var observedUpdateCheckTimeoutTask: Task<Void, Never>?
+    private static let observedUpdateCheckTimeoutNanoseconds: UInt64 = 30_000_000_000
 
     override init() {
         super.init()
@@ -200,9 +202,11 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
     private func beginObservedUpdateCheck() {
         switch updateStatus.state {
         case .updateAvailable, .downloading, .readyToInstall:
+            cancelObservedUpdateCheckTimeout()
             syncReadiness(from: updaterController.updater)
         default:
             setUpdateStatus(.checking, canCheckForUpdates: updaterController.updater.canCheckForUpdates)
+            scheduleObservedUpdateCheckTimeout()
         }
     }
 
@@ -258,6 +262,7 @@ final class SparkleUpdaterController: NSObject, ObservableObject {
     }
 
     private func markNoUpdateAvailable(from updater: SPUUpdater) {
+        cancelObservedUpdateCheckTimeout()
         if updateStatus.availableUpdateVersion != nil {
             trackUpdateCheckFinished(
                 result: "no_change",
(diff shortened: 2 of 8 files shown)


r3dbars:
## Code Review — PR #601: Fix release blocker QA findings

**Verdict**: APPROVE

**Summary**: This PR fixes three distinct release blockers: (1) a race condition where a stale audio start could silently resume recording after the user stopped it, (2) the Sparkle update check UI getting permanently stuck in "Checking…" if Sparkle never completes a cycle, and (3) zero-buffer system audio captures being falsely reported as "excellent" quality. All three fixes are minimal, well-scoped, and backed by tests. The agent-todo-runner cleanup for unauthorized labels is a nice operational hygiene addition.

---

### Blockers
None.

### Should fix

**1. Redundant state reset before `stop()` in `finishSuccessfulStartIfCurrent` (`Audio.swift:767-769`)**

When a stale start is detected, the method sets `isRecording = false` and `isStarting = false`, then immediately calls `stop()`. But `stop()` itself dispatches `isRecording = false` and `isStarting = false` on `DispatchQueue.main.async`. Since `finishSuccessfulStartIfCurrent` is `@MainActor`, the sync assignments happen first, and the async dispatch in `stop()` redundantly sets them again on the next run-loop tick. This isn't a bug — just unnecessary work. Consider either removing the two assignments before `stop()`, or having `stop()` skip the UI-state reset when it detects it's already clean. Low priority since the current code is safe.

### Nits

**1. `try? await Task.sleep` cancellation pattern (`SparkleUpdaterController.swift:321`)**

The `try? await Task.sleep(nanoseconds:)` followed by `!Task.isCancelled` guard is correct and idiomatic. Just noting that an alternative is `try await Task.sleep(nanoseconds:)` without `try?`, letting the thrown `CancellationError` exit the task body naturally — which would make the `Task.isCancelled` check unnecessary. Either approach works; the current one is more explicit about intent, which is fine for a watchdog.

**2. Repo contract test is brittle (`RepoCommandContractTests.swift:53`)**

The new test asserts on exact string matches of Ruby source lines (`issues.select { |issue| active_issue?(issue) || unauthorized_active_issue?(issue) }`). Any whitespace or formatting change to the Ruby file will break this test. This matches the existing pattern in the file, so it's consistent — but worth being aware of.

### What looks good

- **Generation-based stale-start cancellation** (`Audio.swift`): This is the strongest change in the PR. Capturing `recordingSessionGeneration` before the async start and validating it on completion is a clean, race-free way to prevent phantom recordings. The generation counter was already in place for recovery work, so extending it here is a natural fit with zero new infrastructure.

- **Sparkle timeout watchdog** (`SparkleUpdaterController.swift`): The cancel-on-every-terminal-path pattern is thorough — `cancelObservedUpdateCheckTimeout()` is called in `markNoUpdateAvailable`, `markUpdateCheckFailed`, `markUpdaterIdle`, `didFindValidUpdate`, and `didFinishUpdateCycleFor`. The `[weak self]` capture and triple-guard (`self` alive, not cancelled, still in `.checking` state) make this defensively correct.

- **Buffer success rate semantic fix** (`SCKAudioCapture.swift`, `SystemAudioCapture.swift`): Changing the zero-buffer default from `1.0` to `0.0` is a one-line fix that corrects a real data integrity issue — transcripts were getting "excellent" quality metadata when system audio capture produced nothing. The new test with `StubSystemAudioCapture` cleanly validates the full path from capture engine through `RecordingHealthInfo.from`.

- **Unconditional `audio.stop()` in timeout** (`MeetingCaptureBridge.swift`): Removing the `isRecording` guard is the right call. The old code would skip cleanup if the timeout fired while audio was still in the `isStarting` phase — exactly the window where the stale-start bug lived. `Audio.stop()` is already idempotent, so calling it unconditionally is safe.

- **Agent runner unauthorized issue cleanup** (`agent-todo-runner.rb`): Simple and correct. `unauthorized_active_issue?` is the logical complement of `active_issue?` on the `allowed_author?` dimension, and `handle_issue` already routes unauthorized issues to `skip_unauthorized_issue` which strips labels and posts an explanation.

Solid release-blocker fixes. Clean, minimal, well-tested.

---

PR #619, opened 2026-04-30

r3dbars:
## Summary
Automated nightly security audit focused on filesystem handling, SQL injection, input validation, secrets, unsafe operations, race conditions, Sparkle updater security, temporary files, and model download security.

### Medium
- Rejected relative capture-library and transcript save paths sourced from preferences or other persisted inputs. Before this fix, a tampered relative path could be resolved against the process working directory, which could redirect Transcripted writes and permission probes outside the intended user-selected absolute location.

### Low
- No additional in-scope vulnerabilities were confirmed in this review. Reviewed areas included SQLite parameterization in `SpeakerDatabase` and `StatsDatabase`, Sparkle HTTPS/appcast expectations, temporary file permissions and cleanup, and HuggingFace model download transport/integrity controls.

## Fixes
- Added explicit absolute-file-path validation for capture-library preferences.
- Added matching absolute-path enforcement in `RecordingValidator.validateSavePath(_:)`.
- Added regression coverage for relative-path rejection in both storage helper and core validator tests.

## Verification
- `bash build.sh`
- `bash run-tests.sh`

All three commands completed successfully.


r3dbars:
## Code Review — PR #619

**Verdict**: REQUEST CHANGES

**Summary**: This automated security audit adds absolute-path enforcement to `isSafeCaptureLibraryURL` and `RecordingValidator.validateSavePath`, plus regression tests and CLAUDE.md housekeeping. The threat model (relative paths from tampered UserDefaults redirecting writes) is valid and worth fixing. However, the implementation has a significant gap: the guards operate on `URL` objects *after* `URL(fileURLWithPath:)` has already resolved relative paths to absolute ones, making the checks ineffective for the primary attack vector.

---

**1. `URL(fileURLWithPath:)` silently resolves relative paths — guards are dead code for the stated threat**

On all Apple platforms, `URL(fileURLWithPath: "relative-capture-root", isDirectory: true)` resolves the relative string against the process working directory, producing an **absolute** file URL (e.g. `file:///current/cwd/relative-capture-root/`). After construction:

- `url.isFileURL` → `true`
- `url.path` → `"/current/cwd/relative-capture-root"` (starts with `/`)

This means the guard:
```swift
guard url.isFileURL, url.path.hasPrefix("/") else { return false }
```
**never triggers** for the primary attack vector — a relative string in `UserDefaults` that flows through `URL(fileURLWithPath:)` in `captureLibraryURL(userDefaults:fileManager:)`, `validateRecordingConditions()`, `resolvedSaveDirectory(paths:)`, and `TranscriptSaver.defaultSaveDirectory`.

The fix should validate the **raw string** from UserDefaults *before* URL construction. For example, in `captureLibraryURL`:

```swift
if let customPath = userDefaults.string(forKey: captureLibraryLocationKey)?
    .trimmingCharacters(in: .whitespacesAndNewlines),
   !customPath.isEmpty {
    guard customPath.hasPrefix("/") else {
        return fileManager.transcriptedDefaultCaptureLibraryDir
    }
    let candidate = URL(fileURLWithPath: customPath, isDirectory: true)
    // ... existing isSafeCaptureLibraryURL check ...
```

Same pattern needed in `RecordingValidator.validateRecordingConditions`, `resolvedSaveDirectory`, and `TranscriptSaver.defaultSaveDirectory` — anywhere a `UserDefaults` string becomes a `URL(fileURLWithPath:)`.

The URL-level guards can stay as defense-in-depth for the public API (callers could pass URLs constructed via `URL(string:)` etc.), but they shouldn't be the primary defense.

**2. Tests likely don't verify what they claim**

Both new tests construct their "relative" URL via `URL(fileURLWithPath: "relative-capture-root", isDirectory: true)`, which Foundation resolves to an absolute path. So:

- `testRecordingValidatorRejectsRelativePaths` — the URL passed to `validateSavePath` would have an absolute path, pass the guard, and return `.success` (assuming CWD isn't under a forbidden prefix). The `XCTAssertFalse(result.isValid)` assertion would **fail**.
- The storage paths test `"reject relative capture folders"` — `setCaptureLibraryURL(relativeRoot)` would see an absolute file URL, `isSafeCaptureLibraryURL` would return `true`, and the URL would be persisted. The `assertEqual(..., nil, ...)` assertion would **fail**.

If these tests are meant to verify the string-level validation (per the fix above), they should test the code path that reads a relative string from UserDefaults, not one that constructs a URL from it. For example:

```swift
// Test the actual attack vector: a relative string in UserDefaults
UserDefaults.standard.set("relative-capture-root", forKey: TranscriptedStoragePreferences.captureLibraryLocationKey)
// Verify it falls back to default
assertEqual(
    FileManager.default.transcriptedCaptureLibraryDir,
    FileManager.default.transcriptedDefaultCaptureLibraryDir,
    ...
)
```

---

**3. CLAUDE.md file-count bumps appear unrelated to this PR**

The first commit bumps `Sources/TranscriptedCore/CLAUDE.md` from 61→62 files and `Sources/UI/CLAUDE.md` from 48→49 files. The actual counts on `main` already match the *new* values (17 Audio files, 62 total Core files, 49 UI files), and `SettingsRecentCaptureRefreshPolicy.swift` already exists on `main`. These are documentation catch-ups, not changes introduced by this security audit. They should be in a separate commit/PR to keep the security fix atomic and easy to revert if needed.

---

### Nits (take or leave)

- The security comments on the guards are reasonable for a security-sensitive code path. No objection to keeping them.

---

### What looks good

- **Threat model is correct**: relative paths in UserDefaults resolving against an unpredictable CWD is a real concern worth hardening against.
- **Defense-in-depth at two layers**: checking both `isSafeCaptureLibraryURL` (preference persistence) and `validateSavePath` (recording validation) covers the write path from both entry points.
- **Test structure is good**: saving/restoring UserDefaults state with `defer`, testing both the "not persisted" and "falls back to default" invariants — the test *shape* is right, just the URL construction needs fixing.
- **CLAUDE.md documentation updates** are accurate relative to the current file tree.

---

PR #620, opened 2026-04-30

r3dbars:
## Summary
This nightly review focused on changes from the last 7 days, especially meeting capture/transcription paths, actor isolation, and newly expanded settings/home surfaces.

### Meeting capture diagnostics
- preserved the real stop-timeout result when a meeting recording is cancelled
- included that timeout state in cancellation diagnostics/analytics instead of hardcoding `false`

### Settings dashboard loading
- fixed cancelled recent-capture loads so the Home dashboard does not leave loading spinners stuck on
- made the background recent-capture loader honor task cancellation between expensive scans, avoiding wasted work from superseded refreshes

## Review notes
- confirmed `Audio.swift` and `SystemAudioCapture.swift` remain intentionally non-`@MainActor`
- checked recent transcription/meeting pipeline changes for clear logic regressions and only fixed issues that were verified as real
- noted several large files for ongoing maintainability attention, but did not make cosmetic-only splits

## Verification
- `bash build.sh`
- `bash run-tests.sh`


r3dbars:
## Code Review

**Verdict**: APPROVE (with one suggestion worth addressing)

**Summary**: This PR fixes two real bugs — hardcoded `stopTimedOut: false` in meeting cancellation diagnostics, and stuck loading spinners when the Home dashboard's refresh task is cancelled. It also adds cooperative cancellation to `RecentCaptureLoader` so superseded refreshes bail out between expensive file scans. All four changes are well-scoped, correct in intent, and the security surface is unchanged.

---

### Blockers
None.

---

### Should fix

**`LoadTaskBox` has an unsynchronized data race** (`Sources/UI/Shared/RecentCaptureScanners.swift:77-79`)

`LoadTaskBox` is `@unchecked Sendable` with a bare `var task` — no lock, no atomics. The `onCancel` closure reads `.task` concurrently with the operation closure writing it:

```swift
// operation closure (some executor)
taskBox.task = task          // write

// onCancel closure (arbitrary thread, possibly concurrent)
taskBox.task?.cancel()       // read
```

In practice the worst outcome is the cancel "misses" and the detached task runs to completion (the caller has its own cancellation guard), but the unsynchronized read/write is undefined behavior under the Swift memory model. Wrapping the property in `OSAllocatedUnfairLock` (macOS 13+) or `NSLock` would close the gap cleanly:

```swift
private final class LoadTaskBox: @unchecked Sendable {
    private let lock = NSLock()
    private var _task: Task<RecentCaptureSnapshot, Never>?
    var task: Task<RecentCaptureSnapshot, Never>? {
        get { lock.withLock { _task } }
        set { lock.withLock { _task = newValue } }
    }
}
```

---

**Cancelled task can briefly reset a successor's loading flags** (`Sources/UI/Settings/HomeView.swift:69-73`)

When `loadCurrentLimits` is called again while a previous load is in flight:
1. Old task is cancelled (line 58)
2. New loading flags are set (lines 59-60)
3. New task starts and suspends at `await`
4. Old task resumes → hits `guard !Task.isCancelled` → sets `isLoading = false`
5. Briefly, no loading indicator is visible even though the new task is still working

Since both run on `@MainActor`, there's a real (though brief) window where the old cancelled task's cleanup clobbers the new task's loading state. This is still much better than the original stuck-spinner bug — just noting it as a potential follow-up. A generation counter or checking `refreshTask === Task.currentTask` equivalent would close it.

---

### Nits

- `MeetingSessionController.swift:712` — the intermediate `let files = (micURL: stopResult.micURL, systemURL: stopResult.systemURL)` tuple exists only to keep subsequent `files.micURL` references unchanged. Using `stopResult.micURL` / `stopResult.systemURL` directly would remove a line but would touch more downstream references, so keeping it is a fine trade-off for a minimal diff.

- `RecentCaptureScanners.swift:50-52` — the fourth `guard !Task.isCancelled` check (after computing `dictationCounts`, before constructing the snapshot) could be dropped. Constructing the return struct is trivial compared to the three I/O calls already guarded. Not wrong to have it, just not buying much.

---

### What looks good

- **Diagnostics fix is the star of this PR.** Replacing the hardcoded `stopTimedOut: false` with `stopResult.didTimeOut` means cancellation analytics will now correctly surface timeout states — directly useful for debugging audio-stop hangs in the field.
- **Return type change in `stopAndDiscardFiles`** is the right call. Returning the full `CaptureStopResult` instead of stripping it to a tuple preserves information without changing the internal flow.
- **Cooperative cancellation in `RecentCaptureLoader`** is textbook. Interleaving `Task.isCancelled` checks between expensive synchronous scans is exactly how you make a `Task.detached` block responsive to cancellation.
- **`emptySnapshot()` helper** keeps the four cancellation return sites DRY and readable.
- Build verification steps in the PR description are appreciated.

---

PR #622, opened 2026-04-30

r3dbars:

### DIFF
--- a/Sources/UI/Settings/HomeView.swift
+++ b/Sources/UI/Settings/HomeView.swift
@@ -586,13 +586,23 @@ struct HomeRowMoreMenuButton: NSViewRepresentable {
 
 // MARK: - Activity rows
 
-struct HomeDictationRow: View {
-    let entry: SavedDictationEntry
+private enum HomeActivityRowFormatting {
+    static let timeFormatter: DateFormatter = {
+        let f = DateFormatter()
+        f.locale = .current
+        f.dateFormat = "h:mm a"
+        return f
+    }()
+}
+
+private struct HomeActivityRowShell<Content: View>: View {
+    let timeString: String
     let isCopied: Bool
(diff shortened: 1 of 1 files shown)


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR extracts a shared `HomeActivityRowShell<Content>` generic view from the duplicated row layout in `HomeDictationRow` and `HomeMeetingRow`, consolidates two identical `DateFormatter` instances into a single `HomeActivityRowFormatting` namespace, and DRYs up the activity section pattern in `HomeActivityTabsCard` with an `activitySection` helper. Clean, well-scoped refactoring — net -2 lines, single file, no behavioral changes.

**Blockers**: None.

**Should fix**: None.

**Nits** (take or leave):
- In the old `HomeMeetingRow`, only `Text(item.title)` carried `.frame(maxWidth: .infinity, alignment: .leading)`. The new shell now applies that same frame modifier to the *outer* `content()` HStack as well (in addition to the Text still having it). Functionally identical since the Text already forced the HStack to fill, but it's a redundant double-expansion. Cosmetic only — not worth a follow-up unless you're already in the file.

**What looks good**:
- `HomeActivityRowShell` is the right abstraction — it captures exactly the shared chrome (time label, button wrapper, hover state, action buttons, padding, background) without leaking content concerns.
- Using a caseless `enum` for the formatter namespace is idiomatic Swift.
- The `activitySection` generic helper in `HomeActivityTabsCard` eliminates a real copy-paste pattern (VStack + grouped list + conditional load-more button) cleanly.
- Both row types remain easy to read and modify independently despite sharing a shell.
- `private` visibility on the shell and formatter is correct — these are file-scoped implementation details.
- No security concerns: pure SwiftUI view code with no I/O, network, filesystem, or permissions APIs.

---

PR #625, opened 2026-04-30
Add audio diagnostics and support feedback flow

r3dbars:
## Summary
- Add privacy-safe dictation audio route telemetry for Bluetooth/source changes and recovery outcomes.
- Add runtime diagnostics, dirty-shutdown context, and a one-click support diagnostic Sentry event.
- Route feedback to help@transcripted.app and add a cleaner Settings Support page from the menu bar flow.

## Verification
- bash build.sh
- bash run-tests.sh (1171/1171)


r3dbars:
## PR Review: Add audio diagnostics and support feedback flow

**Verdict**: APPROVE

**Summary**: This PR adds three well-scoped features: (1) privacy-safe audio route telemetry for dictation Bluetooth/device-change events, (2) runtime diagnostics with dirty-shutdown detection and session stage tracking, and (3) a dedicated Settings Support page with email feedback and one-click Sentry diagnostic events. The code is clean, security-conscious, and thoroughly tested. All telemetry payloads use bucketed values, device classes (not names), and pass through the existing sanitizer/allowlist pipeline.

---

None.

---

1. **Dead code: `sendFeedback(logger:)` and `feedbackEmailURL(logger:)`** — `TranscriptedSupportActions` retains the old `sendFeedback(logger:)` and `feedbackEmailURL(logger:)` overloads, but nothing calls them anymore (all callers now use the `appState:` variants). These should be removed to avoid confusion.

2. **Dead code: `copyDiagnostics` action** — `TranscriptedSettingsActions.copyDiagnostics` is declared, wired up in `TranscriptedApp.swift`, and implemented in `TranscriptedSupportActions`, but the Support page UI never calls it. Either add the "Copy diagnostics" button to the Support page or remove the plumbing.

3. **Redundant key overwrites in `dictationRouteDiagnosticsContext`** — Lines 1305–1306 of `ParakeetEngine.swift` overwrite `recovering` and `format_ready` with the same values already set by `dictationRouteAnalyticsContext` (lines 1331/1335). Only `generation` is actually new. Consider either removing the redundant overwrites or adding a comment explaining the intent (e.g., ensuring Diagnostics context always reflects current recovery state even if the analytics context shape changes).

---

**Nits** (take or leave):

1. **`output_device_class` duplicates `default_output_class`** in `dictationRouteAnalyticsContext` — both are populated with `defaultOutputClass`. If they're semantically the same, consider dropping one to reduce payload size. If they're intended to diverge later, a brief comment would help.

2. **`SupportDiagnosticsBundle.sentryContext` forwards `audioRoute` and `runtime` dicts without key-level filtering** — The values are safe by construction today (device classes, booleans, bucketed durations), but a future change adding a raw device name to the `audioRoute` dictionary would silently ship it to Sentry. An explicit key allowlist in `sentryContext` would add defense-in-depth.

---

**What looks good**:

- **Privacy discipline is excellent.** Every telemetry path uses device classes (not names), bucketed durations, and boolean flags. The `AnalyticsPayloadSanitizer.redact` pass on clipboard/email content, the `SentryPayloadSanitizer` pass on Sentry events, and the explicit `AnalyticsEventPolicy` / `SentryEventPolicy` allowlists all work together correctly.
- **RuntimeDiagnostics design is solid.** The heartbeat marker + dirty-shutdown detection pattern is well-suited for catching force-quit and silent-exit scenarios. The `@MainActor` constraint on the class combined with `[weak self]` in the timer closure avoids both threading issues and retain cycles.
- **Test coverage is thorough.** New tests cover the runtime diagnostics store round-trip, the support diagnostics bundle (including PII redaction), analytics event policy for all new events, Sentry event policy additions, and the updated feedback email builder. The 7-commit history shows progressive refinement.
- **The `isLikelyBluetoothHandsFreeProfile` heuristic** is a nice touch — detecting HFP via the asymmetric sample rate signature (≤24kHz input + ≥44.1kHz output) without needing to inspect device names.
- **File write safety** in `RuntimeDiagnosticsStore.save` — atomic writes, private directory creation, and owner-only file permissions.


---

PR #632, opened 2026-05-01

r3dbars:
## Summary
- guard cleanup-time file deletion so Transcripted only unlinks app-managed scratch audio and speaker clip files
- add regression tests covering imported-audio rejection and pending speaker-naming cleanup for out-of-sandbox paths
- verified with `bash build-deps.sh --force`, `bash build.sh`, and `bash run-tests.sh`

### Critical
- None.

### High
- None.

### Medium
- **Out-of-sandbox cleanup deletion in transcription cleanup paths**
  - `TranscriptionTaskManager` and speaker-naming cleanup previously removed whichever file URLs were carried in request state.
  - In failure or shutdown paths, a caller-supplied imported audio URL or tampered in-memory naming request could cause Transcripted to delete files outside its managed scratch directories.
  - Fixed by canonicalizing URLs and refusing cleanup unless the target is contained inside explicit Transcripted-managed cleanup roots.

### Low
- None.

## Audit notes
Reviewed and found no confirmed vulnerabilities in:
- path sandboxing for failed transcription storage and app-owned directories
- SQL handling in `SpeakerDatabase` and `StatsDatabase`
- Sparkle feed/signing configuration
- model download HTTPS/integrity checks
- hardcoded secret scanning
- unsafe-operation and race-condition hotspots inspected during this audit

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -230,6 +230,7 @@ final class MeetingSessionController: ObservableObject {
             diarization: services.diarization,
             speakerStore: services.speakerStore,
             speakerClipsDirectory: storagePaths.speakerClips,
+            cleanupDirectories: [storagePaths.audioCaptures, storagePaths.speakerClips],
             retainedAudioDirectoryProvider: { MeetingStoragePaths.audioArchiveFolder },
             statsStore: statsDatabase
         )
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
@@ -484,7 +484,7 @@ public class TranscriptionTaskManager: ObservableObject {
         notifier.notifyTranscriptionFailed(errorMessage: errorMessage)
     }
 
-    private func removeRecordingFile(_ url: URL, label: String) {
+    nonisolated private func removeRecordingFile(_ url: URL, label: String) {
         // Security: only delete scratch files inside Transcripted-managed cleanup roots.
         // `startImportedTranscription` accepts a URL from the caller, so without a containment
         // check a misuse or tampered in-memory request could unlink arbitrary user files.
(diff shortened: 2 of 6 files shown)


r3dbars:
## Code Review: PR #632 — security: nightly security audit findings [automated]

**Verdict**: REQUEST CHANGES

**Summary**: This PR adds path-containment checks to `TranscriptionTaskManager`'s file-deletion methods so that cleanup only removes files inside explicitly configured Transcripted-managed directories (`audioCaptures`, `speakerClips`). The security fix is well-targeted — imported-audio URLs and in-memory speaker-naming requests could previously delete arbitrary files — and the regression tests cover both the imported-audio rejection and pending-naming cleanup paths. One concurrency issue needs fixing before merge.

---

**1. `@MainActor`-isolated instance methods called from a background dispatch queue**

`handleNamingComplete` dispatches to `DispatchQueue.global(qos: .utility)` and calls:
- `self?.cleanupSpeakerClips(clips)` (lines 61, 82, 137 in `SpeakerNamingCoordinator.swift`)
- `self?.removeManagedCleanupFile(...)` (lines 62-63, 84-85, 139-140)

These are instance methods on `@MainActor public class TranscriptionTaskManager`. The **old code** correctly used `nonisolated private static func removeTemporaryAudioFile` and `SpeakerClipExtractor.cleanupClips` (static on a non-actor class) — both nonisolated by definition.

The new code regresses this: it calls `@MainActor`-isolated instance methods from a GCD background queue without dispatching to the main actor. With Swift 5.9's default concurrency checking this compiles, but it's technically a data race and will fail to compile under strict concurrency / Swift 6.

**Fix**: Mark `removeRecordingFile`, `removeManagedCleanupFile`, `isSafeCleanupURL`, and both `cleanupSpeakerClips` overloads as `nonisolated`. This is safe because they only read `cleanupDirectories` (a `let` constant, which is implicitly `nonisolated` on `@MainActor` classes in Swift 5.10+) and call thread-safe APIs (`FileManager.default.removeItem`, `AppLogger`).

---

**2. `filePath == directoryPath` allows deleting the cleanup root directory itself**

In `isFile(_:containedIn:)`:
```swift
return filePath == directoryPath || filePath.hasPrefix(normalizedDirectoryPath)
```

The first condition means a URL whose resolved path equals the cleanup root directory would pass the containment check. In practice `removeRecordingFile` is only called with file URLs, but defense-in-depth says this should be just `filePath.hasPrefix(normalizedDirectoryPath)`.

**3. `SpeakerClipExtractor.cleanupClips` is now dead code**

Both overloads of `SpeakerClipExtractor.cleanupClips` (`[ClipResult]` and `[SpeakerNamingEntry]`) are no longer called from anywhere. They should be removed to avoid confusion about which cleanup path is active.

---

### Nits (take or leave)

**4. `canonicalFileURL` and `canonicalDirectoryURL` are identical**

Both are `url.standardizedFileURL.resolvingSymlinksInPath()`. Consider collapsing them into a single `private static func canonicalURL(_ url: URL) -> URL`. If the intent is to leave room for future divergence (e.g., appending `/` for directories), a comment would clarify that.

---

### What looks good

- **Path canonicalization is solid**: `standardizedFileURL.resolvingSymlinksInPath()` handles `..` traversal and symlink-based escapes. Canonicalizing cleanup roots at init time and file URLs at check time is the right approach.
- **Default fallback preserves backwards compatibility**: `cleanupDirectories ?? [speakerClipsDirectory]` means callers that don't pass the new parameter keep working with a safe default.
- **Regression tests are well-designed**: Both tests create real files outside the cleanup roots and verify they survive cleanup — directly testing the security property rather than implementation details.
- **Clean migration of call sites**: All `SpeakerNamingCoordinator` cleanup paths are consistently routed through `removeManagedCleanupFile`, and the old `removeTemporaryAudioFile` static method is fully removed with no leftover references.
- **The PR description is excellent**: Clear finding classification, explicit audit scope, and verified build commands.

---

PR #633, opened 2026-05-01

r3dbars:
## Summary
Nightly automated code review focused on commits from the last 7 days, with extra attention on transcription, audio capture, meeting capture, threading, and failed-transcription recovery paths.

### Threading
- Marshaled `SystemAudioCapture.errorMessage` updates onto the main thread so `@Published` UI state is not mutated from non-main capture/startup paths.

### Failed transcription recovery
- Fixed `FailedTranscriptionManager`'s retained-audio allowlist to accept archived meeting audio stored under `transcripts/audio/...`.
- Added a regression test covering accepted retained archive paths and rejection of a sibling `audio/` directory outside the intended root.

## Verification
- `bash build.sh`
- `bash run-tests.sh`

### DIFF
--- a/Sources/TranscriptedCore/Audio/SCKAudioCapture.swift
+++ b/Sources/TranscriptedCore/Audio/SCKAudioCapture.swift
@@ -113,7 +113,7 @@ final class SCKAudioCapture: ObservableObject, SystemAudioCaptureEngine, @unchec
         }
 
         self.bufferCallback = bufferCallback
-        errorMessage = nil
+        publishErrorMessage(nil)
 
         // Output handler converts CMSampleBuffer → AVAudioPCMBuffer
         let output = SCKAudioStreamOutput { [weak self] buffer in
@@ -148,7 +148,7 @@ final class SCKAudioCapture: ObservableObject, SystemAudioCaptureEngine, @unchec
         } catch {
             AppLogger.audioSystem.error("SCKAudioCapture: start failed", ["error": error.localizedDescription])
             cleanup()
-            DispatchQueue.main.async { self.errorMessage = error.localizedDescription }
+            publishErrorMessage(error.localizedDescription)
             throw error
         }
     }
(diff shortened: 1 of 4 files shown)


r3dbars:
## Code Review

**Verdict**: REQUEST CHANGES

**Summary**: This PR fixes two real bugs — a threading violation on `@Published errorMessage` in `SystemAudioCapture` and an overly broad path allowlist in `FailedTranscriptionManager` — and adds a solid regression test for the path fix. Both fixes are correct and well-implemented. However, the same threading bug that was fixed in `SystemAudioCapture` also exists in `SCKAudioCapture` and was missed.

---

- **`SCKAudioCapture.swift:116` has the same threading bug this PR fixes in `SystemAudioCapture`.** The `start()` method sets `errorMessage = nil` directly, but `start()` runs on a background thread (confirmed by the semaphore pattern at line 134 and the comment _"synchronous wait since we're on a background thread"_). Meanwhile line 151 already correctly dispatches to main for the error path. This should get the same `publishErrorMessage` treatment (or an inline `DispatchQueue.main.async`) for consistency and correctness.

_None._

### Nits (take or leave)

_None._

### What looks good

- **`publishErrorMessage` helper** is clean: `Thread.isMainThread` avoids redundant dispatch when already on main, `[weak self]` prevents retain cycles. Good pattern.
- **Path allowlist narrowing** is a genuine security improvement. The old path `captures/audio/` was broader than intended — the new `captures/meetings/audio/` correctly scopes to only the transcript-owned archive subdirectory. The existing `isFile(_:containedIn:)` with `resolvingSymlinksInPath()` and trailing-slash normalization handles traversal and prefix-confusion attacks well.
- **The regression test** covers both the positive case (archived audio under `transcripts/audio/` accepted) and the negative case (sibling directory outside transcript root rejected). Combined with the existing traversal, absolute-path, and out-of-home tests, the security boundary is well-covered.
- PR description is clear and verification steps are documented.

---

PR #641, opened 2026-05-02
Fix cleanup helper actor isolation

r3dbars:
## Summary
- mark cleanup URL helper methods as nonisolated so package/dependency builds can call them from nonisolated cleanup paths

## Verification
- bash run-tests.sh
- bash build-deps.sh --force

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
@@ -523,15 +523,15 @@ public class TranscriptionTaskManager: ObservableObject {
         }
     }
 
-    private static func canonicalURL(_ url: URL) -> URL {
+    nonisolated private static func canonicalURL(_ url: URL) -> URL {
         url.standardizedFileURL.resolvingSymlinksInPath()
     }
 
-    private static func canonicalDirectoryURL(_ url: URL) -> URL {
+    nonisolated private static func canonicalDirectoryURL(_ url: URL) -> URL {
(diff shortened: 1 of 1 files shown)


r3dbars:
## Code Review — PR #641

**Verdict**: APPROVE

**Summary**: This PR adds `nonisolated` to three `private static` pure-function helpers on `TranscriptionTaskManager` (`canonicalURL`, `canonicalDirectoryURL`, `isFile(_:containedIn:)`). Because the class is `@MainActor`, these static methods inherited main-actor isolation by default, which prevented the existing `nonisolated` cleanup path (`isSafeCleanupURL` → static helpers) from compiling under strict concurrency without async bridging. The fix is correct, minimal, and safe.

**Blockers**: None.

**Should fix**: None.

**Nits**: None.

**What looks good**:

- **Correct diagnosis**: The three methods are pure functions — they take URL arguments, return URL/Bool, and touch zero instance or class-level mutable state. `nonisolated` is the right annotation here, not `@Sendable` or async bridging.
- **Consistent with existing pattern**: The callers `isSafeCleanupURL` (line 519) and `removeRecordingFile` (line 487) were already `nonisolated`. This change completes the chain so the entire cleanup path is consistently non-isolated.
- **Security-critical path preserved**: The `isSafeCleanupURL` containment check that prevents arbitrary file deletion via `startImportedTranscription` is functionally unchanged — only the concurrency annotation changed, not the logic.
- **Minimal blast radius**: 3 lines changed, 1 file, all `private static` — no public API surface affected.

---

PR #662, opened 2026-05-06
fix: reapply hidden Dock visibility preference

r3dbars:
## Summary
- lets ActivationPolicyController detect when AppKit/updater lifecycle drifts away from the cached desired policy
- reapplies the hidden-Dock/accessory policy on next run loop, Dock preference changes, and app activation
- adds a regression test for the “toggle is off but app is back in the Dock” case

## Test Plan
- [x] bash run-tests.sh
- [x] bash build.sh

User feedback: after updating, Transcripted can show in the Dock even though “Show Transcripted in Dock” remains unchecked.

### DIFF
--- a/Sources/Support/ActivationPolicyController.swift
+++ b/Sources/Support/ActivationPolicyController.swift
@@ -66,7 +66,10 @@ final class ActivationPolicyController {
     }
 
     func setShowInDock(_ visible: Bool) {
-        guard showInDock != visible else { return }
+        guard showInDock != visible else {
+            applyIfDrifted(currentPolicy)
+            return
+        }
         showInDock = visible
         reconcile()
     }
--- a/Sources/Support/ActivationPolicyController.swift
+++ b/Sources/Support/ActivationPolicyController.swift
@@ -41,15 +41,18 @@ final class ActivationPolicyController {
     private var isMeetingRecording: Bool = false
     private var isDictationRecording: Bool = false
 
-    /// Indirection so tests can capture policy changes without touching
-    /// NSApp.
+    /// Indirections so tests can capture policy changes without touching
+    /// NSApp, and so the controller can correct AppKit/updater drift even
+    /// when its cached desired policy has not changed.
+    private let actualPolicy: (@MainActor () -> NSApplication.ActivationPolicy)?
     private let applyPolicy: @MainActor (NSApplication.ActivationPolicy) -> Void
 
     init(
         showInDock: Bool = DockVisibilityPreferences.isVisible(),
         applyPolicy: @escaping @MainActor (NSApplication.ActivationPolicy) -> Void = { policy in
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: REQUEST CHANGES

**Summary**: This PR adds drift detection to `ActivationPolicyController` so that when AppKit, Sparkle, or SwiftUI lifecycle events silently flip the process back to `.regular`, the controller can re-enforce the user's hidden-Dock preference. The approach is sound — an `actualPolicy` closure to read real AppKit state, a `reapplyCurrentPolicy()` repair path, and hooks on launch, preference change, and app activation. However, the production wiring is missing the key ingredient that makes it work.

- **`actualPolicy` is never passed in production — drift detection is completely inert.** `TranscriptedApp.swift:88` constructs `ActivationPolicyController()` with no arguments, so `actualPolicy` defaults to `nil`. In `applyIfDrifted`, the `guard let actualPolicy` fails immediately and returns, making every `reapplyCurrentPolicy()` call a no-op. The three new call sites (launch async, preference change, `didBecomeActive`) all silently do nothing. The test passes because it injects `actualPolicy`, masking the production gap. Fix by either passing `actualPolicy: { NSApp.activationPolicy() }` at the call site, or changing the default from `nil` to `{ NSApp.activationPolicy() }`.

- **Redundant `reapplyCurrentPolicy()` after `setShowInDock` in the preference-change handler.** In `wireActivationPolicy` (line 458–461), `setShowInDock` already calls `reconcile()`, which now calls `applyIfDrifted`. The follow-up `reapplyCurrentPolicy()` is a harmless duplicate, but it suggests the author wasn't sure whether `reconcile` handled drift — worth removing for clarity once the blocker is fixed.

**Nits** (take or leave):

- The `didBecomeActiveNotification` fires on every app activation (popover open, Dock click, Cmd-Tab). Once `actualPolicy` is wired up, this will call `NSApp.activationPolicy()` and compare on every activation. That's fine — it's behind a guard — but worth a mental note that this is a hot path for a menubar app.

**What looks good**:

- The `applyIfDrifted` design is clean — optional closure keeps tests decoupled from `NSApp`, and the guard chain reads well.
- The regression test is well-structured: it simulates the exact drift scenario (AppKit flips to `.regular` behind the controller's back) and validates both the repair action and the cached state.
- Thread safety is correct throughout: `@MainActor` class, all call sites dispatch to main, weak captures in notification sinks prevent retain cycles.
- The doc comment on `reapplyCurrentPolicy()` explains *why* drift happens (Sparkle, SwiftUI scenes, AppKit focus) which is exactly the kind of non-obvious context that belongs in a comment.
- Security review: clean. `setActivationPolicy` only affects Dock/force-quit visibility — no privilege escalation, no injection surface, no memory safety concerns.

---

PR #694, opened 2026-05-08
Reduce idle audio recovery noise

r3dbars:
## Summary
- downgrade idle audio-device recovery timeouts to warning-only diagnostics instead of Sentry errors
- compute meeting/system audio levels across all channels
- add failure_kind to meeting transcription failure diagnostics

## Verification
- swift test --filter AudioDiagnosticsSnapshotTests
- bash build-deps.sh --force
- bash build.sh
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-integration-smoke.sh
- swift test

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -1433,6 +1433,7 @@ final class MeetingSessionController: ObservableObject {
                 context: baseDiagnosticsContext(
                     extra: [
                         "error": message,
+                        "failure_kind": analyticsFailureKind(from: message),
                         "queue_depth": "\(queuedTranscriptionJobs.count)",
                         "trigger": transcriptionTrigger.rawValue
                     ]
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -1184,6 +1184,7 @@ class ParakeetEngine: ObservableObject {
 
             self.configRecoveryTimeoutTask = nil
             self.publishRecoveryState()
+            let failureAction = ParakeetDeviceRecoveryFailurePolicy.action(wasRecording: wasRecording)
             AnalyticsReporter.track(
                 "dictation_audio_route_recovery_timeout",
                 properties: self.dictationRouteAnalyticsContext(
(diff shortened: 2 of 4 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR reduces Sentry noise by downgrading idle audio-device recovery timeouts from errors to warnings, fixes a real bug where multi-channel RMS level computation only read channel 0 (missing the right channel on stereo system audio), and adds a structured `failure_kind` tag to meeting transcription failure diagnostics. Clean, well-scoped, and correctly tested.

---

**Blockers**: None.

**Should fix**: None.

**Nits**:

- `ParakeetDeviceRecoveryFailurePolicy.action(wasRecording:)` currently returns `schedulePrewarmRetry: true` unconditionally, so the `if failureAction.schedulePrewarmRetry` guard in `ParakeetEngine.swift` is a no-op today. This is fine as a forward-looking policy seam, but worth noting it doesn't change runtime behavior yet.

**What looks good**:

- **The multi-channel RMS fix is the real win here.** The old `data.pointee` (channel 0 only) meant stereo system audio from ScreenCaptureKit (48kHz, 2ch, non-interleaved) was measuring only the left channel. The new `normalizedRMSLevel` correctly sums across all channels for both interleaved and non-interleaved layouts, matching how `linearPeak` already worked. Good catch.

- **Deduplication of the RMS calculation.** Two identical 11-line inline computations in `calculateLevel` and `calculateSystemLevel` are replaced by one shared function with better edge-case handling (`sum.isFinite`, `level.isFinite` guards).

- **The policy pattern in `ParakeetDeviceRecoveryFailurePolicy` is consistent** with the existing `ParakeetStartRecordingFailurePolicy` in the same file. Keeps recovery behavior decisions testable and out of the engine's timeout handler.

- **Test coverage is solid.** `testNormalizedSystemLevelUsesAllChannels` creates a stereo buffer with left=0, right=0.5 and asserts level > 0.8. I verified the math: RMS ≈ 0.354, power ≈ -9.03 dB, normalized level ≈ 0.85. This would have been ~0 with the old single-channel code, so the test directly validates the fix.

- **`failure_kind` addition** gives structured failure classification in Sentry/diagnostics context without leaking raw error text — `MeetingFailureKind.classify` maps to a fixed enum, consistent with the privacy-safe diagnostics pattern used elsewhere.

---

PR #730, opened 2026-05-13
[codex] Fix blocked Bluetooth audio recovery

r3dbars:
## Summary

Fixes the Bluetooth/device-change recovery path in `ParakeetEngine` when CoreAudio gets stuck while applying the built-in mic fallback during a route change.

## Root Cause

Local logs showed the failure shape from `APPLE-MACOS-17` / `APPLE-MACOS-1F`: Bluetooth output with a built-in mic fallback, `selection_overrode_default=true`, `format_ready=false`, and device-change recovery timing out.

The recovery timeout tried to rebuild the audio engine by queueing work onto the same audio-engine queue that could already be blocked inside the stale CoreAudio snapshot path. That meant timeout recovery could wait behind the thing it was supposed to escape.

## Changes

- Capture the current `AVAudioEngine` and queue per queued graph operation.
- Add a hard timeout recovery path that swaps to a fresh `AVAudioEngine` and fresh serial queue immediately.
- Retain retired engines briefly so stale CoreAudio callbacks do not hit freed objects.
- Ignore stale recovery callbacks after a newer recovery generation wins.
- Add policy tests for hard timeout recovery and retired-engine lifetime.

## Verification

- `bash build.sh`
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` -> `1757 tests, 1757 passed`
- `bash /Users/redbars/.codex/skills/transcripted-repro-lab/scripts/run_daily_audio_reliability.sh --synthetic` -> `Audio reliability: PASS`, `72 passed`, `0 failed`

## Notes

Live Sentry refresh was blocked locally because `SENTRY_AUTH_TOKEN` was not set, so the issue mapping used prior Sentry memory plus current local Transcripted logs.

### DIFF
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -352,6 +352,28 @@ private struct ParakeetInputDeviceApplication {
     let errorDescription: String?
 }
 
+private final class ParakeetRetiredAudioEngineStore {
+    static let shared = ParakeetRetiredAudioEngineStore()
+
+    private let lock = NSLock()
+    private var engines: [AVAudioEngine] = []
+
+    func retire(_ engine: AVAudioEngine, reason: String) {
+        lock.withLock {
+            engines.append(engine)
+        }
+
+        let delay = ParakeetAudioEngineRetirementPolicy.deferredReleaseDelayNanoseconds
+        DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + .nanoseconds(Int(delay))) { [weak self, engine] in
+            guard let self else { return }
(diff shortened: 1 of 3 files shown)


r3dbars:
PR reviewer pass for 1ddc444493f8c85237df25ed2c3a11d5fd043c30

Verdict: clean

Findings:
- No blocking findings. The hard-reset path avoids queueing timeout recovery behind the potentially stuck audio-engine queue, and the policy tests cover that behavior.

Actions:
- Reviewed PR metadata and diff. No changes made. Did not rerun this branch's tests in this reviewer pass.

PR #731, opened 2026-05-13
Test storage-refactor save-directory and capture-id seams

r3dbars:
## Summary
- Pin `TranscriptSaver.defaultSaveDirectory` behaviour: custom path → `<custom>/meetings`, and relative/traversal/system paths fall back to the default. Catches a future revert of the markdown-first refactor.
- Pin `RecordingValidator.resolvedSaveDirectory` behaviour: disk-space and permission preflight probe `<custom>/meetings` (not the bare custom path) when a capture-library override is set, with matching fallbacks for tampered preferences.
- Pin `TranscriptScanner.parseTranscriptFile` capture-id parsing: round-trip `capture_id` + legacy `transcript_id`, mint a UUID when absent, and truncate adversarial values at 256 chars before they reach the stats database.
- Promote `resolvedSaveDirectory` and `parseTranscriptFile` from `private` to `internal` for `@testable` access, matching the convention used by other Core helpers (`normalizedCustomSavePath`, `validateRawSavePath`, `diskSpaceProbeURL`).

Adds 17 tests across 3 new files in the `TranscriptedCoreTests` package target; no source-behaviour change.

## Test plan
- [x] `swift test` — 153 tests, 0 failures
- [x] `bash build.sh` — app builds cleanly, smoke check + performance budget pass
- [x] `bash run-tests.sh` — 1752 tests, 0 failures

### DIFF
--- a/Sources/TranscriptedCore/Services/RecordingValidator.swift
+++ b/Sources/TranscriptedCore/Services/RecordingValidator.swift
@@ -112,7 +112,7 @@ public enum RecordingValidator {
         }
     }
 
-    private static func resolvedSaveDirectory(paths: CoreStoragePaths) -> URL {
+    static func resolvedSaveDirectory(paths: CoreStoragePaths) -> URL {
         if let customPath = normalizedCustomSavePath(),
            !customPath.isEmpty {
             if validateRawSavePath(customPath) != nil {
--- a/Sources/TranscriptedCore/Storage/TranscriptScanner.swift
+++ b/Sources/TranscriptedCore/Storage/TranscriptScanner.swift
@@ -78,7 +78,7 @@ public enum TranscriptScanner {
 
     /// Parse a transcript file and extract metadata
     /// Returns tuple of (metadata, actionItemsCount) or nil if parsing failed
-    private static func parseTranscriptFile(_ fileURL: URL) -> (RecordingMetadata, Int)? {
+    static func parseTranscriptFile(_ fileURL: URL) -> (RecordingMetadata, Int)? {
         guard let content = try? String(contentsOf: fileURL, encoding: .utf8) else {
             AppLogger.pipeline.warning("TranscriptScanner could not read file", ["file": fileURL.lastPathComponent])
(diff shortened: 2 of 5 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR adds 17 well-structured regression tests pinning the storage-refactor's save-directory resolution, capture-ID parsing, and path-traversal defenses. The only production change is widening two functions from `private` to `internal` for `@testable` access, which is consistent with existing conventions in the same files (`normalizedCustomSavePath`, `validateRawSavePath`, `diskSpaceProbeURL`). Clean and focused.

**Blockers**: None.

**Should fix**:
- `RecordingValidatorResolvedSaveDirectoryTests` creates a UUID-scoped temp directory for `paths` in `setUp()` but never removes it in `tearDown()`. The other two test files properly clean up their temp dirs. Add something like `try? FileManager.default.removeItem(at: paths.transcripts.deletingLastPathComponent())` in `tearDown` to avoid accumulating orphaned directories in CI.

**Nits**:
- `RecordingValidatorResolvedSaveDirectoryTests` and `TranscriptSaverDefaultSaveDirectoryTests` test nearly identical 6-case matrices (no custom, blank, valid custom → /meetings, relative fallback, traversal fallback, system-dir fallback) against two different entry points. This is defensible since they pin two distinct API surfaces that could diverge, but if the duplication bothers you later, a shared parameterized helper could collapse it.

**What looks good**:
- UserDefaults save/restore pattern in setUp/tearDown is correct — original value preserved even on test failure.
- The `writeTranscript(captureIDField:captureID:)` helper is cleanly parameterized and throws for proper propagation.
- Testing the 256-char truncation defense and the `transcript_id ?? capture_id` precedence are high-value pins that will catch accidental regressions in the frontmatter parser.
- Temp directories use class-name + UUID naming, eliminating parallel-execution collisions.
- Visibility widening stays at `internal` (not `public`), so the release module boundary is unchanged.

---

PR #733, opened 2026-05-13
Security audit fixes: HF model integrity, dead beta token, worker proxy route, entitlement docs

r3dbars:
## Summary

Closes four findings from the /cso security audit run on 2026-05-12.

- **HF model integrity** — `ModelDownloadService` no longer falls back to `hf-mirror.com`. The mirror also supplied the manifest SHA-256s used for verification, so a compromised mirror could just hand back a manifest whose `lfs.sha256` matched the malicious file it intended to serve. huggingface.co is now the only source; downloads fail closed instead of silently degrading.
- **Dead beta token** — `BetaConfig.userToken` was no longer consumed by any Swift code but `build-beta.sh` kept baking it into every signed beta binary, where `strings` could recover it. Stripped the constant and the `inject_beta_token` plumbing. `BETA_TOKEN` is retained as an optional positional arg for invocation compat but is no longer substituted into source.
- **Anthropic proxy route in worker archive** — `POST /v1/messages` forwarded to api.anthropic.com using a worker-side `ANTHROPIC_API_KEY`. The app on main no longer uses this route, but any leaked beta token used to be enough to bill operator-funded LLM spend. Removed the route + env field. Docs explicitly call out that the live worker needs to be redeployed and the Anthropic key rotated out of secrets.
- **Entitlement + model integrity docs** — added two sections to `SECURITY.md`. Model integrity section names the huggingface.co-only sourcing and the LFS-vs-non-LFS digest gap (non-LFS files like tokenizer.json are not hash-pinned). Beta entitlement section explains why `app-sandbox`/`disable-library-validation`/`allow-jit` are set, what that means for users, and the practical mitigations.

## Out-of-band action required

The live Cloudflare Worker at `draft-proxy.tz427gsydr.workers.dev` may still serve `POST /v1/messages`. After this lands:

1. Redeploy the worker from `archive/backend-beta-worker/` (or take it offline).
2. Rotate `ANTHROPIC_API_KEY` out of the worker's Cloudflare secrets.
3. Optionally drop the now-dead `api_calls` D1 table.

## Test plan

- [x] `swiftc -parse` clean across `Sources/Beta` + `Sources/TranscriptedCore/Services`
- [x] `bash -n scripts/entrypoints/build-beta.sh` clean
- [x] Worker TS brace-balanced, no orphan `handleMessages` refs, `ANTHROPIC_API_KEY` only present in a deprecation comment
- [ ] `bash build-deps.sh && bash build.sh && bash run-tests.sh && bash run-integration-smoke.sh` — deferred to CI / reviewer (fresh worktree had no prebuilt deps and build-deps is multi-minute)
- [ ] `swift test` — same
- [ ] Smoke beta build: `./build-beta.sh` (no token arg) should succeed end-to-end and the resulting binary should not contain any beta token string when `strings` is run against it

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -43,6 +43,51 @@ Operational caveats:
 - beta builds can optionally contact the update/log proxy for update checks and diagnostics shipping
 - core dictation and transcription do not require cloud APIs
 
+## Model integrity
+
+Model files come from `https://huggingface.co` only. There is no third-party
+mirror fallback: if huggingface.co is unreachable, the download fails closed
+rather than silently degrading to a less-trusted source whose manifest could
+supply attacker-controlled digests.
+
+LFS-tracked files (weights) are verified against the SHA-256 digest reported by
+the HuggingFace API before being moved into the cache. Non-LFS files
+(tokenizer.json, config.json, etc.) ride on HTTPS to huggingface.co with no
+per-file digest — those are documented but not currently hash-pinned.
+
+## Beta build entitlements
+
+The beta build is intentionally less hardened than a typical sandboxed Mac app.
+`config/entitlements/beta.plist` sets:
+
+- `com.apple.security.app-sandbox` = false
+- `com.apple.security.cs.disable-library-validation` = true
+- `com.apple.security.cs.allow-jit` = true
+
+Combined with the TCC permissions Transcripted asks for (microphone, screen
+recording, accessibility, calendars), this means: anything running as your user
+that can write into `Transcripted.app/Contents/Frameworks/` can load its own
+code into the running app and inherit those permissions, including the ability
+to inject keystrokes into other apps via the paste-back path.
+
+The reason library validation is disabled is that the on-device ML stack (MLX,
+WhisperKit, CoreML model bundles) currently ships dylibs that do not all carry
+the same team-id signing as the app, so strict library validation refuses to
+load them. We treat that as a known cost of running ML locally on macOS today.
+
+Practical implications for users:
+
+- install Transcripted from a trusted source (the signed DMG from GitHub
+  Releases) and verify the Developer ID before granting permissions
+- treat user-level malware as already game-over for these permissions, since
+  it could substitute a framework binary and inherit them
+- the local-test entitlements profile (`config/entitlements/local.plist`) is
+  meant for ad-hoc smoke builds only and shares the sandbox-off posture — it
+  is not the shipping profile
+
+If a future version of the ML stack supports team-id-consistent signing for all
+embedded dylibs, the `disable-library-validation` entitlement will be removed.
+
 ## Supported Versions
 
 | Version | Supported |
--- a/Sources/Beta/BetaConfig.swift
+++ b/Sources/Beta/BetaConfig.swift
@@ -1,22 +1,27 @@
 // BetaConfig.swift
-// Per-user beta configuration — only compiled into beta builds.
-// The token placeholder is replaced by build-beta.sh via sed for each user's DMG.
+// Beta-build compile-time configuration. Currently empty.
+//
+// History: this file used to ship a per-user bearer token baked in at build time
+// via sed/perl substitution in build-beta.sh. That token authenticated the app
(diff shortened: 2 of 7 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR closes four findings from a security audit — removing a compromised-mirror fallback in `ModelDownloadService`, stripping a dead beta token that was still baked into signed binaries, removing an Anthropic proxy route from the archived worker, and documenting the entitlement/model-integrity posture in `SECURITY.md`. The changes are well-scoped, the security rationale is clearly documented at every removal site, and the fail-closed approach for model downloads is the right call. Net deletions (-181/+118) are a good sign for a security hardening PR.

---

None.

---

1. **`Sources/Beta/CLAUDE.md` is stale.** Line 9 still says _"`BetaConfig.swift` — `#if BETA_BUILD` token placeholder replaced by `build-beta.sh`"_. The token placeholder is gone — this doc now actively misleads anyone reading it before editing `BetaConfig.swift`.

2. **`build-beta.sh` header comments (lines 2–4) still show the old required-token usage.**
   ```bash
   # Usage: ./build-beta.sh <beta-token> <user-name>
   # Example: ./build-beta.sh transcripted-beta-nate Nate
   ```
   The script now works without a token, so these comments are wrong. Update to reflect that the token arg is optional/ignored.

3. **`build-beta.sh` line 354 still prints the token via `mask_secret`:**
   ```bash
   echo "🔨 Building Transcripted Beta for $USER_NAME (token: $(mask_secret "$BETA_TOKEN"))..."
   ```
   With no token passed, this prints `(token: [redacted])` — confusing for anyone running the build. Since the token is dead, drop the token portion of this line.

4. **Admin endpoint still queries dead `api_calls` table and leaks bearer tokens.** `archive/backend-beta-worker/src/index.ts:141–149` — the `/admin/usage` handler still joins on `api_calls` (which nothing writes to anymore) and includes `u.token` in the JSON response. The join is harmless (LEFT JOIN returns zeros), but exposing raw bearer tokens in the admin response is a pre-existing concern worth cleaning up in a follow-up, especially given this PR's stated goal of reducing the blast radius of leaked tokens.

---

**Nits** (take or leave):

1. **`ModelDownloadService.swift` line 2** — file header still says _"with HuggingFace mirror fallback"_. The mirror fallback is the thing this PR removes.

2. **`mirrors` array naming** — with a single entry that's the primary source (not a mirror), the name `mirrors` is misleading. Consider `endpoints` or just inline the URL where it's used (two call sites).

3. **Error messages still say "mirrors"** — `ModelDownloadService.swift:316` says _"Could not fetch model file list from any mirror"_ and line 405 says _"Failed to download \(filename) from all mirrors"_. These should say "from HuggingFace" or similar now that there's only one source.

4. **Verbose history comment in `BetaConfig.swift`** — the 14-line file header comment (lines 3–17) is thorough context for reviewers, but in a live source file it will become stale fast. Consider moving the historical rationale to the PR description or a commit message and keeping only a one-liner in the file.

---

**What looks good**:

- **Fail-closed model download** — the security reasoning for removing `hf-mirror.com` is exactly right. A mirror that also supplies the verification manifest is not a meaningful integrity check. Good catch, good fix.
- **Token removal is complete** — the `inject_beta_token` function, the `BETA_TOKEN_PLACEHOLDER` string, the cleanup trap, the backup/restore flow — all gone. No dangling references in active code. The backwards-compat shim for the positional arg is a nice touch.
- **Worker cleanup is thorough** — `handleMessages` function removed, route removed, `ANTHROPIC_API_KEY` removed from the `Env` interface, docs updated in both `CLAUDE.md` and `README.md`. The out-of-band action items (redeploy, rotate key, drop table) are clearly called out in the PR description.
- **`SECURITY.md` additions are unusually good** — the entitlements section doesn't just list what's set, it explains why, what the practical attack surface is, and when the entitlement will be removed. The model integrity section honestly documents the non-LFS digest gap. This is the kind of security documentation that actually helps users make informed decisions.
- **Commit structure** — four focused commits, each addressing one audit finding. Easy to review, easy to bisect, easy to revert individually if needed.

---

PR #779, opened 2026-05-16
Release Transcripted 1.1.38

r3dbars:
Deep review follow-up and Transcripted 1.1.38 release metadata.

Changes:
- Hardens failed meeting retry cleanup and retained-audio handling.
- Preserves dictation audio across route recovery without leaking abandoned buffers into later sessions.

- Guards paste-back when focus changes, tightens file permissions/log files, and rejects malformed duration metadata.
- Refreshes build/test freshness checks and per-process fast test runners.
- Bumps app version to 1.1.38, updates Sparkle appcast, and updates the Homebrew cask.

Validation:
- build-deps.sh --force
- build.sh --no-open
- run-tests.sh
- swift test
- Tools/TranscriptedMCP swift test- Tools/TranscriptedCLI swift test

- run-integration-smoke.sh
- notarized/stapled build/Transcripted-1.1.38.dmg

### DIFF
--- a/Casks/transcripted.rb
+++ b/Casks/transcripted.rb
@@ -1,6 +1,6 @@
 cask "transcripted" do
-  version "1.1.37"
-  sha256 "02122f881ab4c914e381cdf0f4834ca287e2b99a44c167fec07d73b31b2432a8"
+  version "1.1.38"
+  sha256 "1d212cf3450567a0e586091346bcbb0f6f0002ce030037b218fa9b160b02f6cf"
 
   url "https://github.com/r3dbars/transcripted/releases/download/v#{version}/Transcripted-#{version}.dmg",
       verified: "github.com/r3dbars/transcripted/"
--- a/Info.plist
+++ b/Info.plist
@@ -11,9 +11,9 @@
 	<key>CFBundleIdentifier</key>
 	<string>com.justinbetker.draft</string>
 	<key>CFBundleVersion</key>
-	<string>1.1.37</string>
+	<string>1.1.38</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.1.37</string>
+	<string>1.1.38</string>
 	<key>TranscriptedSentryDSN</key>
 	<string>https://137ddd7f72199a8d7a06f1644224dce5@o4511202804170752.ingest.us.sentry.io/4511202823045120</string>
 	<key>TranscriptedSentryEnvironment</key>
--- a/Package.swift
+++ b/Package.swift
@@ -2,14 +2,14 @@
 import PackageDescription
 import Foundation
 
-// TranscriptedCore — shared library target co-hosted with Draft.
+// TranscriptedCore — shared library target used by Transcripted's meeting pipeline.
 //
 // Consumed by:
-//   1. Draft's meeting integration via build-deps.sh
+//   1. Transcripted's app build through build-deps.sh
 //   2. `swift test` for the TranscriptedCore smoke tests in this repo
 //
 // Binary dependency layout:
-//   deps-libs/libDraftDeps.a          — prebuilt mega-library (FluidAudio 0.7.9 + MLX + deps + TranscriptedCore)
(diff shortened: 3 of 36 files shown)


r3dbars:
## Code Review — PR #779: Release Transcripted 1.1.38

**Verdict**: APPROVE

**Summary**: This release PR bundles five distinct hardening areas into 1.1.38: dictation audio preservation across route recovery, file permission tightening on all user data paths, focus-change paste guards, failed-meeting retry reliability improvements, and build/test infrastructure freshness. The changes are well-scoped, well-tested, and the implementation quality is high throughout.

---

### Blockers

None.

---

### Should fix

1. **Minor allocation inefficiency in `drainRecordedSamplesForInference`** (`Sources/Speech/ParakeetEngine.swift:2680-2690`): When draining recovered segments in the detached task, `combined` doesn't pre-allocate capacity. For long recordings that span multiple device changes, this means repeated array growth. Consider:
   ```swift
   var combined: [Float] = []
   combined.reserveCapacity(segments.reduce(0) { $0 + Int(Double($1.samples.count) * (TranscriptedConstants.parakeetSampleRate / $1.sampleRate)) })
   ```
   Not urgent — only matters for unusually long dictations with multiple recovery events.

2. **Duplicate help text** in `HomeView.swift` and `TranscriptedSettingsView.swift`: The `retryHelp` computed property now has two consecutive conditions (`!item.hasAudioFiles` and `!item.isRetryable`) that return the exact same string. If these are semantically different failure modes, they could surface different copy to help users understand why retry is disabled. If they're the same condition in practice, one branch is dead code.

---

### Nits

1. In `drainRecordedSamplesForInference` (line 2695-2700), the pattern of `swap(&samples, &sampleBuffer)` followed by `let samplesForResampling = samples` then `samples.removeAll(keepingCapacity: false)` works but reads a bit indirectly. The early-free pattern is valid for memory pressure, though a comment explaining the intent would save future readers a double-take.

2. `DictationPasteTarget.matches(processIdentifier:bundleIdentifier:)` prioritizes PID over bundle ID — correct for runtime identity, but worth a one-line doc note that this is intentional (PIDs are per-launch, bundle IDs survive relaunches, so PID takes precedence for same-session focus checks).

3. The `appcast.xml` entry uses `sparkle:minimumSystemVersion` of `26.0` — just confirming this is intentional for macOS 26 only.

---

### What looks good

- **Audio recovery design**: `RecordedAudioTimeline` as a segment-aware buffer that preserves per-segment sample rates is exactly the right abstraction for cross-device-change recovery. The `drain()` pattern avoids double-buffering, and the `preservingRecordingAcrossRecovery` flag prevents cleanup paths from accidentally discarding in-flight recovery data.

- **Permission hardening strategy**: Moving `restrictFileToOwnerOnly` outside the `if !fileExists` block in `EventReporter` and `ReliabilityPacketRecorder` means pre-existing log files that were created with looser permissions get tightened on next app launch. This is defense-in-depth done right.

- **Focus-change paste guard**: Capturing the paste target at dictation start and checking it before injecting the synthetic Cmd+V is a clean solution to the "user switches apps mid-dictation" race. Falling back to clipboard copy is the safe choice.

- **Failed meeting retry separation**: Distinguishing "retained audio URLs exist for reveal" from "all required audio files exist for retry" (`audioFilesExist()` vs `!audioURLs.isEmpty`) prevents users from hitting retry on partially-retained audio that would immediately fail.

- **`build-deps.sh` freshness checks**: The `deps_are_ready()` function with mtime comparison against source inputs means stale TranscriptedCore artifacts won't silently produce "already built" when the source has changed. Good for CI reliability.

- **Per-process test runner** (`run-tests.sh`): Using `$$` in the generated runner filename with a trap cleanup prevents parallel test invocations from clobbering each other.

- **Delete confirmation for failed meetings with audio**: Adding a confirmation dialog before deleting retained retry audio prevents accidental data loss.

- **Test coverage**: Every behavioral change has corresponding test additions — the `TranscriptionTaskManagerMetadataTests` additions for retry-success-deletes-audio and retry-failure-updates-error are particularly thorough.

---

PR #795, opened 2026-05-18
[codex] Refresh speaker queue after saved meetings

r3dbars:
## Summary
- Refresh the Speakers queue when a saved meeting transcript lands so deferred speaker review badges update immediately.
- Add a focused policy test for concrete save URLs vs nil reset events.

## Verification
- bash build.sh --no-open
- codex-review branch rerun: build, fast tests, integration smoke, swift test all passed

### DIFF
--- a/Sources/UI/Settings/SettingsRecentCaptureRefreshPolicy.swift
+++ b/Sources/UI/Settings/SettingsRecentCaptureRefreshPolicy.swift
@@ -16,6 +16,12 @@ enum SettingsRecentCaptureRefreshPolicy {
     }
 }
 
+enum SettingsSpeakerQueueRefreshPolicy {
+    static func shouldRefreshAfterMeetingTranscriptSave(_ url: URL?) -> Bool {
+        url != nil
+    }
+}
+
 enum SettingsDashboardRefreshPolicy {
     static let passiveRefreshMinimumInterval: TimeInterval = 1.5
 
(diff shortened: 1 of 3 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR adds speaker queue refresh when a meeting transcript is saved, so deferred speaker-review badges update immediately instead of waiting for the next manual navigation. The implementation is clean, minimal, and follows established codebase patterns exactly.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- The policy method `shouldRefreshAfterMeetingTranscriptSave(_ url: URL?) -> Bool` is currently just `url != nil`. If the condition ever grows (e.g., only refresh for meetings with local-speaker splits enabled), having the policy seam already in place will pay off. But if it stays a nil check forever, it's a one-line function behind a 60-character call site — worth being aware of the tradeoff, though the consistency with `SettingsRecentCaptureRefreshPolicy` and `SettingsDashboardRefreshPolicy` in the same file justifies the pattern today.

**What looks good**:
- **Thread safety is solid.** The `.onChange` fires on main, `refresh()` dispatches DB work to a background queue with `[weak self]`, and the `refreshGeneration` counter coalesces rapid-fire calls. No risk of stacking concurrent refreshes.
- **Correct nil filtering.** `lastSavedTranscriptURL` publishes `nil` on reset events (line 1178 of `MeetingSessionController`), and gating speaker refresh on non-nil avoids unnecessary background DB scans on those resets while still letting `refreshRecentCaptures(force: true)` run unconditionally — which is the right behavior for the capture list.
- **Test coverage is focused and useful.** The two assertions (concrete URL → true, nil → false) document the exact behavioral contract and will catch regressions if the policy evolves.
- **Follows established conventions.** Uninhabited enum with static methods, placed in the existing policy file, matching the naming and structure of sibling policies. Easy to find, easy to understand.

---

Issue #803, opened 2026-05-19
Calendar access not requested


javatarz:
**Describe the bug**
Transcripted does not ask me for approval to give calendar permissions making that feature unusable

**Steps to reproduce**
1. Open Transcripted > Settings > Privacy > Permissions > Calendar
2. Confirm app doesn't have permissions for the calendar
3. Click "Allow Calendar Access"
4. Let the System Settings > Privacy & Security > Calendars window show up

**Expected behavior**
Permissions modal shows up asking me to allow Transcripted Calendar access

**Actual behavior**
Confirm Transcripted is not on the list.
No popup is show to allow transcripted to be given calendar permissions and MacOS doesn't have a way to add the app.

**Area**
- Settings

**Environment**
- macOS version: 26.5 (25F71) Tahoe
- Transcripted version/commit: 1.1.38 (though this problem has existed across versions for a month)
- Audio source (mic, Zoom, Meet, etc.): mic (though this has no bearing to the audio source)

**Storage details**
- `~/Library/Application Support/Transcripted/`

**Diagnostics**
Quick grep through the logs in `~/Library/Application Support/Transcripted/logs` yielded nothing. Let me know if I can provide any other information in this space.

r3dbars:
Thanks for the detailed report. I found the root cause: the app was missing the Calendar entitlement, so macOS would not show Transcripted in the Calendar permissions list even though the app was asking correctly.

That fix is merged now and will go out in the next release. Really appreciate the exact repro steps here.

javatarz:
Tested 1.1.42 and the button now prompts the permission check. Thanks for the quick fix and release @r3dbars <3

The settings panel still showed the "Calendar" item with a warning (!) till I restarted the application (even though the OS had given the permission. Looks like there's a missed callback of sorts or a refresh permission check function call missing when the permission is granted. If there isn't a call back, is it worth adding a manual refresh/review permissions button there to make the user experience more seamless?

I'm also OK opening a new ticket for this one, please let me know :)

r3dbars:
You were right: Settings was refreshing its permission snapshot immediately after starting the Calendar permission flow, before the async EventKit request had returned.

I have a follow-up fix open in #813 so the Settings permission rows refresh after the permission request completes. No need to open a separate ticket unless the next build still shows stale state after granting access.

r3dbars:
Following up here: the Settings refresh fix has now merged in #813 too.

So the Calendar permission flow has two fixes on main now:

[...]

PR #811, opened 2026-05-19
[bug-hunter] Refresh settings permissions after prompts

r3dbars:
## Summary
- Refresh Settings permission rows after macOS permission prompts finish.
- Add injectable calendar permission helper tests.

## Signal
- Anchored to Transcripted 1.1.42.
- Sentry: latest-release fatal APPLE-MACOS-S is real, but the stack has no concrete Transcripted source frame, so I did not patch it here.
- PostHog: one latest-release unclean shutdown correlates with active dictation; no checked start-failure or meeting-failure spike.
- GitHub: fresh #803 follow-up says Calendar access is granted in 1.1.42, but the Settings row stays stale until restart.

## Root Cause
Settings refreshed immediately when opening Calendar permission, before the EventKit prompt resolved, and did not listen for permission prompt completion.

## Verification
- bash build-deps.sh
- bash build.sh
- TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh (2210 passed)

Privacy note: this PR does not include user transcript text, audio references, meeting titles, speaker names, emails, tokens, device names, or paths.

PR #813, opened 2026-05-19
Refresh Settings permissions after grants

r3dbars:
## Summary

- Add an awaitable permission action path so callers can refresh after macOS returns from a permission request.
- Update Settings > Privacy permission rows to refresh their snapshot after the request completes instead of immediately before the Calendar grant lands.
- Add a repo contract test for the stale Settings permission refresh path.

## Context

Follow-up to the latest customer comment on #803: 1.1.42 now prompts for Calendar access, but Settings can keep showing the Calendar warning until app restart because the UI refresh happens before the async EventKit request finishes.

## Validation

- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 2208 passed
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash build.sh --no-open`
- `git diff --check`

### DIFF
--- a/Sources/Support/TranscriptedPermissionAccess.swift
+++ b/Sources/Support/TranscriptedPermissionAccess.swift
@@ -64,23 +64,31 @@ enum TranscriptedPermissionAccess {
 
     @MainActor
     static func openSettings(for kind: TranscriptedPermissionKind) {
+        Task { @MainActor in
+            _ = await requestAccessOrOpenSettings(for: kind)
+        }
+    }
+
+    @MainActor
+    @discardableResult
+    static func requestAccessOrOpenSettings(for kind: TranscriptedPermissionKind) async -> Bool {
         switch kind {
         case .microphone:
             switch microphoneAuthorizationStatus() {
             case .authorized:
-                break
+                return true
             case .notDetermined:
-                activateForPermissionPrompt()
-                AVCaptureDevice.requestAccess(for: .audio) { granted in
-                    guard !granted else { return }
-                    Task { @MainActor in
-                        openSystemSettings("x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone")
-                    }
+                let granted = await requestMicrophoneAccessIfNeeded()
+                if !granted {
+                    openSystemSettings("x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone")
                 }
+                return granted
             case .denied, .restricted:
                 openSystemSettings("x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone")
+                return false
             @unknown default:
                 openSystemSettings("x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone")
+                return false
             }
         case .accessibility:
             if !AXIsProcessTrusted() {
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR fixes a race condition where Settings > Privacy permission rows showed stale state after granting permissions — `refreshPermissions()` ran synchronously before the async TCC request completed, so the UI kept showing the warning until restart. The fix extracts an awaitable `requestAccessOrOpenSettings(for:)` method and has Settings `await` it before refreshing. Clean, well-scoped, and correct.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- The `.accessibility` case always opens System Settings and returns `AXIsProcessTrusted()` synchronously, which will almost always be `false` at the time of return (since accessibility requires a manual System Settings toggle, not an async prompt). This means `refreshPermissions()` in Settings won't visually update for accessibility after clicking the row — the user still has to navigate away and back. This matches the old behavior and is inherent to how macOS accessibility works (no awaitable prompt), but it's worth noting that this PR doesn't improve the accessibility path. A follow-up could poll briefly after opening Settings, similar to how `PermissionsOnboardingView` handles it with `startPolling()`.
- The `"x-apple.systempreferences:..."` URL strings appear 7 times across the switch cases. A computed property on `TranscriptedPermissionKind` (e.g., `var systemSettingsURL: String`) could eliminate the repetition. Pre-existing, not introduced here.
- The repo contract tests match exact source strings, which is inherently brittle — but this follows the established pattern in `RepoCommandContractTests.swift`, so no objection.

**What looks good**:
- The old `openSettings(for:)` correctly becomes a backward-compatible shim that delegates to the new async method — still needed by 7 callers in `PermissionsOnboardingView` and `DictationSessionController`.
- The microphone `.notDetermined` case now properly delegates to `requestMicrophoneAccessIfNeeded()`, which already calls `activateForPermissionPrompt()` internally via its default parameter — removing the old redundant direct call and the nested callback-to-Task pattern.
- The `@discardableResult` annotation is the right choice: the fire-and-forget shim discards it, while Settings uses the completion signal.
- The async/await conversion replaces three separate `Task { @MainActor in ... }` blocks inside the switch with a single top-level async function, making the control flow linear and the completion point explicit.
- No security concerns — no new permission escalation paths, no data exposure, no untrusted input handling changes.

---

r3dbars:
verdict: no-findings-but-human

Findings: none from this pass. This overlaps #811 with a different implementation of the same Settings permission-refresh bug, so Justin should choose one path before merge.
Actions: no code changes pushed.
Head SHA: f404cbd6f0b5d517518c9e92e51e735fbcc08bd6
Fix commit: none
Checks: `gh pr view`, `gh pr diff`, `git diff --check origin/main...origin/codex/fix-calendar-permission-refresh`, `Sources/Support/CLAUDE.md`, `Sources/UI/CLAUDE.md`, `Tests/README.md`; GitHub checks reported none.

manual_merge_handoff:
  head_sha: f404cbd6f0b5d517518c9e92e51e735fbcc08bd6
  verdict: no-findings-but-human
  blockers: overlaps #811; permission-prompt UX should get human choice
  checks: gh pr view, gh pr diff, git diff --check, subsystem docs

PR #815, opened 2026-05-20
Add active meeting quit confirmation

r3dbars:
## Summary
- Adds a default-on Quit Safety setting for active meeting recordings.
- Intercepts app quit while a meeting is still recording and offers Keep Recording, Stop & Transcribe, or Save Audio & Quit.
- Adds saved-before-quit recovery copy and focused tests for the new preference/policy path.

## Verification
- `bash run-tests.sh`
- `bash build.sh --no-open`


r3dbars:
**Verdict**: APPROVE with one suggested fix

**Summary**: This PR adds a quit confirmation dialog when the user tries to quit Transcripted during an active meeting recording, offering three clear options: Keep Recording, Stop & Transcribe, or Save Audio & Quit. The feature is default-on, backed by a new `QuitConfirmationPreferences` preference, and integrates cleanly into the existing termination flow, failure taxonomy, and Settings UI. Overall this is well-structured, well-tested, and ready to ship with one defensive fix.

---

- **`MeetingFailureExplanation` retryability/outcome gap for `.savedBeforeQuit`** — The new `.savedBeforeQuit` kind was correctly added to `isSaveStageFailure()` (line 314) and the `captureFailureStage` switch (line 172), but it's missing from the two sibling switch statements in `resolvedRetryability` (lines 226–234) and `resolvedOutcomeKind` (lines 272–276). Both of those list `.saveFailed`, `.speakerNameFinalizationFailed`, and `.speakerFinalizationFailed` explicitly but don't include `.savedBeforeQuit`. If the failed transcription entry ever has `isRetryable: false`, the saved-before-quit entry would resolve as `.permanent` retryability and `.permanentFailure` outcome — directly contradicting the user-facing copy that says "Audio is safe. Finish the transcript from Home when you're ready." In practice `addFailedTranscriptionRetainingAvailableAudio` likely sets `isRetryable: true`, but the four classification methods should stay consistent as a safety net. The fix is adding `.savedBeforeQuit` to both switch cases:

  ```swift
  // resolvedRetryability, around line 231:
  case .saveFailed,
       .speakerNameFinalizationFailed,
       .speakerFinalizationFailed,
       .savedBeforeQuit:
      return .retryableAfterUserAction

  // resolvedOutcomeKind, around line 273:
  case .saveFailed,
       .speakerNameFinalizationFailed,
       .speakerFinalizationFailed,
       .savedBeforeQuit:
      return .recoverableFailure
  ```

---

**Nits** (take or leave):

None. The code follows existing patterns consistently.

---

**What looks good**:

- **Clean three-way decision model** — Keep Recording / Stop & Transcribe / Save Audio & Quit covers every user intent during an active-meeting quit. The safe default (Keep Recording) gets the Return key equivalent; destructive options require a deliberate click.
- **Backwards-compatible classifier** — The `savedBeforeQuit` classify patterns include both the new error messages (`"saved before quit"`) and the old ones (`"quit before this meeting could be transcribed"`, etc.), so pre-existing entries in `failed_transcriptions.json` will now classify as `savedBeforeQuit` instead of falling through to `unexpectedError`.
- **Correct default-on preference** — The nil-key check in `confirmQuitDuringActiveMeetingRecording` returns `true` on fresh installs without writing a default value, which avoids polluting UserDefaults for users who never touch the setting.
- **Safe termination guard change** — Changing the re-entrant `terminationCleanupStarted` guard from `.terminateNow` to `.terminateLater` prevents a second Cmd+Q from cutting off in-flight cleanup. The existing cleanup task's `reply(toApplicationShouldTerminate:)` call on `NSApplication` will satisfy both pending terminate requests.
- **`waitForRecordingFinishBeforeTermination` timeout** — The 20-second deadline with 100ms polling prevents the app from hanging indefinitely during termination if a recording stop gets stuck, while yielding the main actor between polls.
- **Comprehensive test coverage** — Preferences round-trip, stable storage key, policy truth table, failure classification, and user-facing copy are all tested. Tests use injected `UserDefaults` suites with proper cleanup.
- **Documentation** — `Sources/Support/CLAUDE.md` updates are thorough: file list, invariant note ("should default on"), and test index all updated.

---

PR #817, opened 2026-05-20
Fix release gate blockers

r3dbars:
## Summary
- preserve failed meeting audio until user retry/delete or age cleanup
- bundle offline diarizer models in beta/release packaging by default
- keep meeting ASR from blocking dictation UI state
- guard Paste Last Dictation paste targets and reveal consolidated settings deep links
- extend quit preservation wait beyond meeting stop timeout

## Verification
- bash build-deps.sh --force
- swift test --filter FailedTranscriptionManagerTests
- git diff --check && bash -n scripts/entrypoints/build-beta.sh
- bash build.sh --no-open
- bash run-tests.sh
- swift test
- SKIP_NOTARIZATION=1 bash build-beta.sh smoke Justin

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -1050,7 +1050,7 @@ final class MeetingSessionController: ObservableObject {
     }
 
     private func waitForRecordingFinishBeforeTermination() async {
-        let deadline = Date().addingTimeInterval(20)
+        let deadline = Date().addingTimeInterval(TranscriptedConstants.meetingTerminationFinishWaitTimeout)
         while isFinishingRecording && Date() < deadline {
             try? await Task.sleep(nanoseconds: 100_000_000)
         }
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -451,6 +451,7 @@ class ParakeetEngine: ObservableObject {
     private var prefetchedModelPath: URL?
     private var audioWatchdogTask: Task<Void, Never>?
     private var asrInferenceActivity = ParakeetASRInferenceActivityState()
+    private var asrInferenceHandoffCount = 0
     private var asrInferenceWaiters: [CheckedContinuation<Void, Never>] = []
     private var pureSampleTranscriptionActivityCount = 0
     private var asrManagerReady = false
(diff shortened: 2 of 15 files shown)


r3dbars:
## Code Review — PR #817: Fix release gate blockers

**Summary**: This PR bundles five release-blocking fixes into a single commit: preserving failed meeting audio for user retry, bundling offline diarizer models in beta/release builds, decoupling meeting ASR from dictation UI state, adding a focus guard to Paste Last Dictation, expanding consolidated settings deep links, and extending the quit-preservation timeout. The changes are correct, well-tested, and each fix has corresponding contract tests enforcing the invariant. One disk-usage concern worth a follow-up.

---

None.

---

1. **`cleanupOldFailedTranscriptions(olderThanDays:)` has zero call sites** — With `cleanupPermanentFailures()` removed from `init`, there is no automated cleanup path for failed transcription audio. The public method exists (`FailedTranscriptionManager.swift:228`) but nothing invokes it — not on app launch, not on a timer, not on any lifecycle hook. Failed meeting recordings (potentially large WAV files) will accumulate without bound until the user manually deletes them or the files are externally removed. This needs to be wired into a periodic trigger (e.g., app launch, daily timer, or session lifecycle) with a reasonable age threshold.

2. **Shell script duplication in `build-beta.sh`** — The Parakeet model bundling block (lines ~380–401) and the new diarizer model bundling block (lines ~403–425) are structurally identical: same BUNDLE/REQUIRE flag pattern, same skip/copy/error branching, same error messages with different labels. A parameterized shell function like `bundle_model_assets <label> <src> <sentinel_check> <dest> <bundle_flag> <require_flag>` would collapse ~45 lines into two calls plus a ~20-line function, and make adding a third model type trivial.

---

### Nits (take or leave)

- **`presentedPage` staleness** — After a deep link sets `navigationModel.presentedPage = .models`, the value persists even if the user manually navigates to a different page. Currently harmless since nothing reads it outside the `.task(id: presentationID)` block, but resetting it to `selectedPage` at the end of `expandGeneralDisclosureForPresentedPage()` would prevent future accidental stale reads.

- **Paste-last-dictation sequence duplication** — `pasteLastDictationFromSettings()` in `TranscriptedApp.swift:672` and `pasteLastDictationFromMenu()` in `MenuBarPanelController.swift:271` share the identical guard/beep/resolve/capture/activate/paste sequence. A shared helper taking a paster and a pre-paste closure would collapse both sites. Low priority since the contract tests already enforce symmetry.

---

### What looks good

- **`asrInferenceHandoffCount` is well-designed** — It correctly covers the resume-to-begin window between `next.resume()` in `finishASRInference()` and `asrInferenceActivity.begin()` in `beginASRInference()`. Without it, `hasActiveASRWork` would briefly return `false` during the handoff, allowing premature cleanup. The `max(0, ...)` guard is appropriate defensive coding.

- **Privacy improvement** — Replacing `corrected.prefix(80)` / `trimmed.prefix(80)` transcript snippet logging with `chars=\(corrected.count)` in both Parakeet and Whisper engines. Logs should report timing and counts, not user speech content.

- **Termination timeout invariant** — The 35s termination wait exceeding the 30s meeting stop timeout is documented in the constant's doc comment *and* enforced by a unit test (`TranscriptedConstantsTests`). This is exactly how timeout relationships should be maintained.

- **Test coverage is thorough** — Every behavioral change has a corresponding test: contract tests for ASR state decoupling, paste target guard, settings deep links, diarizer bundling, and log sanitization; unit tests for the FailedTranscriptionManager retention policy; constant relationship test for the timeout invariant.

- **DictationPasteTarget focus guard** — Clean pattern: capture the source app's PID and bundle ID before activating, then check it still matches frontmost app before pasting. Falls back to clipboard copy with a clear user-facing message if focus moved. Already well-tested in `ClipboardRestoringTextPasterTests`.

- **Settings deep-link consolidation** — The `presentedPage` / `selectedPage` split cleanly preserves the original page intent through the `consolidatedDestination` mapping, and `expandGeneralDisclosureForPresentedPage()` is a minimal switch that opens the right disclosure group.

---

PR #837, opened 2026-05-21
[codex] Generate meeting playback mix

r3dbars:
## Summary
- generate one derived playback mix for retained live meeting audio before WAV compression
- keep raw microphone and system audio retained for retranscription, retry, and debug flows
- remove the raw Full/System/Mic playback choice path so normal playback does not layer mic bleed over system audio
- preserve issue #500 guardrails by leaving capture, RealtimeAGC, and voice-processing defaults unchanged

## Verification
- bash run-tests.sh
- bash build.sh
- swift test
- codex-review --full-access: clean

### DIFF
--- a/Sources/Meeting/MeetingAudioStorageManager.swift
+++ b/Sources/Meeting/MeetingAudioStorageManager.swift
@@ -12,6 +12,15 @@ protocol MeetingAudioFileValidating {
     func isUsableAudioFile(at url: URL, fileManager: FileManager) -> Bool
 }
 
+protocol MeetingAudioPlaybackMixing {
+    func createPlaybackWAV(
+        microphoneURL: URL,
+        systemURL: URL,
+        destinationURL: URL,
+        fileManager: FileManager
+    ) async throws
+}
+
 struct AVFoundationMeetingAudioConverter: MeetingAudioFileConverting {
     func convertWAVToM4A(sourceURL: URL, destinationURL: URL) async throws {
         let asset = AVURLAsset(url: sourceURL)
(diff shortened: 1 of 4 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE

**Summary**: This PR adds an explicit "Full" playback choice for split-stream meeting audio, letting headphone or clean-room users hear the complete system + microphone mix on demand. The implementation is clean, minimal, well-tested, and correctly scoped — no capture, transcription, or voice-processing behavior is touched.

**Blockers**: None.

**Should fix**: None.

**Nits**:

- The URL-partition pattern (build a `Set` of `standardizedFileURL.path` values, filter remaining URLs) appears in both `splitStreamPlaybackChoices(in:)` and `preferredAudioFiles(in:)`. Not a problem at two call sites in the same struct, but worth noting if a third appears.
- `"full"` is a string literal in the composite choice ID and title, while the other stems (`system_audio`, `microphone`, `playback`, `recording`) have corresponding constants on `MeetingAudioArchiveResolver`. A `compositeStem`/`fullStem` constant would be consistent, though `"full"` is a synthetic concept rather than a filesystem stem, so this is genuinely optional.
- `standardizedFileURL.path` is evaluated ~6 times on the same two URLs across `splitStreamPlaybackChoices` and `compositePlaybackChoice`. Trivially cacheable into two locals, but negligible at typical meeting-audio file counts (2–4 URLs).

**What looks good**:

- **Precedence is correct**: `splitStreamPlaybackChoices` fires after the single-URL early return and before `preferredAudioFiles`, but `MeetingAudioArchiveResolver.attachment(forTranscript:)` already routes `playback.m4a` archives into single-URL attachments, so the split check never hides a pre-mixed playback file.
- **Default stays echo-safe**: System is still first in the choices array, so `defaultPlaybackChoice` and `isCompositePlayback` remain single-source. The "Full" mix is opt-in only.
- **Player handles multi-URL choices correctly**: `MeetingAudioPlayback` already plays multiple `NSSound` instances in sync (same `currentTime`, coordinated pause/resume/seek, `duration = max`), so the composite choice works without any player changes.
- **Test coverage is thorough**: Existing tests for live-meeting and failed-meeting playback are updated to assert the new three-choice ordering, and a dedicated `testPlaybackLoadingPolicyAllowsExplicitFullMix` verifies that the loading policy correctly passes both URLs through only when "Full" is explicitly selected.
- **`compositePlaybackChoice` is well-factored**: Separate method, clear ID format (`full:path|path`), appropriate SF Symbol, and the `[systemURL, microphoneURL]` ordering matches the existing pattern in the rest of the resolver.
- **Remaining URLs are preserved**: Any additional audio files beyond system/mic are appended to the choices array, so the split-stream path doesn't silently drop unexpected files.

---

r3dbars:
Verdict: blocked. Same reviewed head, but GitHub now reports the PR merge state as dirty against main.

Findings: merge conflicts / dirty merge state need resolving before manual merge.
Actions: no fix pushed.
Head SHA: cabdbcb11c117223776ad88d231cfc0710eedf4e
Fix commit: none.
Checks: gh pr view/diff/checks, repo docs, existing marker memory, mergeStateStatus=DIRTY. GitHub checks reported none.

manual_merge_handoff:
  head_sha: cabdbcb11c117223776ad88d231cfc0710eedf4e
  verdict: blocked
  blockers: mergeStateStatus DIRTY against main
  checks: metadata, patch review, merge-state check; no GitHub checks reported

PR #851, opened 2026-05-22
Fix repeated quit termination replies

r3dbars:
## Summary
- reply to every deferred AppKit termination request during shutdown cleanup
- return terminateNow after cleanup has already finished
- add a repo contract test for repeated quit requests

## Verification
- bash build-deps.sh --force
- bash build.sh
- codex-review --mode branch --base origin/main --full-access --parallel-tests "TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh"

### DIFF
--- a/Sources/TranscriptedApp.swift
+++ b/Sources/TranscriptedApp.swift
@@ -77,6 +77,8 @@ class TranscriptedAppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegat
     lazy var meetingPromptDetector = MeetingPromptDetector()
     private var workspaceObservers: [NSObjectProtocol] = []
     private var terminationCleanupStarted = false
+    private var terminationCleanupFinished = false
+    private var pendingTerminationReplyCount = 0
 
     /// Keeps NSApp in sync with the Dock visibility setting and promotes
     /// the app to `.regular` during active recording for force-quit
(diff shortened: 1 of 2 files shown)


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR fixes a real AppKit contract bug where repeated `applicationShouldTerminate` calls during async cleanup would only get a single `reply(toApplicationShouldTerminate:)` back, potentially leaving AppKit hung waiting for replies that never come. The fix correctly tracks the count of deferred requests and replies to all of them once cleanup finishes, plus short-circuits to `.terminateNow` if cleanup already completed.

**Blockers**: None.

**Should fix**:
- In the `guard let self else` path (line 283-285), if `self` were somehow nil, only one reply is sent even though `pendingTerminationReplyCount` may be > 1 by that point. In practice this is unreachable since the app delegate is held by SwiftUI's `@NSApplicationDelegateAdaptor`, but for defensive completeness you could capture the count and reply N times there too. Very minor — not blocking.

**Nits**:
- None. The code is minimal and well-scoped.

**What looks good**:
- The three-state logic (`terminationCleanupStarted` / `terminationCleanupFinished` / count) cleanly separates "in progress", "done", and "how many to reply to" without overcomplicating things.
- Thread safety is correct: `TranscriptedAppDelegate` is `@MainActor`, `applicationShouldTerminate` is called on main by AppKit, and the cleanup Task is `@MainActor`-isolated — so all property accesses are safe without additional synchronization.
- `max(pendingTerminationReplyCount, 1)` in `replyToPendingTerminationRequests` is a nice safety net ensuring at least one reply even if something unexpected happened with the count.
- The test follows the existing repo-contract pattern consistently.

---

PR #854, opened 2026-05-22
Keep mic playback choice for retained meeting audio

r3dbars:
## Summary
- keep retained system and microphone meeting audio as separate playback choices
- keep system audio as the default without layering raw mic+system audio
- add coverage for selecting the mic source so playback only loads that file

## Verification
- bash build-deps.sh --force
- bash build.sh
- codex-review --mode branch --base f3d9b53d --full-access --parallel-tests "TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh"

### DIFF
--- a/Sources/UI/Shared/MeetingAudioArchiveResolver.swift
+++ b/Sources/UI/Shared/MeetingAudioArchiveResolver.swift
@@ -43,10 +43,6 @@ struct MeetingAudioAttachment: Equatable, Sendable {
             return urls.map { playbackChoice(for: $0) }
         }
 
-        if let splitChoice = splitStreamPlaybackChoice(in: urls) {
-            return [splitChoice]
-        }
-
         let preferredURLs = preferredAudioFiles(in: urls)
         if !preferredURLs.isEmpty {
             return preferredURLs.map { playbackChoice(for: $0) }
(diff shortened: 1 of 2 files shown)


r3dbars:
**Verdict**: APPROVE

**Summary**: This PR removes the `splitStreamPlaybackChoice` method that was collapsing retained mic + system audio into a single system-only playback option. Now both sources surface as separate selectable choices ("System" / "Mic") with system remaining the default. The change is minimal, correct, and well-tested.

**Blockers**: None.

**Should fix**: None.

**Nits**: None.

**What looks good**:

- The removal is surgical — deleting `splitStreamPlaybackChoice` and its call site is all that's needed because `preferredAudioFiles(in:)` already orders `system_audio` before `microphone`, so the default stays correct without any new logic.
- The new `testPlaybackLoadingPolicyUsesSelectedSourceOnly` test directly verifies the user-facing contract: selecting Mic loads only `microphone.wav` and doesn't silently fall back to System.
- Existing tests are updated to match the new behavior rather than being deleted, preserving coverage of the invariants that still hold (single-source default, no composite layering, system-first ordering).
- `isCompositePlayback == false` assertions remain, confirming this doesn't accidentally layer both streams during playback.

---

PR #855, opened 2026-05-22
Keep long meetings visible after stop failures

r3dbars:
## Summary
- preserve stop-timeout failed meetings for retry without archiving unfinished scratch audio
- refresh failed-meeting queue entries when late audio finalization completes
- read oversized transcript frontmatter and classify legacy meeting transcripts in previews
- use the longest available meeting audio track for duration metadata

## Verification
- git diff --check
- bash build-deps.sh --force
- bash build.sh
- swift test

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -41,6 +41,7 @@ final class MeetingCaptureBridge: ObservableObject {
     private var cancellables: Set<AnyCancellable> = []
     private let completionAttempt = MeetingCaptureAttempt<CaptureStopResult>()
     private let startAttempt = MeetingCaptureAttempt<Bool>()
+    private var timedOutStopCompletionHandler: ((CaptureStopResult) -> Void)?
 
     init(audio: Audio = Audio()) {
         self.audio = audio
@@ -100,7 +101,9 @@ final class MeetingCaptureBridge: ObservableObject {
     /// expiry (`didTimeOut == true`). On timeout the WAV header may not be
     /// fully patched, so the caller should treat the audio as failed-but-
     /// recoverable rather than enqueuing it for transcription directly.
-    func stopAndAwaitFiles() async -> CaptureStopResult {
+    func stopAndAwaitFiles(
+        onTimedOutCompletion: ((CaptureStopResult) -> Void)? = nil
+    ) async -> CaptureStopResult {
         guard audio.isRecording else {
             return currentStopResult()
         }
@@ -131,6 +134,7 @@ final class MeetingCaptureBridge: ObservableObject {
                         uniquingKeysWith: { _, new in new }
                     )
                 )
+                self.timedOutStopCompletionHandler = onTimedOutCompletion
                 continuation.resume(returning: self.currentStopResult(didTimeOut: true))
             })
         }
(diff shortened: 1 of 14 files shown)


r3dbars:
## Code Review

**Verdict**: APPROVE (with one recommended fix)

**Summary**: This PR solves a real and well-scoped problem — long meetings that time out during stop were getting their audio archived (and potentially deleted from scratch) before WAV finalization completed, making retries point at missing or corrupt files. The fix preserves scratch audio in place for timed-out stops, registers a late-completion handler so the failed queue gets refreshed when finalization does finish, broadens frontmatter parsing so oversized meeting transcripts aren't silently dropped from previews, and uses the longest audio track for duration metadata. The implementation is careful, well-tested, and properly layered across the bridge → controller → task manager → failed manager stack.

---

### Blockers
None.

---

### Should fix

**`updateFailedTranscriptionAudio` should roll back in-memory state on persistence failure**
`addFailedTranscription` (line 139-143 of `FailedTranscriptionManager.swift`) explicitly rolls back the in-memory array when `saveFailedTranscriptions()` fails:
```swift
failedTranscriptions.append(failed)
let didPersist = saveFailedTranscriptions()
if !didPersist {
    failedTranscriptions.removeAll { $0.id == failed.id }
}
```
The new `updateFailedTranscriptionAudio` (line 171-182) mutates the array entry but does **not** restore the original `FailedTranscription` on persistence failure. If disk write fails, in-memory state diverges from disk — the next app launch loads stale audio URLs while the running session sees the new ones. The fix is straightforward: save `existing` before the mutation and restore it on `!didPersist`:
```swift
let existing = failedTranscriptions[index]
failedTranscriptions[index] = FailedTranscription(/* new values */)
let didPersist = saveFailedTranscriptions()
if !didPersist {
    failedTranscriptions[index] = existing
}
```

---

### Nits

1. **`readDocument` re-decodes on every chunk** (`TranscriptFrontmatter.swift:96-108`): Each loop iteration re-creates a `String` from the entire accumulated `Data` and runs `document(in:)` over it. For a 512KB frontmatter that takes ~8 iterations of 64KB chunks, this is fine in practice. But a quick `data.contains` check for the closing `---` byte sequence before attempting the full parse would avoid redundant work on the early chunks that can't possibly contain a complete frontmatter block.

2. **Closure type for `onTimedOutCompletion`**: The parameter is typed as `((CaptureStopResult) -> Void)?` — a plain closure. It's always called from `Task { @MainActor ... }` in `wireCallbacks()`, so it works correctly today. Under strict Swift 6 concurrency, the compiler may want `@MainActor @Sendable` on this type since it crosses isolation boundaries (captured in a `@MainActor` class, stored as a property, invoked later). Not urgent, but worth annotating when the project adopts strict concurrency checking.

3. **`deinit` doesn't nil out `timedOutStopCompletionHandler`**: `MeetingCaptureBridge.deinit` resumes the `completionAttempt` continuation but doesn't clear `timedOutStopCompletionHandler`. Since the object is being deallocated, ARC handles the closure release. But for symmetry with the continuation cleanup — and to avoid a dangling callback if `onRecordingComplete` fires during teardown — explicitly nilling it in `deinit` would be slightly more defensive.

---

### What looks good

- **The `archiveAudio: false` path is well-reasoned.** Skipping the archive copy for timed-out stops is exactly right — the WAV may still be getting its header patched, so copying it prematurely could capture a corrupt snapshot. Keeping scratch in place and letting the late-completion handler update the queue is the correct design.
- **ID threading from controller → task manager → failed manager** is clean. Using a shared `stopTimeoutFailedTaskId` UUID that's passed all the way through ensures the late handler can find and update the right queue entry.
- **`isMeetingTranscript` broadening** is a good change. Checking `capture_type: meeting` in frontmatter values is more reliable than substring-matching section headings, and the fallback to heading checks preserves compatibility with legacy transcripts.
- **Test coverage is thorough.** The new tests cover the stop-timeout retention path, the audio update with metadata preservation, the oversized frontmatter parsing, the legacy heading detection, the longest-track duration, and the Home attention card. The contract tests also verify the structural expectations.
- **The `longestAudioDuration` refactor** correctly handles the case where a short mic placeholder shouldn't drag down the reported meeting duration.
- **Security**: `updateFailedTranscriptionAudio` properly validates URLs through `isSafeAudioURL` before accepting them, matching the existing `addFailedTranscription` pattern.


---

PR #856, opened 2026-05-22
Keep no-speech meetings out of Sentry failures

r3dbars:
## Summary
- classify no-speech, empty-audio, and too-short meeting outcomes as expected skipped transcripts
- keep skipped meeting transcripts out of the Sentry-allowlisted failure path
- mark no-speech and empty-audio runtime outcomes as terminal idle events so they do not become dirty-shutdown noise

## Sentry
- Addresses APPLE-MACOS-1H latest production shape: meeting_transcript_failed with failure_kind=no_speech_detected
- Real model, pipeline, save, and speaker-finalization failures still report through the existing error paths

## Verification
- git diff --check
- bash build-deps.sh --force
- bash build.sh


r3dbars:
## Code Review — PR #856

**Verdict**: APPROVE

**Summary**: This PR widens the existing `recordingTooShort` skip gate in `MeetingSessionController` to also cover `emptyAudio` and `noSpeechDetected` failures, keeping all three expected no-content outcomes out of the Sentry `meeting_transcript_failed` error path. The change is clean, minimal, and well-tested — it addresses a real production noise issue (APPLE-MACOS-1H) without touching any actual error reporting paths.

---

### Blockers
None.

### Should fix
None.

### Nits

- The diagnostic message `"Meeting transcription skipped because the recording had no transcriptable speech"` is slightly misleading for the `recordingTooShort` case, where the issue is duration rather than speech content. This is cosmetic and only appears in local diagnostics, so it's not a blocker — but `"recording produced no usable transcript"` or similar would cover all three cases more accurately.

### What looks good

- **`shouldReportAsSkippedTranscript` on the enum**: Good call putting this as a computed property on `MeetingFailureKind` rather than scattering `==` checks in the controller. Easy to extend if more expected-skip kinds appear later, and it makes the gate in `MeetingSessionController` read naturally.

- **`clearSession` outcome preserves the concrete failure kind**: The old code hardcoded `"recording_too_short"` for all skipped outcomes. Now it passes `failureKind.rawValue`, so the runtime diagnostics marker distinguishes `meeting_empty_audio` from `meeting_no_speech_detected` from `meeting_recording_too_short`. This is the right granularity.

- **Terminal idle events set updated in lockstep**: Both new outcome strings (`meeting_empty_audio`, `meeting_no_speech_detected`) were added to `RuntimeDiagnosticsStore.terminalIdleEvents`, keeping them alphabetically sorted. This ensures dirty-shutdown detection won't flag these as suspicious unclean exits.

- **Thorough test coverage across four files**: Unit tests on the new computed property (positive and negative cases), contract tests asserting the controller wiring, runtime diagnostics loop test covering all three skip outcomes, and Sentry policy tests confirming the event stays out of the allowlist and produces no diagnostic tags. The test expansion in `RuntimeDiagnosticsStoreTests` from a single case to a `for` loop over all terminal skip events is a nice improvement.

- **Existing plumbing already handled**: The analytics allowlist (`AnalyticsEventPolicy`) and reliability packet recorder already had `meeting_transcript_skipped` entries, and `SentryEventPolicy` correctly has no allowlist entry for the skipped event — so the PR only needed to touch the classification and routing, not the downstream consumers.

### Security review
No security concerns. The change only affects internal diagnostics routing — no new data leaves the device, no new Sentry events are created, and the privacy-safe property allowlists are unchanged.

---

PR #899, opened 2026-05-29
Use the source audio's recording date for imported notes (#850)

### LINKED ISSUE #850: Use source audio creation date for imported audio notes
Support/customer feedback from help@transcripted.app. No private audio or transcript content included.

## Request
When using **Settings → General → Audio Files → Transcribe Audio File**, the generated Markdown currently uses the transcription date as the note date.

A customer pointed out that imported audio notes should probably use the input audio file creation date, because that is when the meeting/recording actually happened.

## Why it matters
This keeps imported audio aligned with meeting history, Obsidian/file-based workflows, and chronological note organization.

## Expected behavior
For imported audio files, Transcripted should prefer the source file creation date for the generated meeting/note date when that metadata is available.

## Acceptance
- Imported audio Markdown uses the source file creation date when available.
- If creation date is unavailable or unreliable, Transcripted falls back to the transcription/import date.
- Behavior is covered by a focused test around imported-audio metadata/date selection.
- The app does not mutate the user's source audio file metadata.

r3dbars:
## What & why

Imported-audio notes (Settings → Audio Files → **Transcribe Audio File**) should be dated by **when the recording happened**, not when it was transcribed (issue #850).

Diagnosis surprise: the date *chain* was already wired end to end (`MeetingImportedAudioPreparer.recordingDate` → pipeline → `TranscriptSaver` → Markdown `date:`/`time:` + filename). #875 was a tests-only no-op that referenced `(#850)` without a closing keyword, so the issue never auto-closed. The real defects were in how the preparer *selected* the date — files with stray date tags or copied/downloaded files still got the wrong date.

## Fixes (`MeetingImportedAudioPreparer.swift`)

- **Over-broad matcher** — `isRecordingDateMetadata` matched any key containing the substring `"date"`, so `purchaseDate` / `encodingTime` / release tags qualified. Now an allowlist of genuine creation/recording markers with an explicit denylist (purchase, encode, tagging, release, album, modification).
- **Earliest-wins** — `embeddedRecordingDate` returned `dates.sorted().first`. Now ranks an explicit creation tag above a looser recording tag, ties broken on the earliest date.
- **No "unreliable creation date" handling** — a copied/downloaded/AirDropped file's filesystem `creationDate` is the *copy time*, not the recording time. New `reliableFilesystemDate` drops any timestamp within 120s of import and takes the earliest reliable one (copies only ever push timestamps forward, so the earliest reliable stamp is the best lower bound on the real recording time).
- **Sentinel/future embedded dates** — embedded dates are now bounded both ways: the 1904 QuickTime epoch, the 1970 Unix epoch, and future dates fall through to the file system instead of becoming the note date.

Source audio metadata is still never mutated (unchanged: only the scratch copy is touched).

## Tests (regression prevention)

New focused coverage so this can't silently regress:
- matcher rejects purchase/encode/tagging/release/album/modification dates, accepts creation/recording tags
- creation tags outrank looser recording tags
- filesystem reliability: earliest reliable wins; a copy-time creation date loses to a preserved mtime; all-recent timestamps → no reliable date (fall back to import)
- plausibility guard rejects 1904/1970/future
- end-to-end: a "downloaded" file (copy-time creation, preserved old mtime) uses the preserved recording time
- styler test proving restyle/rename preserves the `date:`/`time:` front matter

## Verification

- `bash build.sh --no-open` ✅
- `bash run-tests.sh` ✅ 2853/2853
- existing `swift test --filter testStartImportedTranscriptionUsesSourceRecordingDateForSavedMetadata` ✅ (date → saved Markdown)

## Known follow-ups (out of scope)

Re-transcribing existing saved audio (`startSavedAudioRetranscription`) resets the note date to now; the legacy `TranscriptSaver.save(text:)` path hardcodes the current date but has no live callers. Neither affects imported audio.

Closes #850

### DIFF
--- a/Sources/Meeting/MeetingImportedAudioPreparer.swift
+++ b/Sources/Meeting/MeetingImportedAudioPreparer.swift
@@ -50,6 +50,16 @@ enum MeetingImportedAudioPreparationError: LocalizedError, Equatable {
 }
 
 enum MeetingImportedAudioPreparer {
+    /// A source timestamp within this window of the import is treated as the copy
+    /// or download act rather than the original recording time (issue #850).
+    static let copyDetectionWindow: TimeInterval = 120
+
+    /// Embedded dates before this are treated as malformed/sentinel values — e.g.
+    /// the 1904 QuickTime epoch or the 1970 Unix epoch written by buggy encoders —
+    /// rather than real recording times (issue #850). 1990-01-01 UTC predates any
+    /// consumer digital recorder that embeds creation metadata.
+    static let earliestPlausibleRecordingDate = Date(timeIntervalSince1970: 631_152_000)
+
     static func prepareImportedAudio(
         from sourceURL: URL,
         scratchDirectory: URL = MeetingStoragePaths.recordingsScratch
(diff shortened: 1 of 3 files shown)


r3dbars (commented):

**Summary**: Clean, well-scoped bug fix. The old `isRecordingDateMetadata` matcher accepted any metadata key containing the substring `"date"`, so stray tags like `purchaseDate` or `encodingTime` could become the note date. The fix replaces that with a proper allowlist/denylist, adds priority ranking so explicit creation tags win over looser recording tags, filters out filesystem timestamps that look like the copy/download act, and guards against sentinel dates from buggy encoders. Test coverage is thorough — every new code path has focused regression tests plus an end-to-end integration test.

**Blockers**: None.

**Security**: No issues. File path handling is safe (standardized URL, UUID scratch names, owner-only permissions). Metadata parsing uses fixed-locale formatters with no injection risk. Plausibility guards prevent sentinel/future dates. Internal visibility broadening is all pure functions on an app target — same trust boundary as private.

**Should fix**: Nothing.

**Nits** (take or leave):

1. **Dual-purpose `copyDetectionWindow`** — This constant is used for two semantically different things: filtering filesystem timestamps that are too close to the import time (in `reliableFilesystemDate`), and tolerating minor forward clock skew on embedded dates (in `isPlausibleEmbeddedDate`). Both happen to want ~120s today, but if the copy detection window ever needs to change (e.g., longer for slow network transfers), the embedded-date future tolerance would change too. A separate `clockSkewTolerance` constant would decouple them, though it's fine as-is if the window stays small.

2. **`isRecordingDateMetadata` and `recordingDatePriority` scan overlapping substrings** — Both methods walk similar marker lists (`"creation"`, `"recorded"`, etc.). You could unify them into a single function returning an `Optional<Int>` priority (nil = not a recording date), which would eliminate the duplicate substring scanning. The current separation is clear and testable though, so this is just an option if you revisit the area.

**What looks good**:

- The allowlist/denylist approach is the right call. The old `contains("date")` was a time bomb for any metadata tag with "date" in it.
- Injecting `now: Date = Date()` for testability without touching callers is clean.
- The "earliest reliable timestamp wins" logic in `reliableFilesystemDate` is a smart insight — copies only push timestamps forward, so the minimum is the best lower bound.
- The priority system for embedded dates (creation > recording > generic) correctly handles files with both a creation tag and a looser recording tag.
- The plausibility guard rejecting dates before 1990 and dates in the future defends against QuickTime 1904 epoch and Unix 1970 epoch sentinels from buggy encoders.
- The styler test proving restyle/rename preserves front matter date is good end-to-end coverage.
- PR description is excellent — clear diagnosis, explains the why behind each change, and documents known follow-ups as out of scope.

---

r3dbars:
verdict: no-findings-but-human

Findings: none.
Actions: reviewed imported-audio date selection, issue #850 acceptance, Meeting/Test docs, fast-test registration, and the added regression coverage; no fix pushed.
Head SHA: 40e5c10097926c31177796b0ed6fb324c09a9e8e
Fix commit: none.
Checks: `gh pr view`, `gh pr diff --patch`, `gh pr checks` (repo-hygiene passed), `gh issue view 850`, `git diff --check origin/main...origin/fix/imported-audio-source-date-850`, local review of `Sources/Meeting/CLAUDE.md` and `Tests/README.md`.

manual_merge_handoff:
  head_sha: 40e5c10097926c31177796b0ed6fb324c09a9e8e
  verdict: no-findings-but-human
  blockers: imported-audio meeting-date behavior needs human signoff
  checks: gh pr view/diff/checks, issue #850 acceptance, git diff --check, repo-hygiene passing, Meeting/Test docs reviewed

PR #904, opened 2026-05-29
Surface stale Claude helper repairs

r3dbars:
## Summary
- show a direct warning when Claude Desktop is using an older Transcripted helper
- make the primary Agent settings action say Update Claude Helper for stale installs
- cover the stale-helper attention message in installer tests

## Verification
- bash build.sh --no-open
- bash run-tests.sh

### DIFF
--- a/Sources/Support/ClaudeDesktopIntegrationInstaller.swift
+++ b/Sources/Support/ClaudeDesktopIntegrationInstaller.swift
@@ -22,6 +22,30 @@ struct ClaudeDesktopIntegrationStatus: Equatable {
     var isInstalled: Bool {
         state == .installed
     }
+
+    var attentionMessage: String? {
+        if !bundledBinaryExists {
+            return "This app build does not include Transcripted direct tools yet."
+        }
+
+        if !configIsReadable {
+            return "Claude Desktop config is not readable JSON. Install will back it up and write a clean config."
+        }
+
+        guard state == .needsRepair else {
+            return nil
+        }
+
+        if !installedBinaryExists {
+            return "Claude Desktop direct tools are missing. Install will copy a fresh helper."
+        }
+
+        if !installedBinaryMatchesBundled {
+            return "Claude Desktop is using an older Transcripted helper. Update now to replace it."
+        }
+
+        return "Claude Desktop points at another Transcripted helper. Repair will update the config."
+    }
 }
 
 struct ClaudeDesktopIntegrationInstallResult: Equatable {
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR adds a computed `attentionMessage` property to `ClaudeDesktopIntegrationStatus` that provides user-facing explanations for various repair states, surfaces that message in the Agent settings page, and relabels the action button to "Update Claude Helper" for stale installs. Clean, well-scoped change with test coverage for the stale-helper case.

**Blockers**: None.

**Should fix**:

1. **Dead code in `ClaudeDesktopStatusRow.detail`** — The `.needsRepair` case in the `switch` at line 570–571 is now unreachable. When `bundledBinaryExists` is true and `configIsReadable` is true, `attentionMessage` always returns non-nil for `.needsRepair` state (it hits `!installedBinaryExists`, `!installedBinaryMatchesBundled`, or the "another Transcripted helper" fallback). So the `if let attentionMessage` check at line 559 always returns before the switch can reach `.needsRepair`. The dead branch still says "server" while `attentionMessage` says "helper" — if someone later removes the `attentionMessage` delegation, they'd silently revert to the old wording. Consider removing the dead `.needsRepair` body or replacing it with the `attentionMessage` fallback text for consistency.

2. **Icon mismatch for `!bundledBinaryExists` when state is `.installed`** — In `agentActionSection`, the attention label uses `claudeDesktopStatusSymbol` and `claudeDesktopStatusTint`. If the bundled binary is missing but the installed one is already configured (dev builds, for example), state is `.installed`, so the label renders as a green checkmark next to "This app build does not include Transcripted direct tools yet." That's visually contradictory. Edge case in practice, but if you're already touching this code, the fix is small — use a dedicated symbol/tint for the attention label or only show it for non-installed states.

**Nits**:

1. The stale-helper condition (`installedBinaryExists && !installedBinaryMatchesBundled`) is checked inline in three places: `attentionMessage`, `claudeDesktopActionTitle`, and `claudeDesktopStatusText`. A small `var isStaleHelper: Bool` computed property on the status struct would DRY that up, but it's fine as-is for three usages.

2. The first two checks in `ClaudeDesktopStatusRow.detail` (`!bundledBinaryExists`, `!configIsReadable`) now duplicate the first two checks in `attentionMessage`. Since `detail` calls `attentionMessage` right after, those pre-checks just skip the property access for the same result. Not wrong, just redundant.

**What looks good**:

- Centralizing the attention logic into a single computed property on the model type is the right call. It keeps the message source of truth out of the view layer.
- The `guard state == .needsRepair else { return nil }` pattern cleanly separates "always show" warnings (missing bundled binary, unreadable config) from repair-specific messages.
- Test covers the specific stale-helper path end-to-end through `currentStatus`, which is the right level to test at.
- The action button relabeling ("Update Claude Helper" instead of "Repair Claude Setup") and status text ("Helper stale") give users a clear, specific call to action instead of generic repair language.

---

PR #929, opened 2026-06-01
Add ghost speaker merge floor

r3dbars:
## Summary
- add a 0.72 minimum similarity floor before ghost speakers can be auto-merged
- apply the floor in both system-audio and mic diarization paths
- add synthetic regression coverage for weak, strong, and no-target ghost speaker cases

## Source truth
Grigory's June 1 support thread reported Transcripted 1.1.44 collapsing a six-person meeting to three speakers because ghost speakers were force-merged at weak cosine similarities.

## Findings
The "Save Names" symptom appears downstream of the merge: the naming sheet only receives already-processed speaker rows. Existing Grigory-style speaker naming regressions still pass, so I did not patch naming finalization separately.

## Verification
- `bash build.sh --no-open`
- `bash run-tests.sh` (3252/3252)
- `swift test` (346 passed, 1 skipped)
- `swift test --filter TranscriptionPipelineHelpersTests` (20/20)
- `swift test --filter SpeakerNamingCoordinatorTests` (21/21)
- deterministic E2E smoke
- Transcripted QA bench deep: 11/12 green; only warning was pre-existing local saved-library/index drift
- Codex review: clean

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
@@ -266,29 +266,33 @@ extension Transcription {
                 // speaker is a ghost we still need a persistent UUID for later transcript
                 // metadata updates, so fall back to creating a best-effort profile.
                 if isGhost {
-                    var bestNonGhostId: Int?
-                    var bestSimilarity: Double = -1
-                    for (otherId, otherMean) in nonGhostMeans {
-                        let sim = Self.cosineSimilarityStatic(meanEmbedding, otherMean)
-                        if sim > bestSimilarity {
-                            bestSimilarity = sim
-                            bestNonGhostId = otherId
-                        }
-                    }
-                    if let targetId = bestNonGhostId {
-                        speakerIdRemap[speakerId] = targetId
+                    let mergeCandidate = Self.bestGhostSpeakerMergeCandidate(
+                        for: meanEmbedding,
+                        nonGhostMeans: nonGhostMeans
+                    )
+                    if let candidate = mergeCandidate, candidate.similarity >= Self.ghostSpeakerMergeSimilarityFloor {
+                        speakerIdRemap[speakerId] = candidate.speakerId
                         AppLogger.transcription.info("Ghost speaker force-merged", [
                             "ghostSpk": "\(speakerId)",
-                            "into": "\(targetId)",
-                            "similarity": String(format: "%.3f", bestSimilarity)
+                            "into": "\(candidate.speakerId)",
+                            "similarity": String(format: "%.3f", candidate.similarity),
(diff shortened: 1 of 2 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR fixes a real user-reported bug where ghost speakers were being force-merged into the closest non-ghost speaker regardless of how weak the cosine similarity was, collapsing a 6-person meeting down to 3 speakers. The fix adds a 0.72 similarity floor, extracts the merge logic into a shared helper (`bestGhostSpeakerMergeCandidate`), applies it consistently in both the system-audio and mic diarization paths, and adds solid regression coverage. Clean, focused change.

**Blockers**: None.

**Should fix**:
- The mic diarization path (lines ~714-720) does not log the ghost speaker merge/standalone decision the way the system-audio path does (lines ~273-295). The system path now logs the threshold, the best candidate's similarity, and the reason for keeping standalone — the mic path silently merges or creates a standalone profile with no log entry. This makes debugging mic-channel ghost speaker behavior harder. Consider adding the same logging to the mic path for parity.

**Nits**:
- `bestGhostSpeakerMergeCandidate` could be simplified from a manual loop to `nonGhostMeans.map { ... }.max(by:)`, which eliminates the force-unwrap on `bestCandidate!.similarity` (line 620). The force-unwrap is safe due to the preceding `nil` check, but the functional version is shorter and avoids the `!` entirely.
- `unitVector(cosineToXAxis:)` is duplicated across at least 3 test files (`TranscriptionPipelineHelpersTests`, `SpeakerMatchingServiceTests`, `EmbeddingClustererTests`). Worth extracting to a shared test helper at some point.
- The 3 new tests exercise the merge floor through the full `processMicChannelWithDiarization` integration path, which is great for regression confidence. A focused unit test of `bestGhostSpeakerMergeCandidate` directly (empty dict → nil, single entry → returns it, picks highest similarity) would also be valuable and faster to run.

**What looks good**:
- The test math is carefully chosen: 0.42 cosine similarity for the weak case (well below 0.72), 0.78 for the strong case (well above), and the all-ghosts edge case where `nonGhostMeans` is empty. All three verified correct.
- Extracting `bestGhostSpeakerMergeCandidate` eliminates the duplicated inline loop and makes the threshold check readable at the call site.
- `GhostSpeakerMergeCandidate` with `Equatable` is a nice improvement over the old `(bestNonGhostId, bestSimilarity)` loose variable pair.
- Enhanced logging in the system-audio path with the threshold value and "below-minimum-similarity" vs "no-non-ghost-speaker-to-merge-into" distinction is genuinely useful for debugging future reports like Grigory's.
- `nonisolated static` is correct for the pure helper — no mutable state, no actor isolation needed.
- Test stubs are properly scoped (`private`, file-level), use `addTeardownBlock` for cleanup, and correctly implement the protocol seams.
- No security concerns: the logging changes only emit speaker IDs (integers) and similarity scores (floats) — no transcript text, audio paths, or PII.

---

PR #945, opened 2026-06-03
Clarify offline speaker post-processing

r3dbars:
## Summary
- Cherry-picks the `8b5d448` follow-up from `codex/grigory-pairwise-overmerge-guard` onto current `origin/main`
- Keeps the patch scoped to offline speaker post-processing and EmbeddingClusterer speaker-test coverage

## Validation
- `git diff --check origin/main..HEAD` passed
- `swift test --filter EmbeddingClustererTests` attempted, but local SwiftPM build is blocked because the checkout is missing required binary dependency folders: `deps-libs`, `deps-frameworks`, and `deps-modules`; compiler error: `no such module 'FluidAudio'`

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
@@ -143,9 +143,9 @@ extension Transcription {
             AppLogger.transcription.info("Running offline diarization on system audio")
             let rawSegments = try await diarization.diarizeOffline(samples: systemSamples, sampleRate: 16000)
 
-            // Post-process diarization segments. Keep PyAnnote/VBx speaker IDs intact
-            // before review: over-segmentation is fixable by merging/name review, but
-            // pre-review pairwise merges can irreversibly collapse real participants.
+            // Post-process diarization segments, but skip the broad pairwise merge
+            // phase for PyAnnote/VBx output. Small-cluster absorption and DB-informed
+            // split still run for noise cleanup and known-speaker corrections.
             let existingProfiles = speakerDB.allSpeakers()
             let speakerSegments = EmbeddingClusterer.postProcess(
                 segments: rawSegments,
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: APPROVE (posted as comment — can't approve own repo PR)

**Summary**: This PR fixes misleading documentation about what `pairwiseMergeThreshold: nil` actually does in `EmbeddingClusterer.postProcess`, and backs the correction with two new tests proving the other post-processing phases (small-cluster absorption, DB-informed split) still run. Clean, well-scoped follow-up to #943.

**Blockers**: None.

**Should fix**: None.

**Nits**:

- `testPostProcessStillRunsDbInformedSplitWhenPairwiseMergeIsSkipped` sits exactly on the `minSegmentsPerProfile * 2` boundary (16 segments, default threshold 8). That's fine since the test uses default params and they're coupled, but if someone ever bumps that default this test would silently stop exercising the split path rather than failing. Not worth changing — just noting.

- Minor style inconsistency: the existing test and the absorption test use `\.speakerId` key path syntax (`Set(processed.map(\.speakerId))`), but the DB-split test uses closure syntax (`Set(processed.map { $0.speakerId })`). Doesn't matter functionally, just a diff in style within the same file.

**What looks good**:

- The old pipeline comment ("Keep PyAnnote/VBx speaker IDs intact") was actively misleading — IDs are *not* kept intact since absorption and DB-split still mutate them. The new wording is precise about what's skipped vs. what still runs.

- The old doc comment "skip pairwise merge entirely" was ambiguous enough to read as "skip all post-processing." The rewrite to "skip only the pairwise merge phase; small-cluster absorption and DB-informed split still run" removes that ambiguity.

- Good test construction — the absorption test picks realistic thresholds (micro-cluster at 4s with 0.95 cosine similarity), and the DB-split test correctly sets up a mixed-speaker cluster with orthogonal embeddings matching two known profiles. Both tests exercise the `pairwiseMergeThreshold: nil` path specifically, which is exactly what the comment clarification claims.

- Renaming `preserved` → `withoutPairwise` and the test function name to `testOfflinePostProcessSkipsTransitivePairwiseCollapse` both improve readability.

**Security**: No concerns. Zero behavioral changes to production code — only comments and test additions. No new inputs, no new data flows, no privacy implications.

---

PR #924, opened 2026-06-01
[docs] Update dictation and live sidecar docs

r3dbars:
## Summary
- Document dictation stop finalization policy, benchmark runner, and mutation lock ownership in Sources/Dictation/CLAUDE.md
- Add LiveMeetingPreviewServer to the meeting subsystem file map

## Verification
- bash scripts/dev/agent-preflight.sh

PR #923, opened 2026-06-01
[nightly-review] Fail live sidecar handoff on cancel

r3dbars:
## Nightly review

Checked current `origin/main`, last-24h PRs, open issues, open nightly PRs, Sentry, and PostHog.

Top risks ranked:

1. Live meeting sidecar final handoff could stay stuck after transcription cancellation. Fixed here.
2. Bluetooth dictation start timeouts still appear in Sentry (`APPLE-MACOS-14`), but open PR #922 is already aimed at stale readiness refreshes, so I did not duplicate that lane.
3. `update_check_finished` still shows `failure_kind=unknown` / `sparkle_1001` on released build `1.1.44`; current `main` already has the no-update classification guard from the prior nightly PR, so this stays watch-only until a newer build emits data.

## Change

- When `cancelActiveTranscription` runs while a live sidecar is waiting for the final Transcripted Markdown, mark that sidecar handoff as failed instead of leaving agents waiting forever.
- Added a `RepoCommandContractTests` guard so this cancel path keeps clearing the pending live sidecar wait.

## Verification

- `git diff --check`
- `scripts/dev/agent-preflight.sh`
- `bash build-deps.sh --force`
- `bash build.sh --no-open`
- `bash run-tests.sh` (`3254 passed`)
- `bash run-integration-smoke.sh`

PR #919, opened 2026-05-31
[bug-hunter] Tighten dictation start timeout recovery

r3dbars:
## Summary

- keep recoverable `parakeet.audio_format_read_timeout` events local-only so Sentry alerts on the final richer `dictation.microphone_start_timeout` when startup actually fails
- prevent a stale readiness-refresh timeout flag from triggering a second post-recovery recording start before fresh readiness refreshes happen
- add focused policy coverage for the stale-timeout/post-recovery path

## Production signal

- Sentry `APPLE-MACOS-1S`, event `fc0e4d70b755483cb00317b1f7c93f81`
- current release `transcripted@1.1.44`
- single `audio_format_read_timeout` during `dictation_start_requested`
- related 7d cluster includes current-release `dictation.microphone_start_timeout` events, which already carry richer start-attempt/recovery context

## Notes

Draft PR #918 touches the broader dictation start recovery tuning path. This PR is intentionally smaller: alert de-noise plus one bounded policy guard.

## Verification

- `bash build.sh --no-open`
- `bash run-tests.sh`

PR #915, opened 2026-05-31
[codex] Support live meeting sidecar and harden audio reliability

r3dbars:
## Summary
- add the live meeting Codex sidecar workspace, preview server, bundled skill, and agent connection copy
- tighten related docs/tests for live meeting handoff, storage paths, and agent setup
- fix Sparkle no-update telemetry and harden audio reliability paths, including APPLE-MACOS-14 dictation start retry limiting

## Root cause / impact
- live meeting handoff needed a durable local workspace and preview loop for agents
- APPLE-MACOS-14 showed repeated dictation mic start attempts on the built-in-mic plus Bluetooth-output route, with no audio samples flowing
- Sparkle no-update was being treated like a failure instead of a successful check

## Verification
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh`
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash build.sh --no-open`

Draft for review because this branch is broad and includes product-facing live sidecar behavior.

PR #963, opened 2026-06-04
Add live sidecar preview server smoke test

r3dbars:
## Summary
- add an executable fast-test smoke for the live meeting preview server token, host, route, and method behavior
- let preview server instances use an injected loopback port while keeping the shared production port unchanged
- compile the preview server into the fast test runner with Network available

## Branch sweep
- sidecar branches #915, #923, #924 are already merged into main
- Parakeet timeout branch #919 is already merged into main
- stale Parakeet/EOU branches were not merged into this continuation branch

## Verification
- scripts/dev/agent-preflight.sh
- bash -n run-tests.sh && bash -n scripts/entrypoints/run-tests.sh
- git diff --check
- bash run-tests.sh — 3575 tests passed
- bash build-deps.sh --force
- bash build.sh --no-open


r3dbars (commented):
**Verdict**: APPROVE (can't formally approve own-repo PRs via API)

**Summary**: This PR adds a comprehensive smoke test for `LiveMeetingPreviewServer` by injecting a port parameter so tests can bind to an ephemeral loopback port instead of the hardcoded production port. The production code change is minimal and backwards-compatible (default parameter preserves existing behavior), and the test covers the important security-relevant paths: token enforcement, host validation, method restriction, and authorized file serving.

**Blockers**: None.

**Should fix**: Nothing.

**Nits** (take or leave):

- The `sendPreviewRequest` multi-line string literal produces `\r\n\r` at the end of headers instead of the HTTP-standard `\r\n\r\n`. This works because the server's custom parser is lenient, but it could trip up someone copying this helper for a stricter server. Adding a trailing newline (e.g. appending `\n` after the multiline literal, or restructuring the string) would make it spec-correct.

- `availableLoopbackPort()` returns 0 on failure, which would cascade into either a `NWEndpoint.Port(rawValue: 0)` rejection or a bind failure — both caught by the existing `assertTrue(false, ...)` early-return. The failure message would be the generic "preview server should start on loopback" rather than "couldn't allocate an ephemeral port". Fine for a smoke test, just slightly less debuggable on obscure failures.

- The `authenticatedPreviewServerURL` instance method falls back to `LiveMeetingCodexSession.authenticatedPreviewServerURL(token:)` if `URLComponents.url` is nil. That static method uses the hardcoded production port, so the fallback would produce an incorrect URL when running with an injected test port. In practice `URLComponents.url` should never be nil for these inputs, so this is purely theoretical — but worth noting that the fallback is technically wrong for the injected-port case.

**What looks good**:

- Clean dependency injection pattern: a single `port: UInt16` parameter with a sensible default preserves `shared` singleton behavior while making the server fully testable.
- Good security coverage: the test hits 401 (missing token), 401 (wrong token), 403 (non-loopback Host header), 405 (mutating method), 204 (favicon bypass), and 200 (authorized routes). That's the entire access-control surface.
- Using raw BSD sockets for the test HTTP client avoids URLSession's caching and redirect behaviors — appropriate for testing exact request/response semantics.
- `availableLoopbackPort()` uses the standard bind-to-port-0 pattern, avoiding hardcoded test ports that might collide.
- The `run-tests.sh` changes correctly add `LiveMeetingPreviewServer.swift` to `APP_SOURCES` and link `-framework Network`.

---

PR #966, opened 2026-06-04
Report first-value progress in CEO brief

r3dbars:
## Summary
- carry first-value PostHog event counts through the nightly digest payload
- show a First-value card in the CEO brief HTML
- keep the existing DAU/workflow event checks and self-test coverage

## Verification
- bash scripts/dev/agent-preflight.sh
- python3 -m py_compile scripts/ops/generate-nightly-digest.py
- python3 scripts/ops/generate-nightly-digest.py --self-test
- python3 scripts/ops/generate-nightly-digest.py --output-dir /tmp/transcripted-digest-check
- bash scripts/ops/health-probe.sh posthog with local ops env

### DIFF
--- a/scripts/ops/generate-nightly-digest.py
+++ b/scripts/ops/generate-nightly-digest.py
@@ -67,6 +67,16 @@ POSTHOG_ACTIVE_EVENTS = (
     "activation_agent_setup_cta_clicked",
     "activation_return_proxy_observed",
 )
+POSTHOG_FIRST_VALUE_EVENTS = (
+    "dictation_completed",
+    "onboarding_first_dictation_saved",
+    "meeting_transcript_saved",
+    "onboarding_agent_cta_clicked",
+    "activation_artifact_action_clicked",
+    "activation_agent_prompt_action_clicked",
+    "activation_agent_setup_cta_clicked",
+    "activation_return_proxy_observed",
+)
 TRUSTED_POSTHOG_HOSTS = {
     "https://app.posthog.com",
     "https://eu.posthog.com",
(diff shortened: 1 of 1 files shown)


r3dbars (commented):
**Verdict**: APPROVE (post-merge review)

**Summary**: Clean, well-scoped change that adds a `POSTHOG_FIRST_VALUE_EVENTS` constant, threads a new `countIf` column through the PostHog HogQL query, carries the count through the digest payload, and renders a "First-value" card in the CEO brief HTML. Self-test coverage is good. Single file, 49 additions — easy to reason about.

**Blockers**: None.

**Should fix**:

1. **DAU semantics shifted slightly.** The `uniq(distinct_id)` in `current_query` now runs over the union of ACTIVE + FIRST_VALUE events. Two events — `onboarding_first_dictation_saved` and `onboarding_agent_cta_clicked` — are new to the WHERE clause. If a user triggers only those onboarding events and nothing from `POSTHOG_ACTIVE_EVENTS`, they'll now count toward DAU. Probably negligible in practice, but if you want DAU to stay pinned to the original active-event definition, use `uniqIf(distinct_id, event IN ({active_event_list}))` instead of `uniq(distinct_id)`.

2. **"First-value" metric is noisy.** 6 of 8 first-value events (`dictation_completed`, `meeting_transcript_saved`, and the 4 activation events) are already in `POSTHOG_ACTIVE_EVENTS`. A power user completing 50 dictations/day inflates `first_value_event_count` identically to a brand-new user's first dictation. The count shows activity volume, not first-time-user progress. If the intent is to track actual first-value *conversions*, consider filtering to `distinct_id`s that triggered the event for the first time (e.g., using `argMin` or a dedicated PostHog cohort). If the intent is just "volume of first-value-relevant activity," the current approach is fine — just be aware the label could mislead.

**Nits**:

- The self-test subset assertion checks only 4 of 8 first-value events. Adding the other 4 (`dictation_completed`, `onboarding_first_dictation_saved`, `meeting_transcript_saved`, `onboarding_agent_cta_clicked`) would guard the full list against drift.
- Minor style: `first_value_event_count` appears 7 times with `isinstance(..., int)` guard checks. A helper or early normalization to `Optional[int]` would reduce repetition, but it follows the existing `event_count` pattern so this is purely a taste call.

**What looks good**:
- `dict.fromkeys(ACTIVE + FIRST_VALUE)` for dedup in the WHERE clause — simple and correct.
- `countIf` approach avoids a second query round-trip. Smart.
- Error paths are consistent — the exception handler returns `{"dau": None, "event_count": None, "error": ...}` and downstream `.get("first_value_event_count")` gracefully yields `None`, handled by the `isinstance` guards.
- HTML output uses `escape()` on all interpolated values. No XSS surface.
- Self-test covers the happy path end-to-end with concrete assertions on both the returned dict fields and the proxy string.

Solid change overall.

---

PR #970, opened 2026-06-04
Avoid recursive prefetch content scan

r3dbars:
## Summary
- avoid recursively walking nested capture archives when checking existing-install model prefetch signals
- keep direct saved meeting/dictation files counting as existing content
- add regression coverage for nested retained-audio archive files

## Verification
- bash build-deps.sh --force
- bash build.sh --no-open
- bash run-tests.sh
- REPETITIONS=5 scripts/dev/benchmark-home-recent-captures.sh
- ruby scripts/ops/dictation-recovery-autoeval.rb --details
- bash run-daily-audio-reliability.sh --synthetic --skip-build --no-launch

## Perf notes
- Home/recent-captures synthetic 10k load: avg 212.3 ms, best 195.9 ms, cancel 0.0 ms
- Dictation recovery autoeval: kept current policy p95 1990 ms vs 2440 ms baseline, 0 unexpected failures
- Synthetic audio reliability: PASS, 72 checks passed, 0 failed

### DIFF
--- a/Sources/Support/ExistingInstallModelPrefetchPolicy.swift
+++ b/Sources/Support/ExistingInstallModelPrefetchPolicy.swift
@@ -61,27 +61,30 @@ enum ExistingInstallModelPrefetchPolicy {
     ) -> Bool {
         ["dictations", "meetings"].contains { subdirectory in
             let directory = captureLibraryURL.appendingPathComponent(subdirectory, isDirectory: true)
-            return directoryContainsRegularFile(directory, fileManager: fileManager)
+            return directoryContainsDirectRegularFile(directory, fileManager: fileManager)
         }
     }
 
-    private static func directoryContainsRegularFile(
+    private static func directoryContainsDirectRegularFile(
         _ directory: URL,
         fileManager: FileManager
     ) -> Bool {
         var isDirectory: ObjCBool = false
         guard fileManager.fileExists(atPath: directory.path, isDirectory: &isDirectory),
-              isDirectory.boolValue,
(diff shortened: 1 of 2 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: Switches the existing-install prefetch content scan from a recursive `FileManager.enumerator` walk to a shallow `contentsOfDirectory` call, so nested audio archive directories under `meetings/` and `dictations/` don't get traversed during startup. Clean fix with good regression coverage.

**Blockers**: None.

**Should fix**: None.

**Nits**:
- The `contentsOfDirectory` call already prefetches `isRegularFileKey` via `includingPropertiesForKeys`, so the subsequent `resourceValues(forKeys:)` call on line 88 should be a cache hit — but you could simplify slightly by checking if the URL is not a directory instead (since `contentsOfDirectory` already excludes hidden files). Not worth changing though, the explicit `isRegularFile` check is clearer about intent.

**What looks good**:
- Method rename from `directoryContainsRegularFile` → `directoryContainsDirectRegularFile` clearly communicates the behavioral change.
- Test covers both the negative case (deeply nested archive file ≠ existing content) and the positive case (direct `.md` file = existing content), in the same test run with incremental filesystem mutations. Solid regression test.
- The storage layout confirms meeting `.md` files live directly in `meetings/` while audio archives nest deeper — so the shallow scan is correct for the actual on-disk shape.
- Good perf numbers in the PR description showing this doesn't regress the home/recent-captures or dictation recovery paths.

---

PR #982, opened 2026-06-05
Harden AirPods dictation route settling

r3dbars:
## Summary
- avoid probing the current AirPods default input before applying the built-in mic override
- require Bluetooth output speech-rate routes to settle before forced built-in fallback is treated as ready
- keep dictation resampling pinned to the tap buffer rate so 48k AirPods tap buffers are not mislabeled as 24k hardware audio

## Verification
- bash build-deps.sh --force
- bash build.sh --no-open
- bash run-tests.sh (3750 passed)
- bash run-daily-audio-reliability.sh --synthetic
- manual AirPods dictation on the patched build: completed/pasted cleanly with built-in mic + Bluetooth output

## Notes
This targets the annoying AirPods headset-mode dictation route without re-enabling VPIO or changing global/default audio devices.

### DIFF
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -1347,17 +1347,10 @@ class ParakeetEngine: ObservableObject {
             "audio_input_selection_load_ms": Self.elapsedMilliseconds(since: selectionStartedAt)
         ]
         if let selection, selection.didOverrideDefault {
-            let deviceCheckStartedAt = CFAbsoluteTimeGetCurrent()
-            let shouldApplyOverride = try await runTimedAudioEngineWork(operation: "\(operation)_device_check") { engine in
-                engine.inputNode.auAudioUnit.deviceID != selection.selectedInput.id
-            }
-            stageTimings["audio_input_device_check_ms"] = Self.elapsedMilliseconds(since: deviceCheckStartedAt)
-            if let recoveryGeneration, recoveryState.isStale(generation: recoveryGeneration) {
-                throw CancellationError()
-            }
-            if shouldApplyOverride {
-                ignoreInputSelectionConfigChangesUntil = CFAbsoluteTimeGetCurrent() + 1.0
-            }
+            // Avoid touching the current default input before the override is applied.
+            // On AirPods routes, even a short read of the default input can briefly
+            // pull playback toward headset-mode audio.
+            ignoreInputSelectionConfigChangesUntil = CFAbsoluteTimeGetCurrent() + 1.0
         }
 
         if let recoveryGeneration, recoveryState.isStale(generation: recoveryGeneration) {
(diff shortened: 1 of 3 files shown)


r3dbars (commented):
**Verdict**: APPROVE (post-merge review)

**Summary**: This PR fixes AirPods dictation by removing a pre-override device probe that could briefly trigger CoreAudio HFP route negotiation, requiring Bluetooth output routes to fully settle before treating forced built-in fallback as ready, and pinning dictation resampling to the tap buffer rate (not the misleading 24kHz hardware rate). The core audio insight is sound and tests are thorough.

**Blockers**: None.

1. **Dead code in `ParakeetEngine.swift:1400`** — `ParakeetInputOverrideSettlePolicy.delayNanoseconds` now returns `TranscriptedConstants.audioRecoveryDelay` (300ms) for *every* case (`.ready`, `.invalid`, `.routeNotSettled`). That means this early-return can never fire:
   ```swift
   if overrideSettleDelay == 0 {
       return snapshot
   }
   ```
   The switch statement with three identical branches also makes the `readiness` parameter meaningless. Either collapse it to a constant (no parameter needed) or, if you expect to differentiate cases later, add a comment explaining the intent.

2. **`selectionReason` is silently ignored** — `ParakeetAudioFormatReadinessPolicy.readiness()` now underscore-prefixes `selectionReason`, but `ParakeetEngine.swift:1208` still passes `selection?.reason`. If this parameter is permanently dead, removing it from the API is cleaner than leaving a no-op argument that callers still compute. If it's expected to matter again, a brief TODO would prevent someone from cleaning it up prematurely.

3. **`hardwareSampleRate` unused in `ParakeetTapSampleRatePolicy`** — Same pattern: underscore-prefixed parameter, tests exercise it, but the implementation ignores it entirely. Looks intentionally scaffolded for a future change where the hardware rate informs the decision. A one-line comment would make that intent explicit.

**Nits**:

- The comment block in `ParakeetEngine.swift:1350-1352` explaining *why* the default input probe was removed is good — exactly the kind of comment that earns its keep.

**What looks good**:

- The core fix is correct: reading `engine.inputNode.auAudioUnit.deviceID` before applying the override can trigger CoreAudio route renegotiation on Bluetooth devices. Removing the probe and unconditionally setting the ignore window is the right call.
- Behavioral shift from "skip settle delay when already ready" to "always settle after forcing AirPods input away" is conservative and appropriate — 300ms is cheap insurance against route flapping.
- Tests are well-updated: existing assertions flipped to match the new behavior (`.ready` → `.routeNotSettled`), test names updated to describe the new semantics, and two new `ParakeetTapSampleRatePolicy` suites cover both the happy path (48kHz tap over 24kHz hardware) and the fallback path (invalid tap rate).
- Test messages are excellent — they describe the real-world AirPods HFP scenario rather than just restating the assertion.

---

PR #984, opened 2026-06-05
Fix Parakeet zombie recovery reset

r3dbars:
## Summary
- fix the Parakeet zombie watchdog reset so it marks recording idle before touching the CoreAudio graph
- clear the device-change restart flag and briefly suppress self-induced config notifications during zombie reset
- add focused fast coverage guarding the zombie reset ordering

## Sentry
- APPLE-MACOS-1T: parakeet.zombie_engine_recovery_failed
- APPLE-MACOS-1V: parakeet.recording_interrupted with reason=recording_restart_budget_exhausted
- route shape: built_in_input_to_bluetooth_output, sample_flow_started=false

## Verification
- bash run-tests.sh -> 3754 passed
- bash build-deps.sh --force
- bash build.sh --no-open

## Patch verdict
Fast-patch-worthy for current 1.1.46 users. Ship through a Sparkle-visible patch build rather than waiting for broader launch work.

### DIFF
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -2293,12 +2293,14 @@ class ParakeetEngine: ObservableObject {
             self.pendingSamplesLock.withLock {
                 self.pendingSamples.removeAll(keepingCapacity: true)
             }
+            self.isRecording = false
+            self.audioLevel = 0
+            self.configChangeWasRecording = false
+            self.ignoreInputSelectionConfigChangesUntil = CFAbsoluteTimeGetCurrent() + 1.0
             await self.eouManager?.reset()
             await self.removeRecordingTap()
             await self.stopAudioEngine()
             self.isEnginePrewarmed = false
-            self.isRecording = false
-            self.audioLevel = 0
 
             try? await Task.sleep(nanoseconds: TranscriptedConstants.audioRecoveryDelay)
             guard !Task.isCancelled else { return }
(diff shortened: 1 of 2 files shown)


r3dbars (commented):

**Summary**: This fixes a race where the zombie watchdog's graph teardown (`removeRecordingTap` / `stopAudioEngine`) fires CoreAudio config change notifications while `isRecording` is still true, causing `handleAudioConfigChange` to schedule spurious device recovery that burns through the restart budget before the watchdog's own retry fires. Clean, minimal fix with a source-level ordering test to guard against regression.

**Blockers**: None.

**Should fix**: Nothing.

**Nits**:
- The zombie watchdog ordering test lives in `ParakeetStartRecordingFailurePolicyTests.swift`, which is about recording *startup* failures. The zombie recovery is a different concern (post-start silence detection). A dedicated test file or at least a comment at the `runSuite` call site noting it's a zombie recovery invariant (not a startup failure policy) would help the next person scanning the manifest. Totally fine to leave as-is though — the test messages themselves are clear about what they guard.

**What looks good**:
- The reordering logic is exactly right: `isRecording = false` before any graph work prevents `handleAudioConfigChange` (line 823–833) from arming `configChangeWasRecording`. Clearing `configChangeWasRecording` explicitly is belt-and-suspenders. The 1-second `ignoreInputSelectionConfigChangesUntil` suppression matches the established pattern for AirPods routes (line 1353) and fully closes the window.
- `removeRecordingTap` and `stopAudioEngine` don't check `isRecording` themselves (they check `inputTapInstalled` and `audioEngine.isRunning` respectively), so moving the reset earlier has no side effects on teardown.
- `isEnginePrewarmed = false` correctly stays *after* `stopAudioEngine()` — it reflects engine state, not the race prevention concern.
- The test is a reasonable structural guard. It's brittle by nature (literal string matching on source), but this pattern is already established in the codebase (`DictationAudioRecoveryTests` does the same), and the failure messages are clear enough that a future developer would immediately understand the invariant if they accidentally reorder lines.
- Good Sentry context in the PR description linking the specific error events and route shape. Makes it easy to verify the fix matches the observed crash path.

---

PR #991, opened 2026-06-06
Fix speaker review duplicate meeting rows

r3dbars:
## Summary
- scope Home speaker-review actions to the exact transcript row instead of global pending review state
- make saved-audio retranscription replace the original transcript and preserve the original recording date/time
- avoid duplicate retained-audio copies when retranscribing an existing saved meeting

## Repro
Synthetic saved-meeting fixture with a reviewed transcript plus retained mic/system audio. Retranscribing the saved audio now rewrites the original Markdown row, keeps the original timestamp, and does not mint a second row.

## Verification
- bash build-deps.sh --force
- bash build.sh --no-open
- bash run-tests.sh
- swift test
- bash scripts/dev/agent-preflight.sh

Privacy: fixtures are synthetic only; no private transcript/audio content used.

### DIFF
--- a/Sources/Meeting/LocalMeetingSummarizer.swift
+++ b/Sources/Meeting/LocalMeetingSummarizer.swift
@@ -195,6 +195,22 @@ enum LocalMeetingSummaryStore {
     static func summaryExists(for transcriptURL: URL, fileManager: FileManager = .default) -> Bool {
         fileManager.fileExists(atPath: summaryURL(for: transcriptURL).path)
     }
+
+    @discardableResult
+    static func removeGeneratedSummary(
+        for transcriptURL: URL,
+        fileManager: FileManager = .default
+    ) throws -> Bool {
+        let url = summaryURL(for: transcriptURL)
+        guard fileManager.fileExists(atPath: url.path),
+              let values = try TranscriptFrontmatter.readValues(from: url),
+              values["capture_type"] == "meeting_summary",
+              values["source_transcript"] == transcriptURL.lastPathComponent else {
+            return false
+        }
+        try fileManager.removeItem(at: url)
+        return true
+    }
 }
 
 enum LocalMeetingTranscriptExtractor {
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -1217,7 +1217,9 @@ final class MeetingSessionController: ObservableObject {
     func retranscribeSavedMeeting(
         micAudioURL: URL?,
         systemAudioURL: URL,
-        title: String?
+        title: String?,
+        transcriptURL: URL? = nil,
+        recordingDate: Date? = nil
     ) async -> Bool {
         guard !(sttRouter.isRecording || sttRouter.isTranscribing) else {
             state = .error("Wait for the current dictation to finish before re-transcribing saved audio.")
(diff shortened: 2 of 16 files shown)


r3dbars (commented):
**Verdict**: REQUEST CHANGES

**Summary**: This PR does three things well: scopes the Home speaker-review button to the specific transcript instead of using global pending-review state, makes saved-audio retranscription overwrite the original Markdown instead of creating a duplicate row, and skips redundant audio archiving when the retained audio already exists. The intent is clear, the test coverage is solid, and the UI change is a clean improvement. However, there is one data-loss scenario in the pipeline cancellation path that needs to be addressed before merge.

## Blockers (must fix before merge)

**Transcript deletion on cancellation during replacement retranscription** (`TranscriptionPipelineRunner.swift:664`)

When `targetTranscriptURL` is set, `TranscriptSaver.saveTranscript()` overwrites the user's original transcript file in-place. The returned `savedURL` points to that same original file. If the pipeline is cancelled after Phase 2 (transcript already overwritten) but before full completion — e.g., during Phase 3 speaker clip extraction — `cleanupCancelledTranscriptSideEffects` calls:

```swift
try? FileManager.default.removeItem(at: savedURL)
```

This deletes the user's original transcript file. Both the old content (already overwritten) and the new content (in the file that just got deleted) are lost. For fresh transcriptions this cleanup is correct — it removes a file the pipeline just created. But for replacement retranscriptions, the file predates the pipeline.

**Suggested fix**: Thread the `targetTranscriptURL` (or a `isReplacement: Bool` flag) into the cleanup path and skip the `removeItem(at: savedURL)` call when the transcript was a replacement overwrite. Alternatively, back up the original content before overwriting and restore it on cancellation, though that's more complex than skipping the delete.

## Should fix (strongly recommended)

None — the rest of the change is well-structured.

## Nits (take or leave)

1. **Unnecessary alias in `TranscriptSaver.swift:199`**: `let explicitFileURL = targetURL` just renames the parameter. Using `targetURL` directly would be slightly cleaner since the name is already descriptive.

2. **`archiveRecordingAudio` could be derived**: In `TranscriptionTaskManager.swift:402`, `archiveRecordingAudio: replacementTranscriptURL == nil` shows the flag is fully determined by whether a replacement URL is provided. The pipeline runner could derive this internally from `targetTranscriptURL != nil` rather than threading a separate boolean parameter, which would shrink the API surface by one parameter.

## What looks good

- **Scoping the review button per-transcript** is the right fix. The old `hasSpeakerReviewWork` computed property was a global check that leaked review state across all meeting rows — if *any* meeting had pending review, *every* meeting with deferred speakers showed the button. The new `hasPendingReview(forTranscript:)` approach with the `hasLoadedProfiles` guard fallback is well-designed: it avoids false negatives during loading by falling back to the per-row `speakerStatus`, then tightens to the precise per-transcript check once profiles are available.

- **`archiveRecordingAudio: replacementTranscriptURL == nil`** is a smart guard. Without it, retranscribing from already-archived audio would create duplicate copies in the same `_audio/` directory.

- **Test coverage** is thorough. The three new tests cover: scoped review action policy, review queue lifecycle after naming, and the full replacement retranscription pipeline including file count and date preservation assertions. The `TranscriptionTaskManagerMetadataTests` test in particular exercises the exact scenario this PR is designed to fix.

- **`recordingDate: item.startDate ?? item.date`** properly preserves the original meeting timestamp through retranscription rather than stamping it with the current time.

## Security review

No actionable security issues. The `targetURL` parameter in `TranscriptSaver` is not externally controllable — it flows from the app's own filesystem scan of the meetings directory through typed Swift parameters. Writes are atomic (`atomically: true` → POSIX `rename(2)`) and permission-hardened (`restrictToOwnerOnly`). The archive-skip path is consistent with the existing retained-audio layout. No TOCTOU gap since the explicit file path skips the `fileExists` loop entirely.

---

PR #1037, opened 2026-06-08
Fix local Gemma summary setup stalls

r3dbars:
## Summary
- prevent Home from starting a local AI meeting summary while Gemma setup is already running
- replace the disabled 'Preparing Gemma...' button with a cancelable setup action
- clarify Beta/Home copy so 'uv installed' is not presented as the model being fully ready
- add coverage for the local Gemma setup single-flight block

## Diagnosis
Justin's M1 feedback screenshots showed Beta stuck on 'Preparing Gemma...' while Home simultaneously showed a meeting stuck on 'Generating AI summary...'. That meant the app could run setup/warmup and an actual summary at the same time, which is especially rough on M1/16GB machines and makes the UI feel frozen.

## Verification
- PASS: bash build.sh --no-open
- RED unrelated: bash run-tests.sh completed 4496 tests with 4494 passing; the only failures were existing release-contract checks in NightlySecurityContractTests.swift:136 and :138 about release fixture / cask vs GitHub asset digest drift, outside this change.

### DIFF
--- a/Sources/UI/Settings/HomeMeetingSummaryBetaPresentationPolicy.swift
+++ b/Sources/UI/Settings/HomeMeetingSummaryBetaPresentationPolicy.swift
@@ -33,3 +33,27 @@ enum HomeMeetingSummaryBetaPresentationPolicy {
         isEnabled
     }
 }
+
+struct HomeLocalSummaryNotice: Identifiable, Equatable {
+    let id = UUID()
+    let transcriptURL: URL
+    let chunkCount: Int
+
+    var title: String { "AI summary saved" }
+    var status: String { "Saved" }
+    var detail: String {
+        let passText = chunkCount == 1 ? "one local Gemma pass" : "\(chunkCount) local Gemma passes"
+        return "The meeting Markdown was enhanced with a generated title and summary preview using \(passText)."
+    }
+}
+
+enum HomeLocalSummaryNoticeDismissalPolicy {
+    static let autoDismissDelayNanoseconds: UInt64 = 6_000_000_000
+
+    static func shouldDismiss(
+        current: HomeLocalSummaryNotice?,
+        scheduledNoticeID: UUID
+    ) -> Bool {
+        current?.id == scheduledNoticeID
+    }
+}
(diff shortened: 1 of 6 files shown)


r3dbars (commented):

**Summary**: This PR fixes a real user-reported problem — M1/16GB Macs stalling when Gemma setup and a summary job run concurrently. The fix threads `isLocalSummaryModelPreparing` into `LocalMeetingSummaryAvailabilityPolicy.unavailableReason()` so Home blocks summary generation while setup is in flight, replaces the disabled "Preparing Gemma..." button with a cancelable "Cancel setup" action, and tightens the Beta copy so "uv installed" isn't misread as "model ready." Clean, focused, well-tested.

**Blockers**: None.

**Should fix**: Nothing.

**Nits** (take or leave):

- The cancel button sets `localSummaryModelPreparationStatus` inline at the call site rather than inside `cancelLocalSummaryModelPreparation()`. This works correctly (the token guard in the continuation prevents races), but it means the cancel function and the button action share responsibility for the post-cancel state. If cancel is ever called from a third site, that site would need to remember to set its own status message. Not a problem today since there are only two callers (button and toggle-off), but worth noting.

- The new `isPreparingLocalSummaryModel` parameter uses a default value (`= false`) in the policy function, which keeps all existing callers compiling. Just flag that the test for the `isPreparingLocalSummaryModel: false` case (the "allows idle saved meetings" suite at line 392) doesn't explicitly pass the new parameter — it relies on the default. That's fine for now, but if someone later removes the default it'll still compile and silently not test the parameter. Very minor.

**What looks good**:

- The token-based guard pattern in the preparation task continuation is solid. It correctly prevents the `CancellationError` / generic error catch blocks from overwriting the user-initiated cancel message, even under `@MainActor` reentrancy.
- Good use of the existing `SettingsInlineActionButton` tone/symbol system — the warning-toned cancel button with `xmark.circle` gives clear visual feedback that this is a destructive-ish action.
- The `unavailableReason` approach is the right call. Threading the single-flight block through the shared policy means both menu item `isEnabled` and the user-facing reason string stay consistent automatically, instead of duplicating checks.
- Test coverage for the new policy gate is clean and well-described.
- Copy improvements ("Runtime ready" instead of "Ready for first summary", the explicit "Use Prepare Gemma to download..." guidance, the M1-specific timing note) are all real usability wins for the target audience.

---

PR #1044, opened 2026-06-09
Fix launch smoke timeout cleanup

r3dbars:
## Summary
- terminate only Transcripted app processes created by the launch smoke run when the UI report times out or validation fails
- preserve pre-existing Transcripted processes from the same build path
- add repo contract coverage for the narrowed cleanup path

## Verification
- bash -n scripts/entrypoints/build.sh
- bash run-tests.sh (4513 passed, 0 failed)
- /Users/redbars/.codex/skills/codex-review/scripts/codex-review --mode branch --base origin/main --full-access --parallel-tests "bash -n scripts/entrypoints/build.sh && bash run-tests.sh"

### DIFF
--- a/Tests/RepoCommandContractTests.swift
+++ b/Tests/RepoCommandContractTests.swift
@@ -1266,6 +1266,15 @@ func testRepoCommandContract() {
             localBuildScript.contains("\"$APP_BINARY\" >\"$smoke_log\""),
             "local launch smoke should not run the app executable directly from sandboxed agent contexts"
         )
+        assertTrue(
+            localBuildScript.contains("pre_launch_app_pids=\"$(snapshot_launch_smoke_app_pids)\"")
+                && localBuildScript.contains("terminate_launch_smoke_app()")
+                && localBuildScript.contains("pgrep -f \"$APP_BINARY\"")
+                && localBuildScript.contains("is_pre_launch_app_pid")
+                && localBuildScript.contains("kill -TERM \"$pid\"")
+                && localBuildScript.contains("kill -KILL \"$pid\""),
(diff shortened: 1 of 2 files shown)


r3dbars (commented):
**Verdict**: APPROVE

**Summary**: This PR narrows the launch smoke cleanup in `build.sh` so that on timeout or validation failure, only app processes spawned by *this* smoke run are killed — pre-existing Transcripted instances are left alone. The approach (snapshot PIDs before launch, diff on cleanup) is clean, correct, and well-tested via a new repo contract assertion.

**Blockers**: None.

**Should fix**: Nothing.

**Nits**:
- The `pgrep -f "$APP_BINARY"` pattern does substring matching on the full command line, so it could theoretically match processes that merely have the binary path as an argument (e.g., a `tail -f` on the binary). In practice this is a non-issue for a build script, and the pre-launch snapshot excludes anything already running, so the blast radius is minimal. Just worth knowing if this pattern gets reused elsewhere.

**What looks good**:
- The SIGTERM → sleep 0.5 → SIGKILL escalation is textbook graceful shutdown. Good to not just `kill -9` immediately.
- All four exit paths in `verify_launch_smoke()` handle cleanup correctly: the two paths where the app is already dead (early crash, post-validation wait failure) correctly skip `terminate_launch_smoke_app`, while the two paths where the app may still be running (timeout, validation failure) call it before killing `open_pid`.
- The `grep -qx` for PID matching is the right choice — exact line match prevents PID "12" from matching PID "123".
- The repo contract test covers the right structural markers without being brittle about implementation details.
- `|| true` on `pgrep` prevents `set -e` from killing the script when no processes match. Easy to forget, glad it's there.

---

PR #1052, opened 2026-06-09
Fix meeting retranscribe dates and review loop

r3dbars:
## Summary
- carry live recording start dates through meeting transcription, failed queues, and retry saves
- group Home meeting rows by transcript recording date while preserving file-recency scan order
- disable saved-meeting re-transcribe for rows with pending speaker review and route users to Speakers first

## Tests
- bash build-deps.sh --force
- swift test --filter TranscriptionTaskManagerMetadataTests/testStartTranscriptionUsesRecordingDateForSavedMetadata
- swift test --filter TranscriptionTaskManagerMetadataTests/testRetryFailedTranscriptionUsesOriginalRecordingDateForSavedMetadata
- bash build.sh --no-open
- bash run-tests.sh
- swift test
- ~/.codex/skills/codex-review/scripts/codex-review --full-access

## Residual risk
- Existing duplicate Markdown rows already created on a user machine are not auto-collapsed by this patch.
- Manual UI proof against Matthew's exact local library was not run; coverage is code-level and scanner/core tests.

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -80,7 +80,8 @@ final class MeetingSessionController: ObservableObject {
                 micURL: URL,
                 systemURL: URL?,
                 healthInfo: RecordingHealthInfo,
-                meetingTitle: String?
+                meetingTitle: String?,
+                recordingDate: Date
             )
             case imported(
                 audioURL: URL,
@@ -108,6 +109,7 @@ final class MeetingSessionController: ObservableObject {
         let healthInfo: RecordingHealthInfo
         let pipelineSnapshot: AudioPipelineDiagnosticsSnapshot
         let suggestedTitle: String?
+        let recordingStartedAt: Date?
     }
 
     private struct BackgroundTranscriptionWorkSnapshot {
@@ -190,6 +192,7 @@ final class MeetingSessionController: ObservableObject {
     private var lastTerminalTranscriptionOutcome: TerminalTranscriptionOutcome?
     private var activeRecordingTrigger: StartTrigger = .unknown
     private var activeRecordingSuggestedTitle: String?
+    private var activeRecordingStartedAt: Date?
     private var activeTranscriptionTrigger: StartTrigger = .unknown
     private var isFinishingRecording = false
     private var shouldSurfaceMeetingWarmupFailure = false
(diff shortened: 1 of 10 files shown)


r3dbars (commented):
**Verdict**: APPROVE (with one test concern to consider)

**Summary**: This PR fixes duplicate meeting rows caused by retranscription stamping "today's" date instead of the original recording date, and prevents a re-transcribe → speaker-review loop by gating the action when speaker review is pending. The recording date is now captured at `Date()` when recording starts and threaded consistently through every path: live transcription, failed-transcription persistence, retry, shutdown preservation, and queue cancellation. The scanner change to use frontmatter dates for Home row grouping is the right fix — rows no longer jump to "Today" after a retranscribe.

**Blockers**: None.

**Should fix**:

1. **Timezone-dependent test assertion** — The updated scanner test compares `meetings.first?.date` against `recentLoaderFrontmatterDate("2026-05-01 14:00:00")`, which parses the string in the local timezone with no explicit TZ. But the frontmatter was written via `writeRecentLoaderMeeting(date: recentLoaderDate("2026-05-01T14:00:00Z"), ...)` which formats the date/time strings in the local timezone representation of 14:00 **UTC**. On a UTC machine (typical CI) these match, but on any non-UTC machine the frontmatter would contain e.g. `time: "07:00:00"` (PDT) while `freshRecordedAt` would represent 14:00 in local time — a 7-hour discrepancy. The fix is to derive `freshRecordedAt` from the same absolute date used to write the frontmatter, e.g.:
   ```swift
   let freshRecordedAt = recentLoaderDate("2026-05-01T14:00:00Z")
   ```
   — or set a fixed timezone on `recentLoaderFrontmatterDate`'s formatter to match the write path.

**Nits**:

1. `hasSpeakerReviewWork(for:)` does a linear scan of `reviewQueueItems` per Home row (via `speakerPeopleModel.hasPendingReview(forTranscript:)`). For a handful of rows this is fine, but if the review queue or visible meetings grow, it could become noticeable. Not a problem today, just something to revisit if performance surfaces.

2. `RecentMeetingRetranscriptionMenuActionPolicy` — the name is a mouthful. The existing `RecentMeetingRetranscriptionActionPolicy` is already verbose; adding "Menu" in the middle makes it harder to distinguish at a glance. Minor naming taste, no action needed.

**What looks good**:

- The date threading is thorough and consistent — every code path that touches the `.recorded` enum case, failed-transcription persistence, shutdown preservation, and retry correctly carries the recording date. No path drops it silently.
- `FailedTranscription.recordingDate` being `Date?` with a `nil` default is the right call for backward compatibility with existing on-disk JSON (auto-synthesized `Codable` decodes missing keys as `nil`). The `failed.recordingDate ?? failed.timestamp` fallback in the retry path is a clean migration strategy for old entries.
- The `RecordingSnapshot` pattern — capturing `activeRecordingStartedAt` into a `let` before handing off to background work — follows the existing thread-safety contract well.
- The speaker-review gate is properly wired at both the menu-item enabled state (via the new policy) and the action handler (guard clause), so the button and the action are always in sync.
- Tests are solid: the two new `TranscriptionTaskManagerMetadataTests` verify both the live and retry paths end-to-end, including filename prefix, frontmatter date/time, and stats metadata. The policy unit tests cover all three meaningful states (pending review, global unavailable, clean).

---

PR #1062, opened 2026-06-10
feat: simplify the Speakers section

r3dbars:
## Summary

Redesigns the Speakers tab (Home) and Speakers settings page around one idea: show the work, not a report about the work.

**Voices to name (new queue)**
- One card per distinct voice instead of one row per voice-per-meeting — `SpeakerReviewQueueScanner.groupedByVoice` groups queue items by `speakerId`, with the newest appearance as the representative and sample text borrowed from an older meeting when the newest has none. Naming the card still names every queued meeting for that voice (existing `namePendingReviewItem` behavior).
- Card layout: primary play button (the avatar), transcript quote, one meta line ("Quick notes · May 30 · Remote · +2 more meetings"), a "Who is this?" field, Save Name, and an open-transcript button.

**All speakers (simplified list)**
- Contacts-style rows: initials avatar (color-hashed), name, "12 meetings · last heard 2 weeks ago" (relative dates), play button (hidden when no clip), and a single ⋯ menu (Rename / Merge Into ▸ / Delete…).
- Possible duplicates compress to one line per pair with playable name chips and a Merge button.

**Removed**
- The status/metric-pills card ("Speaker Inbox", "N to name", "N appearances"), the All/Needs-Review filter, the Show Inbox button, per-row badges duplicating titles, and the four always-visible buttons per row. Counts now appear exactly once, in section subtitles. Rows needing attention still sort to the top (`SpeakerPeopleReviewPolicy` unchanged).

**Reconciliation with #1050**
- #1050 made "Show Inbox" scroll to the pending rows — this redesign removes that button since the queue is now always at the top. Kept from #1050: the `ScrollTarget.reviewQueue` anchor + `transcripted.speakers.inbox` automation ID on the queue section, and the `ScrollViewReader` deep-link so "review speakers" entry points from Home rows still scroll to the queue. Removed: the now-dead `onShowInbox` callback plumbing. The contract test was re-pinned to the new surface.

## Test Coverage

- Two new suites in `Tests/SpeakerReviewQueueScannerTests.swift`: grouping collapses per-meeting items into per-voice groups (order, representative, meeting count) and sample-text fallback.
- `Tests/UIAutomationSurfaceContractTests.swift` re-pins the queue's automation anchor and scroll wiring.
- View-layer changes are covered by the repo's build + contract tests per existing conventions (SwiftUI views are not unit-tested in this repo).

## Pre-Landing Review

- Independent Codex review of the diff: PASS, zero findings (verified the group-naming semantics, removed-member call sites, and pinned contract strings).
- Manual QA on a live build with real data: queue grouping, naming flow, search, ⋯ menu, duplicate-free state, and named/unnamed avatar rendering all verified in-app.

## Verification

- [x] `bash build-deps.sh --force` + `bash build.sh --no-open` pass on the merged state
- [x] `bash run-tests.sh` — 4581 tests, 0 failures (includes the new grouping suites and the re-pinned contract test)


r3dbars:
**Adversarial review (post-merge, full diff):** merge reconciliation sound — no dangling symbols, no crash paths, contract pins verified byte-exact. Fixed in de3b28f1:

- Cold-start scroll race: "review speakers" deep-links fired before the queue loaded and silently no-op'd (scroll target only exists when the queue is non-empty). Now the scroll request is held and re-fired when the queue transitions empty → non-empty.
- Removed dead `SpeakerPendingReviewItem.channelTitle` (only consumer was the deleted row).
- Added accessibility labels to the new icon-only controls (play, open transcript, refresh, row ⋯ menu) so VoiceOver doesn't fall back to SF Symbol names.

Known follow-ups (pre-existing behavior, flagged for a product call, not blocking):
- Merge has no confirmation while Delete does — one click rewrites transcripts with no undo, and the compact rows enlarge the misclick surface.
- Naming a voice clears the field optimistically; if the transcript write fails off-main, the row reappears later with the typed name lost and no feedback.
- A voice heard on both mic and system channels shows a single "In the room"/"Remote" label from its newest appearance.

PR #1108, opened 2026-06-13
Meeting naming: date-prefixed filenames, de-dated titles, and editable titles

acorretti:
## Why

Meeting filenames and titles both carried a `MM dd` date prefix. In the title that's redundant — the frontmatter already has `date`/`time` and the app lists meetings grouped by day and sorted by time — and in the filename a `MM dd` prefix doesn't sort chronologically across years. There was also no way to rename a meeting after the fact.

## Product Impact

- Affects: `meetings` (+ `docs only` for local CLAUDE.md updates)
- Lane: `meeting reliability`
- Why this matters: meeting artifacts now have sortable, date-prefixed filenames and clean, name-only titles, and users can correct a wrong title without leaving the app — the rename keeps the transcript, retained audio, and summary sidecar consistent on disk.

## What changed

- Filenames + retained `audio/<stem>_audio/` directories now use a `YYYY-MM-dd <name>` stem (date from the meeting's recorded date).
- `title:` frontmatter, body `#` heading, and the in-app list drop the date prefix; unnamed meetings fall back to a time-only "Meeting at 3:00 PM" instead of "Jun 13 at 3:00 PM".
- Double-clicking the title in the Home preview sheet opens an inline editor; committing rewrites the title/heading and moves the `.md`, audio dir, and `<stem>.summary.md` sidecar (repointing `source_transcript`) to the canonical stem.
- New `MeetingArtifactRenamer` unifies the rename mechanics shared by the post-save restyle and the manual edit, so naming/sidecar handling can't drift; new `HomeMeetingRename` service is conservative (app-owned `capture_type: meeting` only).
- Folded in a timezone-independence fix for the pre-existing `DictationTranscriptWriter` tests so the branch is green on any machine.

## How I checked it

- [x] `scripts/dev/agent-preflight.sh`
- [x] Selected checks from `.agents/test-matrix.yml` for the files changed
- [x] `bash build.sh --no-open`
- [x] `bash run-tests.sh` — 5248/5248 (added `MeetingTranscriptStyler` updates + new `HomeMeetingRename` coverage)
- [x] Performance budget passed (`build.sh` bundle gate passed; no runtime-sensitive change, so the events-budget script was not needed)
- [x] `bash run-integration-smoke.sh` if I touched `Sources/Meeting/` or `Sources/TranscriptedCore/`
- [ ] `swift test` if I touched `Package.swift`, `Sources/TranscriptedCore/`, or the public core seam — N/A, none of those changed
[...]

PR #1110, opened 2026-06-13
Fix Home dashboard row actions, hover reveal, and stats focus ring

acorretti:
## Why

Three pre-existing Home dashboard bugs (unrelated to the meeting-naming work): the recent-meeting row overflow (⋯) menu actions silently did nothing (Delete, Show transcript/audio in Finder, Report issue), per-row hover only revealed actions in a narrow band over the title text, and the stats line ("Nd streak · …") kept a permanent blue focus ring.

Root cause for the dead menu actions: `HomeRowMoreMenuButton` set the `NSMenuItem.target` to a separately-retained `MenuActionTarget` (a **weak** reference) whose lifetime was tied to the SwiftUI representable's coordinator, then deferred the call via `DispatchQueue.main.async` — the target could be deallocated before the block ran, so the closures never fired. Delete and "Show in Finder" were one bug, not two, and not the URL/custom-library mismatch first suspected (the scanner already resolves the real on-disk URL).

## Product Impact

- Affects: `meetings`
- Lane: `meeting reliability`
- Why this matters: the overflow-menu actions on Home recent meetings (delete, reveal in Finder, report) were completely non-functional, including for capture libraries in custom/legacy locations.

## What changed

- Replace the menu dispatch with a self-targeting `ClosureMenuItem: NSMenuItem` that the `NSMenu` retains for the whole `popUp` loop and invokes synchronously — fixes Delete, Show in Finder, and Report issue. Composes with `main`'s `retainedActionTarget` coordinator-retention fix: the button keeps `showMenu` alive, the menu item owns its handler.
- Give recent-capture rows a full-width `.contentShape(Rectangle())` so hover reveals actions across the whole row instead of only the opaque title text.
- Disable the stats-line button focus ring (`.focusEffectDisabled()`).
- Add a `HomeMeetingDeletion` test covering scan→delete URL resolution for a date-prefixed, accented filename.
- Update the UI automation surface contract to match the merged code, and add source-contract guards for the two affordances that previously had no test coverage at all (`.focusEffectDisabled()` and the full-row row hit shape).

### Changed vs the original PR (after rebasing on latest `main`)

This branch originally also routed delete confirmations through an AppKit `NSAlert` (`presentHomeDeleteConfirmation`), on the basis that SwiftUI `.alert(item:)` didn't present reliably on Home. `main` has since gone the other way — the failed-meeting delete was moved back to the SwiftUI `homeDeleteConfirmation` state, with a test asserting the hand-built NSAlert must not be used. On rebase this branch **drops its NSAlert path entirely** and defers to `main`'s SwiftUI delete confirmation for all deletes, so `TranscriptedSettingsView.swift` ends up unchanged from `main`.

## How I checked it

- [x] `bash run-tests.sh` — full fast suite green (**5260/5260**, UI automation contract 159/159). The `DictationTranscriptWriterTests` timezone flakiness noted in the original PR is gone now that this sits on top of #1108.
- [x] Performance budget passed (bundle gate via `build.sh --no-open`)
- [ ] `bash run-integration-smoke.sh` — N/A (no `Sources/Meeting/` or `Sources/TranscriptedCore/` changes in this branch)
- [ ] `swift test` — N/A (no `Package.swift`/core-seam changes)
- [ ] **Manual re-verification needed**: delete confirmation now goes through `main`'s SwiftUI `.alert(item:)`, which this branch had previously switched away from. Re-verify the Home delete → confirm → delete flow in the built app on latest `main`. Menu firing, full-row hover, and focus-ring fixes are unaffected and were live-verified earlier.

## Risk Review

- [x] Storage path or migration impact reviewed (delete/reveal resolve the real on-disk URL, incl. custom libraries; no migration)
- [x] Public-facing copy stays concrete (delete-confirmation copy unchanged — still `HomeDeleteConfirmationPolicy`)
- [x] Release/update impact reviewed (none)
- [ ] Open item: delete-confirmation presentation now relies on `main`'s SwiftUI alert (see manual re-verification above)

## Notes

- Rebased onto latest `main` (`0c72c29c`), which now carries the meeting-rename feature, the `retainedActionTarget` menu fix, and the SwiftUI failed-meeting delete revert. Conflicts resolved by keeping this branch's `ClosureMenuItem` + hover/focus-ring fixes and adopting `main`'s SwiftUI delete confirmation.

### DIFF
--- a/Sources/UI/Settings/HomeView.swift
+++ b/Sources/UI/Settings/HomeView.swift
@@ -430,6 +430,7 @@ struct HomeCanvasHeader: View {
                 .contentShape(Rectangle())
             }
             .buttonStyle(.plain)
+            .focusEffectDisabled()
             .opacity(isHovering ? 1 : 0.9)
             .onHover { isHovering = $0 }
             .animation(.easeOut(duration: 0.12), value: isHovering)
@@ -799,7 +800,6 @@ struct HomeRowMoreMenuButton: NSViewRepresentable {
 
     final class Coordinator: NSObject {
         var items: [HomeRowMenuItem]
-        private var menuActionTargets: [MenuActionTarget] = []
 
         init(items: [HomeRowMenuItem]) {
             self.items = items
@@ -807,20 +807,11 @@ struct HomeRowMoreMenuButton: NSViewRepresentable {
 
         @objc func showMenu(_ sender: NSButton) {
             let menu = NSMenu()
-            menuActionTargets = []
             // Without this, AppKit auto-enables every item whose target responds
             // to the action, overriding the per-item isEnabled set below.
             menu.autoenablesItems = false
             for item in items {
-                let target = MenuActionTarget(item: item)
-                menuActionTargets.append(target)
-                let menuItem = NSMenuItem(
-                    title: item.title,
-                    action: #selector(MenuActionTarget.perform(_:)),
-                    keyEquivalent: ""
-                )
-                menuItem.target = target
-                menuItem.isEnabled = item.isEnabled
+                let menuItem = ClosureMenuItem(menuItem: item)
                 if let image = NSImage(systemSymbolName: item.symbolName, accessibilityDescription: item.title) {
                     image.isTemplate = true
                     menuItem.image = image.withSymbolConfiguration(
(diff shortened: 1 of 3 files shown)


r3dbars:
Thanks for sending this over, Alfonso. I’m going to compare it against latest `main` and manually re-check the Home delete flow before deciding whether to merge.

r3dbars:
Thanks again. I reviewed this against latest `main`, and the core AppKit lifetime fix looks right to me: making the retained `NSMenuItem` own the closure is the right shape for the weak-target issue.

Local proof on my side:
- `bash build.sh --no-open`
- `bash run-tests.sh` — 5260 passed
- independent Codex review — no actionable findings

Only thing I still want before merging is the manual Home delete confirmation click-through on the built app, since the PR correctly notes that path now relies on the SwiftUI alert already on `main`.

r3dbars:
Thanks again, Alfonso. I’m going to bring this onto `main` now and preserve the PR history with a merge commit.

r3dbars:
Thanks again, Alfonso. I confirmed this Home row-actions work is already merged into `main` at `af12e229`; I also verified the newer auto-call-detection merge on top of current `main`. Appreciate the fixes here too.

PR #1119, opened 2026-06-14
Detect spontaneous calls via mic activity and offer to record

### LINKED ISSUE #1118: Detect spontaneous calls and offer to record, even with no calendar invite
**What's the job to be done?**

When I jump on a call, including a spontaneous Google Meet I start on the spot with no calendar invite, I want Transcripted to notice and offer to record it instead of having to remember to hit ⌥M every time.

**What's awkward today?**

Transcripted already prompts around calendar meetings and when a native conferencing app is open, but a spontaneous call in the browser is invisible. Google Meet runs in a tab, so there's no app to match on, and with no calendar event there's nothing to anchor a prompt to. The ad-hoc calls I'm most likely to forget to record are exactly the ones it misses.

**Area**

- Meetings

**What would feel better?**

A short prompt the moment a call actually starts, like "Call detected, start recording?", the way Notion and Plaud do it. Detect the act of being in a live call generically (any app or browser tab that starts using the mic), attribute it to a provider, and offer one-tap Record (or ⌥M). It should stay quiet for things that aren't calls, and never prompt while Transcripted itself is recording.

**Why this matters**

Ad-hoc calls are where local, on-device capture really shines. They're the conversations cloud meeting bots can't join, and the ones people most often forget to record. Closing this gap makes "just hit record" automatic for exactly those calls.

**Alternatives considered**

Reading the browser's active-tab URL via AppleScript. That one is fragile, needs a per-browser Automation prompt, and is easy to get wrong. A microphone-activity signal is app-agnostic and needs no extra permission. Per-call network heuristics (WebRTC socket inspection) could harden it later, but plain mic-in-use is a strong enough signal on its own for a first cut.

**Anything else?**

Everything stays on-device. It only reads which app holds the mic (metadata), never audio, and needs no extra TCC permission. Should ship behind a clear Settings toggle. Happy to follow up with a PR.

acorretti:
## Why

Transcripted prompts for scheduled or visible meeting apps, but spontaneous browser calls can be invisible. This adds a local mic-activity signal so Transcripted can offer to record ad-hoc calls before users forget. Closes #1118.

## Current status

- Rebased/merged onto current `main` in commit `facd5bc5`.
- GitHub reports the PR conflict-free / mergeable.
- GitHub checks are green: `repo-hygiene` pass, `build-and-test` pass, `hardware-smokes` skipped.

## What changed

- Added `MicActivityMonitor`, a Core Audio process-object watcher that emits non-Transcripted bundle IDs using the mic. It observes metadata only; no audio tap and no new TCC permission.
- Added tested mic bundle/provider heuristics and a Settings toggle for Auto-detect calls.
- Fed mic activity into `MeetingPromptDetector` as a `.micInput` reason while reusing the existing prompt, snooze, dismiss, and backoff machinery.
- Hardened the prompt path after review:
  - suppresses mic prompts while Transcripted is already recording
  - suppresses stale callbacks after the toggle is disabled
  - preserves pending/dismiss/snooze cooldowns across transient mic drops
  - avoids repeat prompts for the same active mic session
  - lets native Zoom/Teams mic candidates keep a matching calendar title
  - keeps generic browser mic prompts neutral to avoid mislabeling Meet vs Zoom-web vs Teams-web
  - backs off CoreAudio fallback polling to 60s
- Updated docs/spec and fast-test coverage.

## Proof from current head

Automation run by Codex after the final hardening commits:

- [x] `bash scripts/dev/agent-preflight.sh`
- [x] `bash -n run-tests.sh`
- [x] `bash -n scripts/entrypoints/run-tests.sh`
- [x] `bash run-tests.sh` — 5342/5342 passed
- [x] `bash scripts/ops/transcripted-qa-bench.sh --mode full` — PASS, 16/16 checks
- [x] GitHub `repo-hygiene`
- [x] GitHub `build-and-test`

Review proof:

- `codex-review` was attempted but the local Codex CLI was usage-limited, so it did not produce a verdict.
- Fallback independent Claude Code review found prompt-spam and labeling issues; the branch was patched.
- Final focused Claude Code review of the last hardening commit: `no actionable findings`.

Manual proof boundary:

- Original author proof says a live spontaneous Google Meet showed browser-call copy and a native Zoom call showed Zoom copy.
- This Codex worker did not rerun manual live-call hardware proof after the rebase; the new proof above is automated/local/CI plus independent review.

## Risk notes

- Browser mic activity is intentionally neutral. A real Google Meet in the browser may show generic browser copy, because the mic process alone cannot distinguish Meet from Zoom-web or Teams-web safely.
- Phase 2 per-call network/WebRTC confirmation remains deferred until real-world false positives justify it.


r3dbars:
Thanks for digging into this, Alfonso. The spontaneous-call angle is useful, and I’m going to review it carefully against latest `main` before deciding whether this should land as-is or needs reshaping.

r3dbars:
Thanks again. I reviewed this against latest `main`; the underlying idea is strong and the CoreAudio process-object approach looks like a real way to close the spontaneous-call gap.

I would not land this branch as-is yet, though. Current `main` has moved more prompt policy into the synthetic prompt evaluator, and this branch predates that work. The auto-merge only hard-conflicts in `Tests/FastTests.manifest`, but the mic-input path also needs to be rebased/adapted so it fits the newer prompt architecture and keeps the synthetic coverage story clean.

Suggested next step: rebase on current `main`, keep both `SyntheticMeetingPromptTests` and the new mic tests in the manifest, remove the small duplicate `@available` typo in `TranscriptedApp.swift`, then rerun the mapped checks: `bash build-deps.sh --force`, `bash build.sh --no-open`, `bash run-tests.sh`, and `bash run-integration-smoke.sh`.

r3dbars:
Thanks again, Alfonso. I see this has been updated/rebased cleanly now, so I’m going to bring it onto `main` after the Home row-actions PR and run the repo checks on the combined result.

r3dbars:
Thanks again, Alfonso. This is merged into `main` now at `ee798805`.

I reviewed the updated branch against the newest `main`; the mic-activity approach looks solid, and the backoff/self-capture guards are covered by tests. I also ran the repo gates on the merged result:

- `bash build.sh --no-open`
- `bash run-tests.sh` (5347 passed)
- `bash -n run-tests.sh && bash -n scripts/entrypoints/run-tests.sh`
- `codex-review --mode branch --base origin/main` (clean, no actionable findings)

Really appreciate the contribution.

PR #1126, opened 2026-06-14
Fix dead Home "Delete meeting" confirmation

### LINKED ISSUE #1125: Home "Delete meeting" confirmation never appears
Reported on `main` after rebasing onto upstream.

## What happens

In Settings → Home, opening a recent meeting's ⋯ overflow menu and choosing **Delete meeting** does nothing: no confirmation dialog, no deletion. The other menu actions behave like this:

- **Show transcript in Finder** / **Show audio in Finder** — work
- **Report issue** — works (opens the feedback sheet)
- **Delete meeting** — dead (no popup, no action)

## Root cause

Two stacked defects on the Home recent-meeting row path:

1. **Menu handler ran inside the menu modal loop.** `ClosureMenuItem.invoke()` (in `HomeView.swift`) called its handler synchronously, while still inside `NSMenu.popUp`'s modal event-tracking loop. SwiftUI will not present an `.alert(item:)` / `.sheet(item:)` from that nested loop, so the Delete handler set `homeDeleteConfirmation` but nothing showed. PR #1110 fixed the handler firing at all (it previously no-op'd on a deallocated weak target), which exposed this next layer.

2. **Three legacy `.alert(item:)` were stacked on one view.** `homeDeleteConfirmation`, `homeDeleteFailure`, and `pendingAudioRetentionWindow` each had their own `.alert(item:)` on the settings root (`TranscriptedSettingsView.swift`). SwiftUI keeps only the last stacked legacy alert, so the delete confirmation (first) and the delete-failure alert were both shadowed. Only the audio-retention alert (last) could present.

**Report issue** opens a `.sheet`, and stacked sheets do not shadow each other, so it only needed fix 1.

## Repro

1. Settings → Home
2. ⋯ on any recent meeting → **Delete meeting**
3. Expected: confirmation dialog. Actual: nothing.

## Fix

- Defer `ClosureMenuItem`'s handler to the next main-runloop turn (off the menu loop), keeping the handler owned by the item so it cannot deallocate first.
- Collapse the three stacked alerts into a single `.alert(item: rootAlertBinding)` that presents whichever state is set; call sites unchanged.

Fix + source-contract coverage in the linked PR.

acorretti:
## Why

The Home recent-meeting ⋯ menu's **Delete meeting** action did nothing: no confirmation dialog, no deletion. Fixes #1125.

Two stacked defects on the Home row path:

1. `ClosureMenuItem.invoke()` ran its handler synchronously, inside `NSMenu.popUp`'s modal tracking loop. SwiftUI won't present an `.alert(item:)` / `.sheet(item:)` from that nested loop, so the Delete handler set `homeDeleteConfirmation` but nothing showed. #1110 made the menu handler fire at all (it previously no-op'd on a deallocated weak target), which uncovered this layer.

2. Three legacy `.alert(item:)` modifiers (`homeDeleteConfirmation`, `homeDeleteFailure`, `pendingAudioRetentionWindow`) were stacked on the settings root view. SwiftUI keeps only the last stacked legacy alert, so the delete confirmation (first) and the delete-failure alert were both shadowed. Only the audio-retention alert (last) could ever present.

The tell while debugging: **Report issue** (a `.sheet`) started working once the defer landed, but **Delete** (an `.alert`) still didn't, because stacked sheets don't shadow each other and stacked legacy alerts do.

## Product Impact

- Affects: `meetings`
- Lane: `meeting reliability`
- Why this matters: deleting a recent meeting from Home was impossible through the menu, and the delete-failure alert (plus any non-last stacked alert) was silently shadowed too.

## What changed

- `ClosureMenuItem.invoke()` now defers its handler to the next main-runloop turn via `DispatchQueue.main.async { [handler] in handler() }`, off the `NSMenu.popUp` tracking loop. The handler stays owned by the item (the capture keeps it alive), so the deallocation trap #1110 fixed does not come back.
- The three Home alert states route through a single `.alert(item: rootAlertBinding)` instead of three stacked `.alert(item:)`. A small private `RootAlert` enum plus `activeRootAlert` / `rootAlertBinding` pick the active state and clear it on dismiss. All seven existing assignment sites are unchanged.
- Source-contract guards in `UIAutomationSurfaceContractTests.swift`: the menu handler must defer, the three alerts must share one presenter, and no `.alert(item: $homeDelete…)` / `$pendingAudioRetentionWindow` may be re-stacked.

## How I checked it

- [x] Selected checks from `.agents/test-matrix.yml` for the files changed (`Sources/UI/**` + root `Tests/*.swift` map to build + fast tests)
- [x] `bash run-tests.sh` (5353/5353)
- [x] Performance budget passed (`build.sh --no-open` bundle gate: "Performance budget OK")
- [ ] `bash run-integration-smoke.sh` (N/A, no `Sources/Meeting/` or `Sources/TranscriptedCore/` changes)
- [ ] `swift test` (N/A, no `Package.swift` / core seam changes)
- [x] Manual check: Settings → Home → ⋯ → **Delete meeting** shows the confirmation; confirm deletes, cancel dismisses cleanly. **Report issue** sheet and **Show in Finder** still work. The failed-meeting delete and the Storage audio-retention prompt (the other two states now sharing the presenter) still confirm.

## Risk Review

- [x] Storage path or migration impact reviewed (delete still resolves the real on-disk URL through the existing `HomeMeetingDeletion` path; no migration)
- [x] Public-facing copy stays concrete and matches current product scope (delete-confirmation copy unchanged, still `HomeDeleteConfirmationPolicy`)
- [x] Release/update impact reviewed (none; no version, appcast, or updater flow touched)
- [ ] UI changes include sanitized `.agent-review/visuals/` evidence (not attached; this makes an existing dialog present with unchanged copy, and the live Home list shows private meeting titles. Happy to add a cropped capture of just the confirmation dialog if wanted.)

## Notes

- Fixes #1125. Builds on #1110, which made the row-menu handlers fire; this fixes the next layer (menu-loop timing plus alert shadowing).
- No legacy-branch impact.

### DIFF
--- a/Sources/UI/Settings/HomeView.swift
+++ b/Sources/UI/Settings/HomeView.swift
@@ -837,13 +837,20 @@ struct HomeRowMoreMenuButton: NSViewRepresentable {
 
     /// A menu item that owns its action closure and acts as its own target.
     ///
-    /// The menu retains its items for the whole `popUp` tracking loop, so the
-    /// handler stays alive and fires no matter how SwiftUI tears down and
-    /// recreates this representable's coordinator while the menu is open. The
-    /// earlier design stored a separate target on the (weakly referenced)
-    /// `NSMenuItem.target` and deferred the call with `DispatchQueue.main.async`;
-    /// the target could be deallocated before the deferred block ran, so the
-    /// closures silently never fired (delete, reveal, and report all no-op'd).
+    /// Two things have to hold for a handler to both fire *and* be able to drive
+    /// SwiftUI presentation:
+    ///   1. The item owns its handler instead of pointing `NSMenuItem.target` at
+    ///      a separately, weakly-retained object. The `NSMenu` retains its items
+    ///      for the whole `popUp` tracking loop, so the handler can't be torn
+    ///      down with the SwiftUI coordinator while the menu is open. (The old
+    ///      design's weak target could deallocate first, so closures silently
+    ///      never fired — delete, reveal, and report all no-op'd.)
+    ///   2. The handler runs on the next main-runloop turn, *after* `popUp`'s
+    ///      modal tracking loop exits. A handler that mutates SwiftUI state to
+    ///      present an `.alert(item:)`/`.sheet(item:)` (e.g. the Home delete
+    ///      confirmation) won't present if it runs synchronously inside that
+    ///      loop. The async block captures `handler` strongly, so deferring is
+    ///      safe here — the dealloc trap from (1) does not reappear.
     final class ClosureMenuItem: NSMenuItem {
         private let handler: () -> Void
 
@@ -861,7 +868,7 @@ struct HomeRowMoreMenuButton: NSViewRepresentable {
 
         @objc private func invoke() {
             guard isEnabled else { return }
-            handler()
+            DispatchQueue.main.async { [handler] in handler() }
         }
     }
 
--- a/Sources/UI/Settings/TranscriptedSettingsView.swift
+++ b/Sources/UI/Settings/TranscriptedSettingsView.swift
@@ -216,32 +216,38 @@ struct TranscriptedSettingsView: View {
                 }
             )
         }
-        .alert(item: $homeDeleteConfirmation) { confirmation in
-            Alert(
-                title: Text(confirmation.title),
-                message: Text(confirmation.message),
-                primaryButton: .destructive(Text(confirmation.confirmTitle)) {
-                    confirmation.perform()
-                },
-                secondaryButton: .cancel()
-            )
-        }
-        .alert(item: $homeDeleteFailure) { failure in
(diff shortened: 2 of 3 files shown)


r3dbars:
Thanks Alfonso — this is the right area and the CI/fast suite is green, but I'm holding merge for one issue from codex-review.

The new shared alert binding clears all three alert states when SwiftUI writes nil. If a confirmation action fails synchronously during dismissal, for example dictation delete throws or failed-meeting delete returns false, the follow-up homeDeleteFailure can be cleared before it presents.

Could you adjust the dismissal path to clear only the dismissed alert, or defer the failure alert until after dismissal?

Proof run here: codex-review --mode branch against origin/main; inside that, bash run-tests.sh passed 5353/5353.

PR #1131, opened 2026-06-15
Fix Home delete-failure alert cleared on confirmation dismissal

acorretti:
## Why

Follow-up to #1126, which consolidated the three stacked Home `.alert(item:)` into one shared `rootAlertBinding`. codex-review flagged (and it shipped in the merge): the binding clears all three alert states whenever SwiftUI writes nil on dismissal. When a confirm action fails *synchronously* it sets `homeDeleteFailure` inside the confirm closure, then SwiftUI dismisses the confirmation by writing nil, which wipes the just-set failure before it can present. So a failed delete silently shows no error.

Affected paths:
- Dictation delete throws → `presentHomeDeleteFailure(...)` in the confirm closure
- Failed-meeting delete returns false → same

(The meeting-delete path fails asynchronously, after the alert is gone, which is why it wasn't caught at merge time.)

## Product Impact

- Affects: `meetings`, `dictation`
- Lane: `meeting reliability`
- Why this matters: a delete that fails (locked file, permissions, etc.) gave the user no feedback at all. The "Could not delete…" alert was created and then immediately wiped by the confirmation's own dismissal.

## What changed

- Dismiss only the active alert, not all three. Routing goes through a new Foundation-pure `HomeRootAlertPolicy.activeSlot`; the confirmation outranks the failure, so dismissal clears the confirmation and a freshly-set failure survives to present next.
- Defer `presentHomeActionFailure` to the next runloop turn so a failure raised inside a confirm action lands after the confirmation finishes dismissing (SwiftUI won't present a second alert mid-dismissal).
- Add `HomeRootAlertPolicyTests` covering the priority/dismissal, including the exact regression case (confirmation + failure both set → confirmation dismissed first, failure survives), plus a source-contract guard that the binding clears only the dismissed alert.

## How I checked it

- [x] Selected checks from `.agents/test-matrix.yml` (`Sources/UI/**` + root `Tests/*.swift` map to build + fast tests)
- [x] `bash run-tests.sh` (5413/5413)
- [x] Performance budget passed (`build.sh --no-open` bundle gate: "Performance budget OK")
- [ ] `bash run-integration-smoke.sh` (N/A, no `Sources/Meeting/` or `Sources/TranscriptedCore/` changes)
- [ ] `swift test` (N/A, no `Package.swift` / core seam changes)
- [x] Manual check: normal delete path verified (confirm deletes, cancel dismisses). The failure path is hard to trigger by hand, so it's locked down by `HomeRootAlertPolicyTests` instead.

## Risk Review

- [x] Storage path or migration impact reviewed (deletion path unchanged; this only touches alert presentation/dismissal)
- [x] Public-facing copy stays concrete and matches current product scope (alert copy unchanged)
- [x] Release/update impact reviewed (none)
- [ ] Agent PRs link the issue/workpad and stay draft until human review (N/A, human-overseen PR)
- [ ] UI changes include sanitized `.agent-review/visuals/` evidence (not attached; behavioral fix to alert sequencing with no visual change, and the live Home list shows private titles. Can add a cropped capture of the failure alert if wanted.)

## Notes

- Fixes the codex-review finding on #1126: https://github.com/r3dbars/transcripted/pull/1126#issuecomment-4703740095
- Regression was introduced by #1126's alert consolidation; this is the follow-up.
- No legacy-branch impact.


acorretti:
@r3dbars I was working on this follow-up to #1126 when I saw you went ahead and merged it.
I'm creating this PR with the fix instead.

PR #1135, opened 2026-06-16
fix(audio): stop dictation engine before removing input tap (isSink || tap != nullptr crash)

r3dbars:
## Crash being fixed

Fatal audio-thread crash, found via Sentry telemetry (org `r3dbars`, project `apple-macos`):

> `com.apple.coreaudio.avfaudio: required condition is false: isSink || tap != nullptr`

Sentry issue **7479438309** (`level=fatal`, `mechanism=nsexception`, `handled=no`). Crash context: `session_kind=meeting/dictation`, `session_stage=recording`. Stack terminates in `AVAudioEngineGraph::InputAvailable` → `+[NSException raise:format:]` on the CoreAudio IO thread.

This crash family correlates with the **#1 error event** `parakeet.device_change_rewarm_failed` (×207 / 14d) and `parakeet.zombie_engine_recovery_failed` — all in the dictation engine's restart/teardown path. Native audio-thread crashes are systematically under-reported, and `app.unclean_shutdown_detected` (×166) indicates many more uncaptured aborts.

## Root cause

`ParakeetEngine` removed the `AVAudioEngine` input tap **while the engine was still running**, then stopped the engine afterward. The canonical stop path is:

```swift
await removeRecordingTap()   // removes tap while engine STILL recording
await stopAudioEngine()      // stops engine afterward
```

Between those two steps (and in four explicit teardown sites that did `removeTap()` → `stop()` → `reset()`), CoreAudio's IO thread can deliver an `InputAvailable` callback to an input node that now has **neither a tap nor a downstream sink**, which trips AVFoundation's fatal `isSink || tap != nullptr` assertion and aborts the process.

The meeting/mic capture path already does this correctly via `AudioInputTapTeardownPolicy` (**stop → drain in-flight callbacks → removeTap**). The dictation path did not.

## Fix

- Promote `AudioInputTapTeardownPolicy` / `AudioInputTapTeardownStep` to `public` so both capture paths share one source of truth.
- Add `ParakeetEngine.safelyRemoveInputTap(on:)` applying that policy, and route **all five** dictation teardown sites through it:
  - `removeRecordingTap` (the dominant stop path)
  - `cleanUpLateAudioStart`
  - `resetAudioGraphAfterStartFailure`
  - `rebuildAudioEngine`
  - `deinit`
- The start-path `removeTap` immediately preceding `installTap` (prewarm optimization) is intentionally left as-is; see "Residual" below.

## Test

Added `testRunningEngineTeardownNeverRemovesTapBeforeStopping` asserting a running-engine teardown never removes the tap before `.stopEngine`. Both capture paths now share this policy, so the contract guards both. (Direct `ParakeetEngine` unit tests aren't feasible — it requires a live `AVAudioEngine` and isn't part of the SPM test target.)

## Verification

- All changed files pass `swiftc -frontend -parse`.
- Full `swift test` requires the heavy `build-deps.sh` ML toolchain (multi-GB, network) which wasn't completed in the authoring sandbox; the fix mirrors the already-shipping, already-tested meeting path exactly. Please confirm CI green before merge.

## Residual / follow-ups (not in this PR)

- `installTapAndStartEngine` removes then immediately re-installs the tap while the (prewarmed) engine may be running — a much smaller race kept to preserve prewarm latency. Worth a follow-up.
- Other fatal Sentry crashes ranked for follow-up: `EXC_BAD_ACCESS` in `objc_release` (×3, no in-app frames), `NSInvalidArgumentException` unrecognized-selector on an NSButton action (×2, selector name scrubbed by privacy sanitizer), `EXC_BREAKPOINT` in `TdtDecoderV3.runJointPrepared` (×1, CoreML-internal, likely OOM).

### DIFF
--- a/Sources/Speech/ParakeetEngine.swift
+++ b/Sources/Speech/ParakeetEngine.swift
@@ -232,11 +232,33 @@ class ParakeetEngine: ObservableObject {
         }
     }
 
-    private static func cleanUpLateAudioStart(on audioEngine: AVAudioEngine) {
-        audioEngine.inputNode.removeTap(onBus: 0)
-        if audioEngine.isRunning {
-            audioEngine.stop()
+    /// Remove the input tap without tripping AVAudioEngine's
+    /// `required condition is false: isSink || tap != nullptr` assertion.
+    ///
+    /// Removing a tap while the engine is still running leaves the input node
+    /// with neither a tap nor a downstream sink; the next IO-thread
+    /// `InputAvailable` callback then crashes the process. Stop the engine and
+    /// let in-flight input callbacks drain BEFORE removing the tap, mirroring
+    /// the meeting/mic path's `Audio.tearDownInputTapSafely` via the shared
+    /// `AudioInputTapTeardownPolicy`.
+    ///
+    /// Must be called on `audioEngineQueue` (callers already hop there); the
+    /// drain step briefly blocks that queue, never the CoreAudio render thread.
+    private nonisolated static func safelyRemoveInputTap(on audioEngine: AVAudioEngine) {
+        for step in AudioInputTapTeardownPolicy.steps(engineIsRunning: audioEngine.isRunning) {
+            switch step {
+            case .stopEngine:
+                audioEngine.stop()
+            case .waitForStoppedInputCallbacks:
+                Thread.sleep(forTimeInterval: AudioInputTapTeardownPolicy.inputCallbackDrainDelay)
+            case .removeInputTap:
+                audioEngine.inputNode.removeTap(onBus: 0)
+            }
         }
+    }
+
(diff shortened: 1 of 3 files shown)


r3dbars:
✅ **Local verification now complete** (deps built):

- `swift test --filter AudioInitializationTests` → **12/12 pass**, including the new `testRunningEngineTeardownNeverRemovesTapBeforeStopping` and the existing `testInputTapTeardownStopsRunningEngineBeforeRemovingTap`. The `public` policy change compiles cleanly in `TranscriptedCore`.
- `bash build.sh` → **full app build succeeds** (exit 0), so the `Sources/Speech/ParakeetEngine` changes compile in the app target, and the post-build launch smoke check + performance budget pass.

Supersedes the "build not yet confirmed" caveat in the PR description.

PR #1138, opened 2026-06-16
feat(audio): share meeting mic with dictation (resurrected branch — needs review)

r3dbars:
**Resurrected branch** — opened as a draft by branch-triage. Was pushed 2026-06-06 with no PR and would otherwise have been pruned. Needs human review before it's ready.

## What it is
Lets meeting capture and dictation share the same microphone instead of blocking each other. Earlier the branch experimented with hard exclusivity ("Block overlapping meeting and dictation capture"), reverted it, and landed on a shared-mic relay instead.

Adds (all novel — none of this exists on `origin/main`):
- `MeetingSessionController.canShareMicWithDictation` + `installSharedDictationMicRelay()` / `clearSharedDictationMicRelay()`
- `STTRouter` / `ParakeetEngine`: `startRecordingFromSharedMeetingMic()`, `handoffActiveRecordingToSharedMeetingMic()`, `appendSharedMeetingMicBuffer()`, `isRecordingFromSharedMeetingMic`, resume-after-shared-mic handling
- `MeetingCaptureBridge` shared dictation-mic handler plumbing
- `DictationSessionController` integration + `RepoCommandContractTests` coverage

7 files, +298/-7, including tests.

## Status / caveats
- 3 commits, branch is ~462 commits behind `origin/main` but **merges cleanly (0 conflicts)** against current main as of triage.
- `grep canShareMicWithDictation origin/main` → 0 hits, so this is not superseded by anything merged.
- Touches real-time audio paths (ParakeetEngine, CoreAudio capture relay) — needs careful review against the "no I/O in CoreAudio callbacks" rule and the recent parakeet tap-teardown crash fix (PR #1135) before merge.

Draft intentionally. Close if the shared-mic direction has been abandoned.


r3dbars:
HOLD review: structural fix needed before merge.

The shared-mic direction may be right, but this implementation puts too much state and callback work in the wrong places.

Blocking findings:

1. The borrowed dictation path does real work on the meeting mic live-buffer callback. `MeetingCaptureBridge+LivePreview` fans the callback straight into `appendSharedMeetingMicBuffer`, and that method extracts samples, takes `NSLock`, grows/trims arrays with `append` / `removeFirst`, updates shared flags, and schedules a `Task` (`Sources/Meeting/MeetingCaptureBridge+LivePreview.swift:60`, `Sources/Meeting/MeetingSessionController.swift:468`, `Sources/Speech/ParakeetEngine.swift:2351`). That callback is explicitly real-time-sensitive. This should be a bounded copy/enqueue handoff like the existing live meeting transcriber pattern, with dictation buffer mutation off the capture callback.

2. Meeting start is not atomic. `MeetingSessionController.startRecording` installs the relay and hands active dictation into shared-mic mode before `capture.startRecording()` has proven the meeting mic is running (`Sources/Meeting/MeetingSessionController.swift:403`). If meeting start stalls or fails, dictation has already torn down its own tap and entered a borrowed state with no guaranteed source, then recovery tries to patch it back (`:410`-`:415`). This needs an explicit transaction/coordinator shape: acquire meeting capture readiness, switch dictation source, and rollback from one owner.

3. `ParakeetEngine` now has a second recording mode spread across booleans and cleanup paths: `sharedMeetingMicRecording`, `acceptsSharedMeetingMicBuffers`, `isRecording`, `preservingRecordingAcrossRecovery`, plus repeated resets in failure/cancel/cleanup (`Sources/Speech/ParakeetEngine.swift:30`, `:2248`, `:2474`, `:3023`). That makes the dictation engine know too much about meeting ownership and makes recovery harder to reason about. This wants a typed recording-source state or a small shared-audio relay abstraction, not more flags in the 3k-line engine.

Also, the added test is mostly a string-contract test proving symbols exist, not behavior under start failure, stop while dictating, route change, wake, or meeting cancel.

Checks I ran: `gh pr view`, full `origin/main...HEAD` diff, `bash scripts/dev/agent-preflight.sh`, `git diff --check`, and `bash build-deps.sh --force`. `build-deps --force` compiled the Swift dependency graph and archives, then failed locally at `LLVM ERROR: Failed to write .../deps-libs/mlx.metallib`, which left deps incomplete; `bash build.sh --no-open` then correctly refused to run because deps were missing. Manual mic/system-audio proof remains UNKNOWN.

Recommendation: hold and rebuild/decompose around a real shared mic relay or source-state abstraction before merge.

r3dbars:
Dirty PR cleanup note: still conflict-dirty and still high risk because it changes meeting/dictation audio ownership. Keep as release hold; if revived, it needs a clean branch plus explicit audio/hardware proof rather than dirty merge cleanup.

r3dbars:
Closing this stale shared-mic experiment. Current meeting + dictation behavior is acceptable, and this branch touches risky live audio paths that we do not want to carry forward.

PR #1215, opened 2026-06-19
Add privacy-safe agent capture query telemetry

r3dbars:
## Summary
- add MCP-side PostHog telemetry for successful saved-capture reads/searches as agent_capture_query_observed
- allowlist the event in app analytics policy and update funnel/health docs and scripts
- add privacy tests for enum/bucket-only payloads and forbidden query/transcript/path/title/source fields

## Privacy proof
- emitted only after successful MCP list/read/search/recent/who_is/recap results against saved captures
- properties are limited to agent_target, query_kind, artifact_kind, result, surface, return_window_bucket, capture_age_bucket, source_count_bucket
- tests prove raw query text, transcript text, file paths, meeting titles, speaker names, source app names, URLs, tokens, and raw IDs are stripped or rejected

## Verification
- swift test --package-path Tools/TranscriptedMCP
- bash build-deps.sh --force
- bash build.sh --no-open
- bash run-tests.sh


r3dbars:
Superseded by #1240, which rebases and tightens this work onto current main: MCP emits agent_capture_query_observed only for successful read/search/recap/speaker lookup, with opt-out-aware helper config and passing verification.

PR #1229, opened 2026-06-20
fix(privacy): exclude live overlay + transcript from screen share

r3dbars:
## Problem

The floating overlay panels leave `sharingType` at the AppKit default (`.readOnly`), which ScreenCaptureKit captures. `grep -rn sharingType Sources/ Tests/` returned **zero hits** before this change.

During a recorded call the user is screen-sharing, the **live transcript drawer** (hosted in `MeetingOverlayPanel`) — including the other party's transcribed **"Them:"** lines — and the dictation overlay get painted back onto the shared screen, silently, with no macOS warning. For a tool whose whole premise is *privately* recording a call, this is the worst-case trust failure, and it fires on the exact happy path.

## Fix

Set `self.sharingType = .none` in all **three** overlay `NSPanel` inits:

- `Sources/UI/Overlay/FloatingOverlayPanel.swift` — dictation overlay
- `Sources/UI/Overlay/MeetingOverlayController.swift` → `MeetingOverlayPanel` — recording pill + live transcript drawer host (the on-screen render path: `MeetingOverlayController` → `MeetingLiveTranscriptDrawerView`)
- `Sources/UI/Overlay/MeetingOverlayController.swift` → `MeetingOverlayTooltipPanel` — the hover tooltip (the easy-to-miss third panel)

`.none` excludes each window from screen capture / sharing.

## Test

`Tests/OverlayScreenSharePrivacyTests.swift`:
- **Behavioral** — instantiates `FloatingOverlayPanel` (dependency-free) and asserts the real runtime `sharingType == .none`.
- **Source-contract** — guards that every overlay `NSPanel` init sets `sharingType = .none`. The two meeting panels live in a heavy controller the fast runner can't compile in isolation, so they're guarded at source level.

`FloatingOverlayPanel.swift` was added to the fast-runner `APP_SOURCES` so the behavioral test can link it.

## Verification

- `bash build.sh --no-open` ✅ (launch smoke passed)
- `bash run-tests.sh` ✅ (10235 passed, incl. the new guard)
- `python3 scripts/dev/check-build-source-lists.py` ✅

## Manual proof still needed

⚠️ Confirm on a **real screen share** (Zoom/Meet/QuickTime screen recording) that the dictation overlay and the meeting recording pill + live transcript drawer no longer appear in the shared/captured video. `sharingType = .none` is the documented mechanism, but the end-to-end "it's actually hidden from the share" check needs a human with a live share.


r3dbars:
Superseded by #1267, which merged today and covers the same privacy fix on the current main line. This branch is now conflict-dirty, so keeping it open would make it look like a release blocker when the fix is already in main.

PR #1175, opened 2026-06-19
feat(speaker): optional ERes2Net on-device voiceprint for speaker identity

r3dbars:
## What

Adds **ERes2Net** (Alibaba 3D-Speaker, 192-d) as an **optional** on-device speaker-identity voiceprint, replacing the diarizer's built-in WeSpeaker (256-d) for same-voice consolidation and cross-call `SpeakerDatabase` matching. ERes2Net is markedly more robust to compressed Zoom/phone audio, where WeSpeaker's different-speaker embeddings drift together and get merged into one person.

Default behavior is unchanged (WeSpeaker). Opt in via **Settings → "Speaker voiceprint"** or `TRANSCRIPTED_SPEAKER_EMBEDDER=eres2net`.

Productionizes the research in `Tools/SpeakerEvalHarness/` (ERes2Net ~halves within-meeting DER on compressed audio and is the only accuracy co-winner that converts cleanly to CoreML).

## How

- **Fused CoreML model**: raw 16 kHz audio → 192-d in one graph. The kaldi-fbank frontend (framing + povey window + preemphasis + DC removal + DFT) is folded into a single `Conv1d`, so Swift does **zero DSP**. Parity vs the modelscope PyTorch reference: **min cosine 0.99974**.
- `DiarizationService.diarizeOffline` re-embeds each segment with an injected `SpeakerSegmentEmbedder`; everything downstream is dimension-agnostic.
- ERes2Net uses a **separate `speakers_eres2net.sqlite`** so 192-d and 256-d vectors never mix in a `SpeakerProfile` row.
- `build.sh`/`build-beta.sh` bundle the model best-effort (falls back to WeSpeaker when absent).

## Scope / honest caveats

- **Does not** change FluidAudio's internal "who spoke when" (WeSpeaker is baked in the binary). This swaps the **identity/matcher** embedding — exactly the "two people merged into one on Zoom" axis.
- Matcher thresholds stay WeSpeaker-tuned; recalibrate on labeled cross-call audio (`scripts/calibrate_eres2net_thresholds.py`, `scripts/run_zoom_eval.sh`).
- Model distribution is staged/bundled locally; shipping to users needs a download path (HF mirror) or committing the artifact.

## Test plan

- ✅ Swift golden-parity test (cosine ≥ 0.999) — `ERes2NetEmbedderParityTests`
[...]

PR #1236, opened 2026-06-20
fix(speaker): write-time contamination gate + decoupled link/merge floor

r3dbars:
Closes the two **write-path** speaker-accuracy holes from `docs/FIX_ROADMAP.md` (#6, #8 — see [#1227](https://github.com/r3dbars/transcripted/pull/1227)). Both ride the same matching surface, so they're one coherent PR, validated on the SpeakerEvalHarness.

> **Draft.** Do not merge / do not cut a release. Thresholds (0.78 link floor, 0.80/0.72 write bands) are deliberately conservative and should be confirmed on a full AMI/VoxCeleb sweep before landing — that needs network + CoreML models, which weren't available in the authoring environment, so validation here is on controlled-geometry synthetic corpora through the **real** TranscriptedCore code path.

## #6 — write-time voiceprint contamination gate
The EMA write-back fired on *every* successful match (`TranscriptionPipeline` :309/:731), blending the session mean into the persisted voiceprint at a fixed `alpha=0.15` (`SpeakerDatabase` :250-262) with **no quality/margin/confidence gate at write time**. A mature, named profile (`callCount>4`, no maturity bonus) accepting a 0.70 cosine match from an under-segmented cluster with no clear runner-up permanently, invisibly contaminated the stored fingerprint — compounding across meetings.

**Fix:** new `SpeakerWritePathPolicy.voiceprintBlendAlpha` — confident `0.15` / cautious `0.05` / frozen `0.0`, keyed on similarity **and** a margin-to-runner-up guard (reuses the certified ladder's `0.12`). Threaded through `addOrUpdateSpeaker` via an **additive** `SpeakerStore` requirement + forwarding default extension, so the existing 2-arg callers and the test double are untouched. Low-quality matches still **NAME** (display) but **freeze** the voiceprint (the appearance is still recorded). The write bar is decoupled from the 0.92 *display* auto-name bar so legitimate drift keeps adapting — no contamination-for-under-adaptation trade.

## #8 — decouple link/merge floor from per-utterance attach
The adaptive 0.70 floor both **attached** utterances *and* **fused** clusters that matched the same profile (`profileToSpeakers`, :334-347). Two distinct people each landing 0.70–0.92 against one profile (no runner-up within 0.05) were silently fused into one row; review then named the fused cluster and the mis-attribution survived.

**Fix:** new `SpeakerWritePathPolicy.crossClusterLinkFloor` (0.78) + `Transcription.planCrossClusterLinks`. Clusters that matched the same profile fuse **only when their mean embeddings are directly similar to each other** (the genuine over-segmentation signature) — strictly stricter than the 0.70 attach floor, but below the 0.88 within-meeting consolidation bar so legitimate de-fragmentation is preserved. A distinct voice is **spun off** to its own fresh profile so review can name it independently. Write-back is **deferred** until after this decision, so a spun-off voice never contaminates the shared profile (matching reads only the pre-meeting snapshot, so deferring changes no match decision).

Both the system and mic pipeline paths are updated identically; the decision is a pure, unit-tested helper.

## Coordination with in-flight work
- **Certified naming ladder (0.92 / 0.12):** raises only the auto-*name* (display) bar — it does **not** gate the write path, which is exactly this gap. This PR adds the write-path gates *next to* it without changing it; `voiceprintBlendAlpha` reuses the same `0.12` margin constant. No conflict.
- **ERes2Net #1175 (embedding-model swap):** does not conflict — these are pure decision functions over cosine geometry + a tunable link floor. If #1175 changes the embedding distribution, only the **threshold constants** in `SpeakerWritePathPolicy` need a re-sweep (all centralized in one file, harness-tunable). Flagging rather than fighting.

## Validation — SpeakerEvalHarness before/after
Added `--write-path-fixes off|on` to the harness `replay` so the same dumps run both ways, plus a network-free synthetic generator (`scripts/gen_synthetic_speaker_eval.py`) with controlled cosine geometry. `off` = legacy path; `on` = `SpeakerWritePathPolicy` gates (mirrors `TranscriptionPipeline`). `MATCH=0.70`, collar 0.25.

| corpus | metric | fixes **off** | fixes **on** |
|---|---|---|---|
| **normal** (no-regression control) | DER / fragmentation / false-merge / re-ID / profiles | 0.0 / 1.0 / 0 / 1.0 / 5 of 5 | **identical** |
| **twopeople** (#8) | false-merge | `1` → profile spans `[X,Y]` | **`0`** |
| | profiles at end (ideal 2) | `1` | **`2`** |
| | mean DER | `0.1875` | **`0.0`** |
| **contamination** (#6) | false-merge profile | spans `[A, D, T]` (3 people) | spans `[A, T]` — 3rd party **D rescued** |
| | profiles at end (ideal 3) | `1` | **`2`** |

- **normal** proves the gate does **not** cause under-adaptation / DER / fragmentation / re-ID regressions (the task's back-out criterion).
- **#8** is directly fixed: the intruder is no longer fused (false-merge 1→0, DER 0.1875→0).
- **#6**: the contaminated profile no longer swallows the unrelated 3rd party D. The residual `[A,T]` merge is the matcher legitimately matching A@0.71 ≥ the 0.70 floor (a threshold property, not a write-path one) — out of scope here.

> The synthetic corpora are a reproducible, network-free *probe* of the exact failure modes, **not** a substitute for an AMI sweep. Re-run on real corpora before landing.

Plus **15 new unit tests** (`SpeakerWritePathPolicyTests`, `SpeakerWriteBackGateTests`) — alpha bands, margin/ambiguity guard, frozen-alpha-leaves-voiceprint-byte-identical, two-people-one-profile spin-off, over-segmentation still fuses.

## Authoritative verification (all green)
`build-deps.sh --force` · `build.sh --no-open` · `run-tests.sh` **10152/10152** · `swift test` **511 (1 skipped, 0 failures)** · `run-integration-smoke.sh` · `swift build --package-path Tools/SpeakerEvalHarness -c release`.

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
@@ -346,21 +346,25 @@ extension Transcription {
             // off to its own profile so it stays independently nameable.
             let linkPlan = Self.planCrossClusterLinks(
                 matchedProfileBySpeaker: speakerMatchResults.mapValues { $0.persistentId },
+                matchSimilarityBySpeaker: speakerMatchResults.mapValues { $0.similarity },
                 meanBySpeaker: matchedMeanPerSpeaker,
                 segmentCountBySpeaker: embeddingsPerSpeaker.mapValues { $0.count }
             )
-            for (other, canonical) in linkPlan.remaps {
-                speakerIdRemap[other] = canonical
+            // Spin-off representatives: a distinct voice (or fragments of one) that only resembled the
+            // matched profile — give it its own identity so review names it separately. Removed from
+            // the matched set so it never writes back into the shared profile.
+            for rep in linkPlan.spinOffs {
+                let repMean = matchedMeanPerSpeaker[rep]
+                    ?? Self.computeMeanEmbedding(embeddingsPerSpeaker[rep] ?? [])
+                let spinoff = speakerDB.addOrUpdateSpeaker(embedding: repMean, existingId: nil)
+                speakerMatchResults.removeValue(forKey: rep)
+                speakerNewProfiles[rep] = spinoff.id
             }
-            for other in linkPlan.spinOffs {
-                // Distinct voice that only coincidentally matched the same profile — give it its own
-                // identity so review names it separately instead of inheriting the matched profile's
-                // name. Removed from the matched set so it never writes back into the shared profile.
-                let otherMean = matchedMeanPerSpeaker[other]
-                    ?? Self.computeMeanEmbedding(embeddingsPerSpeaker[other] ?? [])
-                let spinoff = speakerDB.addOrUpdateSpeaker(embedding: otherMean, existingId: nil)
+            // Group members fuse into their representative; drop their own match so only the
+            // representative writes back (keeper group → shared profile, spin-off group → new profile).
+            for (other, rep) in linkPlan.remaps {
+                speakerIdRemap[other] = rep
                 speakerMatchResults.removeValue(forKey: other)
-                speakerNewProfiles[other] = spinoff.id
             }
             if !linkPlan.remaps.isEmpty {
                 AppLogger.transcription.info("Merged speaker IDs with same DB profile", [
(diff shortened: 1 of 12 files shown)


r3dbars:
**Independent review (codex-reviewer) + follow-up — commit b79f0c86**

PASS gate (no P1s). Two P2 robustness gaps in `planCrossClusterLinks` were found and fixed:

1. **Keeper was picked by segment count, not match strength** — a long *distinct* voice that weakly matched a profile could keep the name while the real (shorter) returning speaker got spun off. Now the keeper is the highest-similarity-to-profile cluster (ties: more segments, then lower id).
2. **Spin-off compared each cluster only to the single largest** — with 3+ clusters on one profile, two mutually-similar fragments of a distinct voice were spun off *separately*. Now same-profile clusters are grouped by mutual cross-cluster similarity via **union-find**: the best-matching group keeps the identity, every other group fuses internally and gets one fresh profile.

`planCrossClusterLinks` is now `public` so the harness calls the **same** planner (no divergent reimplementation). Added unit coverage: best-match-keeps-identity-with-fewer-segments, and 3-fragment-two-mutual-fuse. Harness synthetic A/B numbers unchanged; full matrix (`build-deps --force` + `build.sh` + `run-tests` 10152 + `swift test` 512 + integration smoke) re-run green.

Reviewer found nothing on the deferred write-back restructure, spin-off routing through review, the additive protocol witness, or concurrency.

PR #1218, opened 2026-06-19
Add speaker review analytics funnel

r3dbars:
## Summary
- Add privacy-safe PostHog funnel events for speaker review prompt/action/completion from the speaker naming sheet.
- Track only coarse/bucketed properties: `participant_count_bucket`, `unresolved_count_bucket`, `review_reason`, `surface`, `action`, `result`, `meeting_age_bucket`, and `update_count_bucket`.
- Update analytics allowlists, sanitizer coverage, PostHog activation query helpers, health probe output, and observability docs.

## Privacy proof
- No speaker names, labels, transcript text, meeting titles, paths, audio refs, or raw counts are sent.
- Avoided `speaker_count_bucket` and `needs_name_count_bucket` because analytics sanitization drops keys containing sensitive `speaker` / `name` fragments.
- Added tests proving sensitive keys like `speaker_count_bucket`, `speaker_name`, and `transcript_title` are dropped while the coarse review funnel properties survive.

## Verification
- `python3 scripts/dev/check-build-source-lists.py`
- `python3 scripts/ops/posthog-activation-funnel.py --self-test`
- `bash -n run-tests.sh`
- `bash -n scripts/entrypoints/run-tests.sh`
- `bash -n scripts/ops/health-probe.sh`
- `python3 -m py_compile scripts/ops/posthog-activation-funnel.py`
- `bash run-tests.sh --filter AnalyticsEventPolicy` (4662/4662 passed)
- `bash run-tests.sh --filter SpeakerNamingPolicy` (25/25 passed)
- `bash run-tests.sh --filter AnalyticsPayloadSanitizer` (passed)
- `bash build-deps.sh --force` (passed; third-party warnings only)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` (10049/10049 passed)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash build.sh --no-open` (passed)
- `codex review --base origin/main` initially found a staging-only miss for the new helper file; after staging the helper and rerunning, final verdict: no actionable regressions.

PR #1216, opened 2026-06-19
Add local summary PostHog funnel

r3dbars:
## Summary
- replace local summary analytics with the A+ `meeting_summary_requested` / `meeting_summary_finished` funnel
- retain model-prepare lifecycle events with bucketed/enum-only properties
- update AnalyticsEventPolicy, privacy docs, health probes, nightly digest schema, and PostHog activation report helper

## Privacy proof
- allowed properties are limited to `artifact_age_bucket`, `duration_bucket`, `latency_bucket`, `model_state`, `provider`, `model_family`, `result`, `failure_kind`, and `surface`
- raw summary text, transcript text, titles, paths, local model paths, URLs, speaker names, and raw errors are not sent
- `Tests/AnalyticsEventPolicyTests.swift` pins the summary funnel allowlist and verifies unsafe keys are dropped

## Verification
- `bash build-deps.sh --force`
- `bash build.sh --no-open`
- `bash run-tests.sh`
- `bash run-tests.sh --filter AnalyticsEventPolicy`
- `bash run-tests.sh --filter NightlySecurityContract`
- `bash run-tests.sh --filter RepoCommandContract`
- `python3 -m py_compile scripts/ops/posthog-activation-funnel.py scripts/ops/generate-nightly-digest.py scripts/ops/nightly-security-check.py`
- `python3 scripts/ops/posthog-activation-funnel.py --self-test`
- `python3 scripts/ops/generate-nightly-digest.py --self-test`
- `bash -n scripts/ops/health-probe.sh`
- `bash scripts/dev/agent-preflight.sh`
- `/Users/redbars/.codex/skills/codex-review/scripts/codex-review --mode branch` clean on final rerun

## Codex review
- accepted and fixed: optional summary events no longer block the ordered activation funnel
- accepted and fixed: failed/cancelled summary finishes no longer count as first-value events
- final helper run: clean, no accepted/actionable findings reported

PR #1238, opened 2026-06-20
Stop telemetry PRs colliding: single-source the analytics allowlist to a union-mergeable .psv

r3dbars:
## Problem

Telemetry PRs were permanently conflicting. Every privacy-reviewed event PR appended to the **same multi-line, brace-balanced region** of `Sources/Observability/AnalyticsEventPolicy.swift` — the `allowedPolicies` dictionary — plus a lockstep `reviewedNonBucketProperties` set in the test. Two PRs adding different events collided pairwise at the shared dict anchor, so any one merge dirtied all the others.

`docs/analytics-taxonomy-merge.md` already diagnosed this and warned that you **cannot** naively `merge=union` the `.swift` (the trailing `]`/`)` align as shared context, so union merge silently produces unbalanced, non-compiling Swift). The durable fix it prescribed — and the same pattern that fixed the identical cascade in the sibling SteadyType repo — is to extract the data **out of** Swift into a union-mergeable data file.

## Approach chosen: extract the registry to a `.psv` + `merge=union` + normalizer (option 3)

This is the documented, least-invasive option that eliminates the pairwise conflict while preserving every guardrail:

- **`Resources/analytics-events.psv`** — the registry, one `event_name|prop,prop` line per event. Each line is independent, so it is safe to `merge=union`.
- **`AnalyticsEventPolicy` compiles that file at runtime** (app bundle in a shipped build; repo checkout under `swiftc`/CI). It **fails closed**: an unreadable registry yields an empty allowlist and `AnalyticsReporter` drops every event rather than forwarding an unreviewed payload off-device. The public API (`policy(forEvent:)`, `allEventNames`, `allPolicies`) is unchanged, so `AnalyticsReporter` and every existing assertion keep working.
- **`Resources/analytics-reviewed-properties.psv`** — the privacy guardrail allowlist (reviewed non-bucket properties) moved out of the test's hardcoded Swift set, also `merge=union`, so reviewing a property is a one-line append.
- **`*.psv merge=union`** in `.gitattributes`; **`scripts/ops/normalize-analytics-taxonomy.py`** re-sorts/dedupes after a union merge (`--check` is a hygiene gate).

### Why not the alternatives
- *Per-event self-registration in Swift / split Swift files*: still `.swift`, still risks the silent unbalanced-merge trap; and the build globs `Sources/**` + CI's `check-build-source-lists.py` would fight new generated/registered Swift files.
- *Build-time codegen*: would add a generated `.swift` to wire into `build.sh`, `run-tests.sh`, and the source-list contract — more moving parts, more invasive. Runtime-load needs none of that because `Resources/.` is already bundled.

## Guardrails preserved
All read the same data files now:
- docs↔source parity, forbidden-fragment scan, and reviewed-non-bucket review checks (now load the `.psv`s)
- **new** duplicate-event guard (catches a same-event union merge)
- nightly PostHog schema check (`nightly-security-check.py`) repointed at the registry via a new `analytics_event_registry` manifest path; `testNightlySecurityContract` still runs the strict checker and it still passes (and still fails the stale fixture). Analytics privacy coverage is actually *stronger* per-PR: the fast "blocks sensitive property names" suite scans every `.psv` property by substring (a superset of the nightly exact-key scan).

Generated `.psv` parity was verified to be byte-for-byte equivalent to the previous compiled table (73 events, all `.union(...)` compositions expanded), cross-checked against the docs allowlist.

## Verification — the cascade is broken

Reproduced in scratch branches (registry conflict markers counted):

| Scenario | Result |
|---|---|
| **Baseline** — two new events at the shared end-of-`allowedPolicies` anchor (old Swift) | ❌ conflict in `AnalyticsEventPolicy.swift` |
| **Fix** — the same two events as `.psv` lines | ✅ clean union merge |
| **#1216 ⊕ #1218 ported** — both PRs' real registry changes expressed as `.psv` lines (6 + 3 events, plus `model_family` / `review_reason` reviewed props and the privacy-doc bullets) | ✅ clean merge, 82 events, no duplicates, parity holds, all guardrails pass |

(A literal rebase of the *unmodified* #1216/#1218 still conflicts once against this branch where it restructured the Swift — that's the expected one-time port noted below, not a recurring pairwise cascade.)

## Next step

**Rebase the 8 dirty telemetry PRs onto this fix** (one-time port of each PR's events into `analytics-events.psv` + `analytics-reviewed-properties.psv` + the privacy-doc list), then send them through merge-room. After porting, they no longer collide with each other on the registry.

Out of scope (documented as follow-ups in `docs/analytics-taxonomy-merge.md`): the registry **mirrors** that still hand-merge — `scripts/ops/health-probe.sh` (single-line event string) and the per-line event tuples in `scripts/ops/posthog-activation-funnel.py` / `generate-nightly-digest.py`.

https://claude.ai/code/session_012wEuZwpteht4kAUCfE1ced

---
_Generated by [Claude Code](https://claude.ai/code/session_012wEuZwpteht4kAUCfE1ced)_


r3dbars:
Superseded by #1297, which rebuilds the union-mergeable analytics registry from latest main and preserves the newer telemetry rows that landed after this draft. Closing this dirty draft so the registry cleanup has one canonical PR.

PR #1241, opened 2026-06-20
Ad-hoc call detection: make browser calls fire reliably + add camera-on signal

r3dbars:
Two linked goals around ad-hoc meeting capture. Draft — not for release.

## PART A — diagnosis: why spontaneous browser calls felt invisible

The wiring and detector policy are **correct and were live-verified** (spec, 2026-06-14): a spontaneous Google Meet with no calendar invite *does* produce a "Call detected in your browser" prompt. So this was **not** a hard "never fires" bug in the lab. The real reasons it felt invisible, ranked with evidence:

1. **Untested CoreAudio delivery layer (primary).** The only low-latency trigger was a single `kAudioDevicePropertyDeviceIsRunningSomewhere` listener on the **default input device**. That edge fires only on a 0→1 transition of the device's *aggregate* running state, so a call that starts while the device is **already running** (another app held the mic, or our own warmup touched it) or on a **non-default device** never trips it — and detection collapses onto the **60s backstop poll**, which a short spontaneous call slips right through. Nothing in CI exercised `start()→edge→emit`, so this stayed invisible to the test gate.
2. **Discoverability (secondary).** Onboarding's calendar stage framed detection as calendar-only ("Want meeting reminders?", "look at your calendar"); nothing told users spontaneous browser calls are auto-detected. The only place that said so was a default-on Settings toggle users rarely visit.
3. **No "is this really a call?" confirmation.** Phase 2 (UDP hardening) was deferred, so there was no min-duration gate and the path had to stay conservative.

Ruled out with evidence: monitor not started/wired (it is, default-on, `onChange` before `start()`); browser provider ineligible (`micInputCandidates` does **not** gate on `supportsRuntimeOnlyPrompt`; `com.google.Chrome.helper` → `.googleMeet`); own-capture / enabled gates (false when idle).

### What was loosened / tightened
- **Backstop poll 60s → ~5s** (`MicActivityMonitor`). The process-object poll is device-agnostic, so it catches the already-running / non-default-device cases the edge misses, within seconds instead of a minute. Reads are cheap → high-value, low-risk.
- **`SustainedActivityConfirmer`** (shared pure gate): mic/camera activity must be held continuously ~3s before prompting, with a scheduled re-scan at the confirmation deadline so a real call still surfaces fast. This is the on-device min-duration sanity check (the lightweight alternative to the deferred UDP layer) that lets us poll aggressively **without** prompting on momentary mic access.
- **Discoverability:** onboarding now adds one additive line that spontaneous calls are auto-detected (kept minimal to avoid colliding with the in-flight, **currently doc-only** calendar pre-arm work — `docs/MEETING_CAPTURE_PROMPTING.md`); the Settings "Auto-detect calls" info copy now mentions the camera signal.

## PART B — camera-on detection (net-new complementary signal)

A camera turning on while we are not recording is strong evidence of a meeting (e.g. a camera-on, **mic-muted** Meet join the mic path misses).

- **API (researched, not guessed):** CoreMediaIO exposes **no** public per-process camera API. The one public, on-device, metadata-only signal is **`kCMIODevicePropertyDeviceIsRunningSomewhere`** per CMIO device — the exact camera analog of the mic device boolean. `CameraActivityMonitor` enumerates `kCMIOHardwarePropertyDevices` and watches that boolean. It **never opens a stream / reads a frame**.
- **Needs extra permission? → NO.** No camera TCC prompt, no `NSCameraUsageDescription`, no `com.apple.security.device.camera`. Confirmed: the built app's entitlements gained nothing (only the existing audio-input/audio-capture/calendars), and the build's launch smoke started the monitor with no permission hang. Same posture as the mic monitor; matches OverSight/Guard, which detect camera-on this exact way.
- **Attribution + de-dupe:** the camera boolean has no attribution, so it is attributed to a provider via the **frontmost call app** (`cameraCallProvider`): frontmost browser → `.googleMeet`, frontmost native conferencing app → that provider, **anything else (Photo Booth / QuickTime) → no prompt** (the non-call sanity check). The camera is only a *standalone* signal when **nothing holds the mic**; when the mic is active it already names the call, so a normal mic+camera video call raises **exactly one prompt** (`.micInput` wins) and the camera never retargets to a different frontmost app. New `.cameraInput` reason for analytics; reuses the `.runtimeApp` source so snooze/dismiss/backoff are unchanged. Gated behind the same `AutoCallDetectionPreferences` toggle and `isOwnCaptureActive`; started/stopped in lockstep with the mic monitor.

## Tests
`SustainedActivityConfirmerTests`, `CameraActivityMonitorTests`, and new heuristics / synthetic-evaluator / detector cases: camera attribution, Photo-Booth-quiet, mic+camera one-prompt de-dupe, **cross-provider non-override**, own-capture/disabled gates, plus the sustain-gate boundary/re-arm semantics.

## Verification (all green)
`build-deps.sh --force` + `build.sh --no-open` (launch smoke, no camera entitlement) + `run-tests.sh` (**10,203** pass) + `run-integration-smoke.sh`. Independently reviewed by Codex + a second reviewer — both PASS; the one P2 they raised (cross-provider camera misattribution) is **fixed** in this PR (camera is standalone only when the mic is idle) and covered by a regression test.

## Known tradeoffs (P3, intentional)
- Common-path mic latency goes from ~1s to ~3s by design (the sustain gate); a genuine call still prompts within a couple of seconds.
- The camera boolean can read "in use" for a stuck-on virtual camera (OBS); it only prompts if a call app is *also* frontmost, and cooldowns limit any nagging.

## Manual proof status
The mic browser path was live-verified previously (spec, 2026-06-14). The camera path is verified here by build (no TCC prompt, no entitlement) + unit tests end-to-end through the synthetic evaluator. A live GUI dogfood (join a real Meet camera-on/mic-muted and confirm the prompt; start a real spontaneous browser call and confirm the faster prompt) is the recommended next step via `run-live-capture-smoke.sh` / manual dogfood — not run in this session.


r3dbars:
Superseded by #1271, which merged the current meeting prompt call-detection slice on main. This branch is still conflict-dirty, so any remaining ideas here should come back as a fresh narrow PR after release rather than block release cleanup.

PR #1323, opened 2026-06-25
feat(mcp): recap/recent_context return summary, not raw greetings

r3dbars:
## What

From `docs/NEXT_WORK.md` #2 — cheap, high-value MCP fix.

The MCP `recap` tool previewed each meeting with the **first 15 raw dialogue lines** (`extractDialogueLines(...).prefix(15)`), which are almost always greetings, audio checks, and small-talk — not the substance an agent wants. `recent_context` had the same problem via the index's first-utterance preview.

Both now return the **structured local summary** (overview + decisions + action items) that the app already embeds in the meeting Markdown, falling back to raw dialogue lines only when no summary exists.

## How

- New `MeetingSummaryPreview` — a small, self-contained port of the app-side `RecentMeetingSummaryPreviewParser` (`Sources/UI/Shared/RecentCaptureScanners.swift`). The MCP server has no compile-time dependency on the app target, so the parser is reproduced here against the same on-disk shape (the `<!-- transcripted:local-summary -->` body block, with a `local_summary*` frontmatter fallback).
- `handleRecap` now builds its preview via a shared `meetingRecapPreview(...)` helper (summary → else dialogue).
- `handleRecentContext` overrides the index's first-utterance preview with the summary when one is present.

Diff is scoped to the recap/recent_context handlers + Models + the new files, to stay clear of the cross-meeting-tools thread also editing `ToolHandlers.swift`.

## Tests

`MeetingSummaryPreviewTests`:
- recap returns summary content (decisions/action items), **not** the raw "Good morning" greeting, when a summary is present
- recap falls back to dialogue lines when there's no summary
- frontmatter-only summaries are parsed

`swift test --package-path Tools/TranscriptedMCP` → 77 passing.


r3dbars:
Closing as superseded by #1331, which merged the combined ask-meeting-history path covering this draft's scope.

PR #1325, opened 2026-06-25
fix(dictation): make daily transcript appends crash-safe (atomic write)

r3dbars:
## Why

Dictation is the highest-frequency capture surface, and its day-file save was the **one capture write that wasn't atomic**. Everything else (meeting artifacts, journals, JSON state) already uses `Data.write(options: .atomic)` / `String.write(atomically: true)`. Dictation appends did not.

`DictationTranscriptWriter.appendSection` opened the existing `Dictations_YYYY-MM-DD.md` with a `FileHandle`, seeked to the end, and wrote the new section **in place**. A crash, power loss, or interruption part way through that write would leave a half-written section glued onto the file and corrupt **every earlier entry for that day**.

## What changed

- `Sources/Dictation/DictationTranscriptWriter.swift` — `appendSection` now reads the existing day file, concatenates the new section, and writes the whole file atomically via `String.write(atomically: true)` (write-temp-then-rename) — the same safe-write pattern the rest of the app uses. The prior content survives untouched until the new file is renamed into place, so a failed write can lose only the in-flight section, never the existing day. Removed the now-unused `fileEndsWithBlankLine` byte-probe helper (replaced by a `hasSuffix("\n\n")` check on the in-memory string).
- New-file creation (`String.write(atomically: true)`) and the delete-rebuild path (`DictationTranscriptStore`) were already atomic — this closes the last gap.

Markdown layout is unchanged (same `\n\n` separator and section shape), so the `TranscriptedCaptureKit` day-file parser needs no change.

## Test

Added a regression suite in `Tests/DictationTranscriptWriterTests.swift`: it saves a first entry, then simulates an **interrupted/failed append** by making the day folder read-only so the atomic temp write throws mid-write, and asserts the existing day file is **byte-for-byte intact** with the second (doomed) entry never leaking in.

## Verification

- `bash build.sh --no-open` ✅
- `bash run-tests.sh` ✅ (10157 passed, 0 failed)

### DIFF
--- a/Sources/Dictation/DictationTranscriptWriter.swift
+++ b/Sources/Dictation/DictationTranscriptWriter.swift
@@ -216,30 +216,17 @@ enum DictationTranscriptWriter {
         return size > 0
     }
 
+    // Append by rewriting the whole day file atomically (write-temp-then-rename) rather
+    // than seeking to the end of the existing file and writing in place. A crash or
+    // interruption part way through an in-place `FileHandle` write would leave a half-written
+    // section glued onto the day's data and corrupt every earlier entry in the file. With an
+    // atomic write the prior content survives untouched until the new file is renamed into
+    // place, so a failed write can lose the in-flight section but never the existing day.
     private static func appendSection(_ section: String, to url: URL) throws {
-        guard let handle = FileHandle(forWritingAtPath: url.path) else {
-            throw NSError(
-                domain: "DictationTranscriptWriter",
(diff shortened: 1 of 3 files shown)


r3dbars:
Superseded by clean replacement PR #1364. Current main already has the atomic append implementation; #1364 salvages only the missing failed-append regression test. Closing this dirty draft so it is not merged directly.

PR #1329, opened 2026-06-25
perf(home): read recent-meeting metadata from a SQLite index, not file re-parse

r3dbars:
## What

The Home capture list (`RecentMeetingsScanner.loadRecent`) re-parsed each **visible** meeting transcript on every refresh — full Markdown read + frontmatter parse + speaker-label regex + summary parse — which doesn't scale to a large library (NEXT_WORK #10).

This points the Home list at an index: a small SQLite cache (`RecentMeetingMetadataCache`) keyed by transcript path and validated by the transcript + summary-sidecar **modification time and size** (cheap `stat`s, no content reads). On a warm hit the row is rebuilt straight from the index and **no transcript content is read**; on a miss or when the on-disk file changed it parses exactly as before and repopulates the index.

## Why a new cache instead of an existing table

The Home row needs the summary preview, speaker-needs-review count, and audio-health hint — none of which the existing `StatsDatabase.recordings` table or the MCP `TranscriptIndex` (separate package, JSON-indexed, not app-linked) store. A pure "read the existing table" swap would drop fields and break the **behavior-identical** requirement. So this caches exactly the derived row, keyed by file identity, and falls back to the live parse whenever the file changes.

## Scope / honesty

- The scanner already only parsed the **top `limit`** rows (it breaks early), so this removes those per-row content reads/parses. The O(N) directory `stat` enumeration (needed for ordering + cache validation) is unchanged. A fully index-ordered `ORDER BY date DESC LIMIT n` query that skips the directory entirely would need an authoritative, save-path-synced index + migration — out of scope here and a bigger conflict surface.
- Retained-audio attachments are still resolved live each refresh (cheap dir probe, already invalidation-managed by `HomeCaptureRefreshObserver`).
- Degrades safely to the old cold-parse behavior if SQLite fails to open or a payload fails to encode/decode.

## Test

`Tests/RecentCaptureScannersTests.swift`:
- **Warm index serves without re-reading**: warm the index, `chmod 000` the transcript, prove the row still loads with identical fields (impossible if it re-read the file), plus a control showing the unreadable file yields **no** row when the index is cold.
- **Invalidation**: a changed transcript (new mtime) re-parses and reflects the new title.

## Verification

- `bash build.sh --no-open` ✅
- `bash run-tests.sh` ✅ (10167 passed)
- `bash run-e2e-smoke.sh` ✅ (added the new file to its swiftc source list)
- source-list validator ✅

Manual proof (Home feels fast with a big library) is for Justin.

## Conflict note

Touches `RecentCaptureScanners.swift` (data-source swap only) — kept deliberately narrow to coexist with the in-flight UI-polish PRs that also touch Home. No Home view changes.

### DIFF
--- a/Sources/UI/Shared/RecentCaptureScanners.swift
+++ b/Sources/UI/Shared/RecentCaptureScanners.swift
@@ -354,14 +354,20 @@ enum RecentMeetingsScanner {
     private static let excludedMarkdownFilenames: Set<String> = ["AGENT.md", "CLAUDE.md"]
     private static let summaryPreviewByteLimit = 64 * 1024
 
-    static func loadRecent(limit: Int = 3, directory: URL? = nil) -> [RecentMeetingItem] {
+    static func loadRecent(
+        limit: Int = 3,
+        directory: URL? = nil,
+        cache: RecentMeetingMetadataCache? = .shared
+    ) -> [RecentMeetingItem] {
         guard limit > 0 else { return [] }
 
         let dir = directory ?? MeetingStoragePaths.transcriptsFolder
         let fm = FileManager.default
         guard fm.fileExists(atPath: dir.path) else { return [] }
 
-        let keys: [URLResourceKey] = [.creationDateKey, .contentModificationDateKey, .isRegularFileKey]
+        let keys: [URLResourceKey] = [
+            .creationDateKey, .contentModificationDateKey, .isRegularFileKey, .fileSizeKey
+        ]
         let requestedKeys = Set(keys)
         guard let urls = try? fm.contentsOfDirectory(
             at: dir,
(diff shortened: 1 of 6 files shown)


r3dbars:
Held, not merged. I rebuilt the useful Home metadata-cache slice on current main at branch `codex/salvage-1329-home-index`, and targeted checks passed, but the required `bash build-deps.sh --force && bash build.sh --no-open` proof failed twice at the final LaunchServices launch-smoke step with `_LSOpenURLsWithCompletionHandler() failed with error -10810` after compile/signature verification succeeded. Leaving this dirty draft open/held until that launch-smoke failure is resolved or explicitly waived.

r3dbars:
Hold recommendation from fresh-main salvage pass (2026-07-03): do not merge this PR directly.

This branch is conflict-dirty and broad across app/audio/meeting/speech/tooling surfaces. I did not find a small, proven replacement slice safe enough to ship from it in this pass without widening risk or claiming missing manual/audio proof.

Current salvage outcome: only the deterministic Home performance-budget guardrail from #1367 was rebuilt separately on branch `codex/perf-security-audio-salvage` (`6ea7067f`), and even that replacement is held from PR open/merge until the local `bash build.sh --no-open` LaunchServices smoke failure (`-10810`) is resolved. Please keep this PR out of release/merge queues unless a future fresh-main rebuild narrows it to a small green slice with mapped proof.

r3dbars:
Closing as replaced by merged PR #1380, which rebuilt the useful Home recent-meeting metadata cache work on current main.

PR #1411, opened 2026-07-03
Add Dayflow timeline QA hardening contracts

r3dbars:
## Summary

- Add a synthetic timeline QA contract fast test with a fake capture provider and fake timeline DB fixture.
- Register the fast test manifest coverage and route timeline/test fixture/manual QA edits through source-list, build, fast-test, and privacy-sweep checks.
- Extend the privacy leak sweep with timeline-specific screen-derived leak classes and synthetic timeline artifact surfaces.
- Add a short manual QA checklist that keeps automated fixture proof separate from real Screen Recording/TCC proof.

## Verification

Passed:

- `python3 -m py_compile scripts/ops/privacy-leak-sweep.py`
- `python3 scripts/ops/privacy-leak-sweep.py --write-report build/privacy-leak-sweep-report.json`
- `python3 scripts/dev/check-build-source-lists.py`
- `bash run-tests.sh --filter TimelineQAContractTests` — 25/25 passed
- `bash run-tests.sh` — 12015/12015 passed, captured in `build/run-tests-full.log`
- `scripts/dev/agent-preflight.sh`
- `bash -n run-tests.sh`
- `bash -n scripts/entrypoints/run-tests.sh`
- `git diff --check`

Blocked on this host:

- `bash build.sh --no-open` initially failed because deps were missing/stale and requested `bash build-deps.sh --force`.
- `bash build-deps.sh --force` failed while rebuilding MLX/FluidAudio deps with `No space left on device`; `/System/Volumes/Data` had ~4.3 GB free afterward. This is a host disk-space blocker, not a test failure from the patch.

## Maestro lanes

- Codex: implemented, reviewed diff, ran source-list/privacy/fast-test/preflight checks, committed/pushed/opened draft PR.
- Local: scout only via `~/.codex/bin/maestro-delegate`; proof path `/Users/redbars/.codex/maestro/runs/20260703T215742Z-dayflow-phase10-test-scout-local`.
- Claude: skipped; `maestro-smoke` reported Claude CLI not logged in.
- Windows: skipped; `maestro-smoke` reported Windows worker not configured.


r3dbars:
COORD_REVIEW: YELLOW at head 3a08e84fa5436eec57d31b069b4a5cfd871b354a.

Reviewed against origin/main 6a1cb209aba392855d8ba8b2ac87fd6890ca6382. repo-hygiene SUCCESS, build-and-test IN_PROGRESS at review time, hardware-smokes skipped, git diff --check clean.

Do not mark ready/merge yet. Dayflow timeline QA hardening contracts are draft with CI still running.

Minimal path: wait for build-and-test, confirm CLEAN after ready-state recompute, then re-review exact head. If this is only contracts/tests and CI passes, it may be a GREEN candidate next.

r3dbars:
Flagging two issues found in `Tests/TimelineQAContractTests.swift` during review — both worth addressing before merge, but neither is a quick one-line fix so noting here rather than pushing a change.

**1. The "contract" tests don't exercise any real Timeline code (lines ~9-38, ~134-163)**

`FixtureTimelineCaptureProvider` and `FixtureTimelineDatabase` are brand-new private mock structs defined inside this test file. They just echo back whatever JSON fixture they're constructed with — they never call into `Sources/Timeline/**` or `Sources/UI/Timeline/**`.

I checked out this branch locally to confirm: the merge-base with the PR's stated base predates the Timeline directories existing at all in that tree. So as it stands, this "QA hardening for Timeline" suite currently compiles and passes in a state where the feature it claims to guard doesn't exist on this branch. A real bug in `ScreenCaptureEngine`'s TCC gating, or in `TimelineDatabase`'s overlap/ordering logic, would ship with this suite staying green, since the tests never touch that code.

Recommend: rebase onto current `main` so `Sources/Timeline` actually exists on this branch, then replace (or at minimum supplement) the fixture-based mocks with tests that exercise the real `ScreenCaptureEngine` / `TimelineDatabase` code paths. Fixture-only contract tests can still be useful for fast iteration, but they shouldn't be the only coverage claiming to harden the real Timeline feature.

**2. `timelineOverlapCount` has a real logic gap (lines ~165-174)**

After sorting by start time, it only compares each card to its immediately-preceding neighbor's `endTS` — not a running max end time. That means a card fully nested inside an earlier, longer card gets missed. Example: cards `A(0,100)`, `B(1,2)`, `C(50,60)`, sorted as A, B, C — the overlap between C and A is never checked, because C is only compared against B (its immediate predecessor), not against A.

This doesn't fail today only because the current fixture data happens to be contiguous/non-overlapping in a way that doesn't trigger the gap. It should track a running max end time across all prior cards (not just the immediate predecessor) so nested-overlap cases are actually caught.

Neither of these seems like a small tweak given the scope of the current test suite, so surfacing for the author/maintainer to decide how to re-scope rather than attempting a fix in-line.

---

PR #1487, opened 2026-07-07
feat(speaker): negative exemplars — learn "this voice is NOT this person" from corrections

r3dbars:
## Problem

When a user corrects a wrongly auto-assigned speaker, the system only **freezes** the wrongly-suggested profile (dispute count) and teaches the correct one. The rejected embedding itself is discarded — every correction is thrown-away training signal. Worse, the dispute freeze is **cleared on repair** (`resetDisputeCount`), so once the profile is un-frozen the *same wrong voice* can re-match it. The correction taught nothing durable.

## What this does

Records the rejected embedding as a **negative exemplar** ("explicitly not this person") against the wrongly-suggested profile. At match time, an embedding that resembles one of a profile's negative exemplars *at least as much as it resembles the profile itself* is **vetoed** — the profile is dropped from candidacy (best **and** runner-up), even if it clears the similarity floor. Negative exemplars persist across dispute reset, so the learning outlives the freeze.

This is the mirror image of the positive fingerprint: positive exemplars = "sounds like", negative exemplars = "explicitly not". They compose cleanly with the in-flight multi-exemplar-voiceprint work — orthogonal signals, both consumed in the same matcher.

## Design (additive, backward-compatible, invisible)

An **additional** veto layered on top of the existing floors, maturity bonus, ambiguity separation, and dispute freeze. It can only make matching *more* conservative, never less:

- `SpeakerNegativeExemplarPolicy` — pure veto decision. `vetoFloor = 0.80` (must strongly resemble an *explicitly rejected* sample; coincidental similarity never fires) and `competitiveMargin = 0.03` (a voice clearly closer to the fingerprint than to any rejected sample is trusted, so a **genuine returning voice is never vetoed**).
- `SpeakerNegativeExemplarStore` — append-only per-profile table in `speakers.sqlite`, embeddings only (no names / transcript text / audio, matching the existing privacy posture), newest-32 retained per profile.
- Recorded on the `.corrected` review path (`SpeakerNamingCoordinator`), where both the rejected session embedding and the wrongly-matched snapshot are already in hand alongside the existing `restoreProfile` + `incrementDisputeCount`.
- Consumed by **both** matchers: `matchAgainstProfiles` (in-memory snapshot, used by the pipeline + eval harness — gained an optional `negativeExemplarsByProfile` param that defaults empty) and `SpeakerDatabase.matchSpeaker` (persistent; fetches in-queue to avoid `queue.sync` re-entrancy).

Profiles with no negative exemplars — and callers that pass no map — match **exactly** as before.

## Validation

- **Full package suite: 697 tests, 0 failures** (12 pre-existing skips).
- The runnable deterministic eval/simulation harness — `SpeakerNamingSimulationRunnerTests` (the deep speaker-naming regression suite) — **still passes at 1.0 exact-label accuracy with zero confusion/false-merge/false-split indicators**, confirming no regression from the matching change. Its runner was updated to mirror the production correction path so the harness exercises negative exemplars end to end.
- The heavy corpus sweep (`scripts/run_speaker_eval.sh`, AMI/ICSI/VoxConverse) needs downloaded corpora not present in this environment, so it was **not** run — flagging explicitly.

New focused tests:
- Policy unit tests (veto floor, competitive margin, boundaries, dimension-mismatch, empty).
- Store round-trip + newest-N retention.
- In-memory matcher: negative exemplar excludes the wrong profile **but keeps the correct match**; a genuine voice still matches despite the exemplar.
- Persistent matcher: vetoes an **un-disputed** profile (proving the signal is independent of, and outlives, the dispute freeze).
- **End-to-end through the real correction path** via the simulation harness: a correction records the exemplar, it survives a simulated repair (`resetDisputeCount`), and the returning voice then matches the corrected-to person instead of the vetoed one.

## Scope boundary

Recording is scoped to `.corrected` (the canonical "wrong person" verdict). Collapse-to-you and discard also increment dispute count but are different intents ("that's me" / "don't save this sample"); recording negatives there is a deliberate follow-up, not this PR.

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipeline.swift
@@ -153,6 +153,8 @@ extension Transcription {
             // consolidation (collapses one over-segmented voice so the user names
             // each person once), and DB-informed split still run.
             let existingProfiles = speakerDB.allSpeakers()
+            // Rejected-sample vetoes for matching (empty until a correction records one).
+            let negativeExemplarsByProfile = speakerDB.negativeExemplarsByProfile()
             let speakerThresholds = diarization.activeSpeakerThresholds
             let speakerSegments = EmbeddingClusterer.postProcess(
                 segments: rawSegments,
@@ -317,7 +319,7 @@ extension Transcription {
                 // so a cluster that turns out to be a distinct voice (spun off) never blends into
                 // the shared profile. Matching reads only the `existingProfiles` snapshot, so
                 // deferring the blend cannot change any match decision.
-                if let matchResult = Self.matchAgainstProfiles(meanEmbedding, profiles: existingProfiles, threshold: adaptiveThreshold) {
+                if let matchResult = Self.matchAgainstProfiles(meanEmbedding, profiles: existingProfiles, threshold: adaptiveThreshold, negativeExemplarsByProfile: negativeExemplarsByProfile) {
                     speakerMatchResults[speakerId] = (matchResult.profileId, matchResult.similarity, matchResult.secondBestSimilarity)
                     let matchedProfile = existingProfiles.first(where: { $0.id == matchResult.profileId })
                     AppLogger.transcription.info("Speaker matched DB profile", [
@@ -899,6 +901,9 @@ extension Transcription {
             .filter { !ghostSpeakerIdSet.contains($0.key) }
             .reduce(into: [:]) { $0[$1.key] = Self.computeMeanEmbedding($1.value) }
 
+        // Rejected-sample vetoes for matching (empty until a correction records one).
+        let negativeExemplarsByProfile = speakerDB.negativeExemplarsByProfile()
+
         for (speakerId, embeddings) in embeddingsPerSpeaker {
             let meanEmbedding = Self.computeMeanEmbedding(embeddings)
             let isGhost = ghostSpeakerIdSet.contains(speakerId)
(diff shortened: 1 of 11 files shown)


r3dbars:
## Independent adversarial review + fixes

Ran an independent review of the diff (Codex was quota-blocked; used a separate in-conversation reviewer). Result: **no blocking bugs** — deadlock-clean (both matchers call the in-queue `*Impl` variants, never re-entrant `queue.sync`) and backward-compat provable (empty map ⇒ veto branch never executes ⇒ byte-for-byte prior behavior). SQL is parameter-bound, the prune query correctly keeps newest-N, no statement leaks.

One substantive finding (P2) and two P3s, all addressed:

**P2 — veto math could drop a genuine returning voice.** The original `competitiveMargin = 0.03` handicap (`negSim ≥ posSim − 0.03`) created a window where a legitimate owner whose voice was *closer to their own fingerprint than to the confuser* (`negSim < posSim`, e.g. posSim 0.82 / negSim 0.80) would still be vetoed and mint a duplicate profile — in exactly the confusable population this targets. **Fixed:** dropped the handicap. Veto now requires `negSim ≥ vetoFloor (0.80)` **and** `negSim ≥ posSim`. The target case (same wrong voice returns) has `negSim ≈ 1.0 ≫ posSim`, so it still fires robustly; a genuine owner (closer to their own fingerprint) is never vetoed. Fixing this also exposed an unrealistic degenerate geometry in the e2e test (profile, exemplar, and query all at the same angle → `posSim ≈ negSim ≈ 1.0`, float-noise-decided); rewrote it so the confuser voice is offset from the profile fingerprint, which is both realistic and robust.

**P3a — same-id positive+negative write:** confirmed **unreachable**. When a `matchedProfileSnapshot` exists, the session's `persistentSpeakerId == matchedProfile.id`, and the corrected-to target is resolved via `exactNamedTarget(excluding: persistentSpeakerId)`, so the target can never be the matched profile. Added a clarifying comment documenting the invariant.

**P3b — discard/collapse don't record negatives:** deliberate scope choice (discard = "don't save this sample", not "different known person"). Added a one-line comment at the discard path making the boundary explicit.

Re-validated: **full suite 698 tests, 0 failures**; deterministic speaker-naming regression suite still 1.0 exact accuracy.

PR #1620, opened 2026-08-04
Share the save-path rule across app and tools

r3dbars:
## Summary

Fixes the class of cross-tool contract drift that commit 649ec7c7 patched symptom-by-symptom, across four findings:

**1. Save-path safety predicate (3 hand copies → 1 shared file)**

`TranscriptedStoragePaths.isSafeCaptureLibraryURL`, `RecordingValidator.validateSavePath`, and `CaptureLibraryResolver.validatedConfiguredDirectory` each had their own copy of the "is this path safe to use as a capture-library root" rule, with a "keep this rule in lockstep with the other two" comment on each.

Extracted the rule into one dependency-free, pure-Foundation file: `Sources/Support/CaptureLibraryPathSafety.swift`. Since the three call sites live in three build units that compile in total isolation (app target via raw `swiftc`, `TranscriptedCore` via SPM, `TranscriptedCaptureKit` via SPM) and can't import across that boundary, the file is vendored into the other two via real git symlinks:
- `Sources/TranscriptedCore/Services/CaptureLibraryPathSafety.swift`
- `Tools/TranscriptedCaptureKit/Sources/TranscriptedCaptureKit/CaptureLibraryPathSafety.swift`

Both symlinks resolve correctly for `swift build`/`swift test` (which operate on the live checkout). `build-deps.sh` ditto-copies only `Sources/TranscriptedCore` into an isolated scratch tree, which would leave the Core-side symlink dangling — that script now materializes the real file there right after the copy (see the comment next to that `cp` call). All three call sites now delegate to `CaptureLibraryPathSafety.evaluate(...)`; the "keep in lockstep" comments are gone.

Since `-type f` in the build scripts' staleness-digest functions skips symlinks, `Sources/Support/CaptureLibraryPathSafety.swift` is listed explicitly in `build-deps.sh`, `build.sh`, `build-beta.sh`, and `run-integration-smoke.sh`'s digest/mtime functions so an edit to only that file is still detected as invalidating the built archive.

**2. Frontmatter parsers — already fixed, verified only**

Commit `eebbcd7b` (already on `main`) already solved this: a shared `Tests/Fixtures/frontmatter-corpus/` golden corpus plus a parity test suite in each of the three independent parsers (`TranscriptFrontmatter`, `CaptureMarkdownParser`, MCP's `frontmatterBlock`). No changes made here — ran all three suites as part of verification and they're green.

**3. mcp-directories.json (writer/reader round-trip pin)**

`TranscriptedMCPDirectoriesManifest` (app writer) and `CaptureDirectoryManifest` (TranscriptedCaptureKit reader) are independently declared `Codable` structs in separate build units with no compiler-enforced link, and the existing resolver test only decoded a hand-written JSON literal shaped to match the reader's own expectations — it never touched the real writer.

Added `Tests/Fixtures/mcp-directories-manifest/golden.json`, generated from the real app writer with fixed inputs. `Tests/TranscriptedStoragePathsTests.swift` now writes a manifest with the real writer and asserts its JSON key set and decoded fields match the fixture exactly. `CaptureLibraryResolverTests.testResolveDecodesRealAppWrittenManifestFixture` stages that same fixture as a temp home's `mcp-directories.json` and confirms `CaptureLibraryResolver.resolve()` reads it correctly end-to-end — so a field rename/add/remove on either side now fails a test instead of drifting silently.

**4. TranscriptedQA's 4th copy of the legacy Draft/legacy-shared layout**

`QADataDirectories.resolve`'s `inferBaseLayout` hardcoded its own `draftRoot`/`legacySharedRoot` paths, independent of `CaptureLibraryResolver`'s legacy-fallback list (the thing `a5a766cc`/`b2b54268` had to fix before). Since TranscriptedQA already depends on `TranscriptedCaptureKit`, deduplicated instead of just pinning: added `CaptureLibraryResolver.legacyCaptureDirectories(homeDirectory:)`, a small public accessor exposing the same legacy roots `resolve()` already falls back to (and `resolve()` itself now calls it too, removing the internal duplication as a bonus). `QADataDirectories.resolve` now derives its legacy layout from that instead of re-declaring it. Added `LegacyCaptureDirectoriesContractTests` proving an explicit `--path` at either resolver-vended legacy root still resolves through QA's inference to the matching sibling directories.

## Notable side finding (not fixed here)

Fixing `build-deps.sh`'s staleness-digest function to include the new shared file surfaced that `build.sh`, `build-beta.sh`, and `run-integration-smoke.sh` each hand-copy the identical `dependency_input_paths`/`dependency_input_listing` function — the same "N hand copies of one rule" pattern this PR fixes for app code, just in bash. All four copies were updated to include the new file consistently (verified `build.sh` catches a stale digest correctly), but deduplicating them into one sourced lib file is out of scope here. Flagged as a follow-up.

## Verification

All run from a clean worktree with prebuilt deps copied in:

- [x] `bash build-deps.sh --force` — succeeded (one transient "disk full" retry from a concurrent build elsewhere on the machine, auto-retried per the script's existing retry logic, not related to this change)
- [x] `bash build.sh --no-open` — succeeded
- [x] `bash run-tests.sh` — 12598 tests, 0 failed
- [x] `swift test` (root `TranscriptedCore` package) — 872 tests, 0 failed
- [x] `swift test --package-path Tools/TranscriptedCaptureKit` — 53 tests, 0 failed
- [x] `swift test --package-path Tools/TranscriptedCLI` — 78 tests, 0 failed
- [x] `swift test --package-path Tools/TranscriptedMCP` — 191 tests, 0 failed
- [x] `swift test --package-path Tools/TranscriptedQA` — 58 tests, 0 failed
- [x] `python3 scripts/dev/check-build-source-lists.py` — passed
- [x] `bash run-e2e-smoke.sh` — passed

## Concerns / notes for reviewers

- The predicate now explicitly rejects the literal root path `/` in all three call sites (previously only `CaptureLibraryResolver.validatedConfiguredDirectory` did this; `TranscriptedStoragePaths`/`RecordingValidator` would have technically accepted `/` as "safe" before falling over on the subsequent directory-creation/write-permission checks). This is a small behavior tightening, not a regression — no test exercised the old gap.
- Symlinks are a real vendoring mechanism here (confirmed empirically: SPM auto-discovers and compiles symlinked sources; `ditto` preserves relative symlinks as-is, which is why `build-deps.sh` needed the explicit materialization step after its copy).


r3dbars:
Addressed the P2 symlink-portability finding from review.

**Change:** Replaced both `CaptureLibraryPathSafety.swift` git symlinks with real, byte-identical files (confirmed via matching git blob hashes: `84eee389...` in both `Sources/TranscriptedCore/Services/` and `Tools/TranscriptedCaptureKit/Sources/TranscriptedCaptureKit/`). On a checkout with `core.symlinks=false`, git would have materialized those symlinks as plain text files containing the literal relative link target, breaking `swift test`, Xcode/SPM, and the CaptureKit package build.

- Updated the shared header comment (present identically in all three copies) to describe the "synced copy — edit all three together" contract, replacing the symlink-vendoring explanation.
- Reverted the `build-deps.sh` dangling-symlink materialization workaround (`rm -f` + `cp` after the `ditto` copy) — no longer needed since the Core-side copy is now a real file the existing `find "Sources/TranscriptedCore" -type f` staleness check already picks up correctly.
- Reverted the digest-list additions I'd made to `build-deps.sh`, `build.sh`, `build-beta.sh`, and `run-integration-smoke.sh` for the same reason — all four scripts are now byte-identical to `main` again.
- Added `Tests/CaptureLibraryPathSafetySyncTests.swift` (`testCaptureLibraryPathSafetySync`, following the root fast-test naming convention, using the existing `repoFixtureURL` helper) as the enforcement mechanism: it reads all three copies from disk and fails, naming the divergent paths, if any one of them drifts from the others by even a byte.

**Verification:**
- `swift test` (root `TranscriptedCore` package, compiles Core from source directly — not via the prebuilt archive) — 872 passed
- `swift test --package-path Tools/TranscriptedCaptureKit` — 53 passed
- `swift test --package-path Tools/TranscriptedQA` — 58 passed
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12600 passed (includes the new sync test)
- `python3 scripts/dev/check-build-source-lists.py` — passed

**Not verified this round:** `bash build-deps.sh --force`, `bash build.sh --no-open`, and `bash run-integration-smoke.sh`. `build-deps.sh --force` failed 8 consecutive times over about an hour with "No space left on device" mid-compile, at varying progress points — disk on this box fluctuated between roughly 350Mi and 1.6Gi available, consistent with contention from other concurrent builds on a shared machine rather than anything caused by this change (the identical command succeeded earlier in this session against the old symlinked version). `build.sh`/`run-integration-smoke.sh` both need a fresh archive to pick up the new file bytes (the digest changed even though the compiled logic is unchanged), so they're blocked on the same thing. The `swift test` run above does give real compile coverage of the new Core-side file, since SPM builds TranscriptedCore from source rather than linking the archive — I'm fairly confident `build-deps.sh`/`build.sh` will pass cleanly once disk pressure clears, but that's not yet independently confirmed. Flagged to the coordinating session; will retry when resources free up.

r3dbars:
Disk pressure on the shared build machine has cleared — completed the remaining verification from the previous comment:

- `bash build-deps.sh --force` — succeeded (`libDraftDeps.a` rebuilt with the new real-file `CaptureLibraryPathSafety.swift`; symbol validation: `libDraftDeps.a` defines 9655 TranscriptedCore symbols, `libExternalDeps.a` defines 0, as expected)
- `bash build.sh --no-open` — succeeded
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-integration-smoke.sh` — passed

Combined with the previous comment's results (`swift test` root package 872 passed, `swift test --package-path Tools/TranscriptedCaptureKit` 53 passed, `swift test --package-path Tools/TranscriptedQA` 58 passed, `bash run-tests.sh` 12600 passed including the new sync test, `check-build-source-lists.py` passed), the full verification matrix for this branch is now green. No further changes pending.

PR #1622, opened 2026-08-04
Unify speaker matching on the pipeline algorithm

r3dbars:
## Summary

`SpeakerDatabase.matchSpeaker` (the persistent-store matcher, `Sources/TranscriptedCore/Speaker/SpeakerEmbeddingMatcher.swift`) had a hand-copied cosine-similarity loop that had drifted from `Transcription.matchAgainstProfiles` (`Sources/TranscriptedCore/Speaker/SpeakerMatchingService.swift`) — the in-memory-snapshot matcher the **live** `TranscriptionPipeline` actually calls (`TranscriptionPipeline.swift:334`, `:946`). The stale copy was missing:

- the per-profile **maturity bonus** (+0.08 similarity required for `callCount <= 2`, +0.04 for `callCount <= 4`)
- the **ambiguity separation check** (reject a match if the runner-up is within 0.05 similarity)
- the **disputed-profile filter** (`profile.disputeCount == 0`)

The doc comment on `matchAgainstProfiles` claimed "Same logic as `SpeakerDatabase.matchSpeaker`" — no longer true before this change.

## Fix

Deleted the duplicate loop in `matchSpeakerImpl` and made it a thin wrapper: snapshot `allSpeakersImpl()` + `allNegativeExemplarsImpl()` under the existing `queue.sync`, delegate to `Transcription.matchAgainstProfiles`, then resolve the returned `profileId` back to a `SpeakerProfile` from that same snapshot. One matching algorithm now backs both the persistent-store path and the in-memory-snapshot path.

This intentionally **aligns** every `matchSpeaker` caller onto the pipeline's algorithm (maturity bonus + separation check + dispute filter) — that's the point of the change, not a side effect.

## Callers whose behavior changes

Grepped every call site of `matchSpeaker`/`matchSpeakerImpl` across `Sources/`, `Tools/`, `Tests/`:

- **`Tools/SpeakerEvalHarness/Sources/speaker-eval-harness/main.swift`**
  - Line 211 (`!writePathFixes` / "legacy" sweep mode): now applies the maturity bonus, separation check, and dispute filter where it previously didn't. This harness already has a `writePathFixes` mode (line 227) that calls `matchAgainstProfiles` directly — the "legacy" mode's `db.matchSpeaker` call now scores consistently with it instead of diverging.
  - Line 267 (final cluster-resolution pass, used by both harness modes): same alignment — now dispute-filters and applies maturity/separation gating on the post-merge re-match.
  - This is opt-in eval tooling (used for threshold sweeps / accuracy reports), not shipped production code, but it's worth knowing the harness's baseline numbers may shift on re-run.
- **`Tests/TranscriptedCoreTests/SpeakerTests/`** — `SpeakerEmbeddingMatcherTests.swift`, `SpeakerMultiExemplarMatchingTests.swift`, `SpeakerDBDimensionIsolationTests.swift`, `SpeakerNegativeExemplarTests.swift` all call `db.matchSpeaker` directly. All pass unchanged except one boundary test (see below).
- **Production runtime path is unaffected**: `TranscriptionPipeline.swift` already called `matchAgainstProfiles` directly, not `matchSpeaker` — so live meeting transcription behavior does not change at all. The only production-adjacent caller of `SpeakerDatabase.matchSpeaker` I could find via grep is the eval harness above; app-side (`Sources/Meeting/`, `Sources/Speech/`) does not call it.

## Test changes

- `SpeakerEmbeddingMatcherTests.testMatchSpeakerThresholdBoundaryIsInclusive` inserted a single-call (callCount 1) profile and asserted an exact-boundary match at `similarity == threshold == 1.0`. With the maturity bonus now applied through delegation, a callCount-1 profile needs `similarity >= threshold + 0.08`, so this would've started failing. Fixed by maturing the profile past callCount 4 first (bonus becomes 0) so the test still isolates the raw threshold-boundary behavior it's named for.
- Added `testMatchSpeakerAppliesMaturityBonusForNewProfile` to directly document/lock in the new alignment: an immature (callCount 1) profile at cosine ~0.65 against threshold 0.6 now correctly returns nil because the effective threshold is 0.68.

## Verification

All run from a clean worktree with prebuilt deps rebuilt for this change (`bash build-deps.sh --force`, required since `Sources/TranscriptedCore/**` changed):

1. `bash build-deps.sh --force` — succeeded (fresh `libDraftDeps.a`/`libExternalDeps.a`)
2. `bash build.sh --no-open` — **Build complete!**
3. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — **12596 tests, 12596 passed, 0 failed**
4. `bash run-integration-smoke.sh` — core smoke, wake smoke, and `MicRecordingFileMergerTests` all passed
5. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift test` (full suite) — **873 tests, 0 failures** (13 skipped, pre-existing/platform-gated)
6. `swift build --package-path Tools/SpeakerEvalHarness` — **Build complete!** (confirms the harness, which links `TranscriptedCore` from source, compiles against the delegated implementation)

Scoped `swift test --filter '^SpeakerTests\.'` also run standalone during iteration: 264 tests, 0 failures.

No drive-by refactors — diff is limited to the two Speaker files plus the one test file needed to keep coverage accurate.

### DIFF
--- a/Sources/TranscriptedCore/Speaker/SpeakerEmbeddingMatcher.swift
+++ b/Sources/TranscriptedCore/Speaker/SpeakerEmbeddingMatcher.swift
@@ -5,7 +5,10 @@ import Foundation
 @available(macOS 14.0, *)
 extension SpeakerDatabase {
 
-    /// Match an embedding against all stored speakers using cosine similarity.
+    /// Match an embedding against all stored speakers.
+    /// Delegates to `Transcription.matchAgainstProfiles` — the same matcher the live
+    /// transcription pipeline uses — so persistent-store callers (this) and in-memory-snapshot
+    /// callers (the pipeline, the offline eval harness) never drift onto two different algorithms.
     /// Returns the best match above threshold with similarity score, or nil for a new speaker.
     public func matchSpeaker(embedding: [Float], threshold: Double = 0.6) -> SpeakerMatchResult? {
         return queue.sync {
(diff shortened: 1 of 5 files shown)


r3dbars:
Addressed the two P2s from Codex review (tests + docs only, no `Sources/` changes):

**1. Pinned the maturity-bonus constants** (`Tests/TranscriptedCoreTests/SpeakerTests/SpeakerEmbeddingMatcherTests.swift`). Replaced the single loosely-bounded test with four tight boundary tests:
- `testMatchSpeakerMaturityBonusTier1AcceptsAtEffectiveThreshold` / `...RejectsJustBelowEffectiveThreshold` — pins the callCount<=2 tier at exactly +0.08 (similarity 0.68 accepted, 0.66 rejected, against base threshold 0.6).
- `testMatchSpeakerMaturityBonusTier2AcceptsAtEffectiveThreshold` / `...RejectsJustBelowEffectiveThreshold` — pins the callCount 3...4 tier at exactly +0.04 (similarity 0.64 accepted, 0.62 rejected), matured to callCount 3 by repeating the same embedding direction across writes (keeps the blended average — and therefore the score — unchanged while only callCount advances).

A change to either constant, or a collapse of the two tiers into one, now fails a test instead of silently passing.

**2. Flagged the stale eval baselines.** Added a provenance blockquote at the top of both `Tools/SpeakerEvalHarness/BASELINE_REPORT.md` and `Tools/SpeakerEvalHarness/SCALEUP_REPORT.md`: results were generated against the pre-unification `SpeakerDatabase.matchSpeaker` loop (the harness's "legacy"/`!writePathFixes` sweep mode this data comes from), which now applies the pipeline's maturity gate via this PR's delegation. Repro commands and profile-count/threshold claims are called out as not expected to reproduce exactly. Explicitly did **not** regenerate the baselines — flagged as separate follow-up work.

**Verification** (scoped to SPM tests since only tests + docs changed):
- `swift test --filter '^SpeakerTests\.'` — 267 tests, 0 failures (was 264: net +3 from removing 1 loose test and adding 4 tight ones)
- `swift test` (full suite) — 876 tests, 0 failures, 13 skipped (pre-existing/platform-gated)

Pushed to the same branch (`claude/w0-speaker-matching-dedup`, commit `feb6b1b4`). Still open, not merged.

PR #1623, opened 2026-08-04
Share smoke source lists and deps staleness checks

r3dbars:
## Summary

Pure behavior-preserving build-infrastructure dedup, per audit findings on drifting hand-maintained swiftc source lists / deps-staleness checks (16 prior commits in this repo's history have been titled around fixing a \"source list\").

**1. Shared smoke source lists** — `scripts/entrypoints/lib/shared-smoke-sources.sh` (new)

`run-tests.sh` (APP_SOURCES, 185 files), `run-e2e-smoke.sh` (SWIFT_SOURCES, 48 files), and `run-slow-pasteback-smoke.sh` (SWIFT_SOURCES, 3 files) each hand-list the raw-swiftc source files they compile — no SPM target, so no automatic dependency discovery, and deliberately not full `find`-based auto-discovery either (that would balloon fast-test/smoke compile times). Computed the actual overlap byte-for-byte before touching anything:

- `run-tests.sh` ∩ `run-e2e-smoke.sh` = 26 files → `SHARED_TEST_STORAGE_SOURCES`
- `run-tests.sh` ∩ `run-slow-pasteback-smoke.sh` = 2 files → `SHARED_PASTEBACK_SUPPORT_SOURCES`
- `run-e2e-smoke.sh` ∩ `run-slow-pasteback-smoke.sh` = 0 files (no third array needed)

All three scripts now source `lib/shared-smoke-sources.sh` and keep only their own delta appended after the shared arrays.

**2. Shared deps-staleness-check functions** — `scripts/entrypoints/lib/deps-staleness.sh` (new)

`build-deps.sh`, `build.sh`, and `run-integration-smoke.sh` carried six functions (`dependency_input_paths`, `dependency_input_listing`, `dependency_input_digest`, `newest_dependency_input`, `deps_build_stamp_info`, `deps_build_stamp_digest`) copy-pasted byte-for-byte (diffed all three pairs before unifying — confirmed identical). Extracted into `lib/deps-staleness.sh`, sourced by all three.

`build-beta.sh` was also checked and **does diverge** — it only has 3 of the 6 functions, inlines `dependency_input_paths` directly into `dependency_input_listing` without a `sort`, and never does the sha256-digest staleness check (mtime-only). This is a real, pre-existing behavior difference, not a copy/paste accident, so **left untouched** rather than folded in — aligning it would add a stronger staleness check to beta builds, which is a behavior change out of scope for this refactor. Documented in place with a comment explaining why.

**3. STRETCH item skipped** — the `-I`/`-F` flag lockstep between `lib/swiftc-app-args.sh` (bash, auto-discovers deps-modules subdirs via `find`) and the three SPM manifests (`Package.swift`, `Tools/TranscriptedCLI/Package.swift`, `Tools/SpeakerEvalHarness/Package.swift`, which each hard-code `FastClusterWrapper`/`MachTaskSelfWrapper`/`yyjson`) turned out gnarlier than a same-day fix: `deps-modules/` on this machine also contains `Cmlx` and `_NumericsShims` directories that none of the three SPM manifests list at all, and it's unclear from a fresh read whether that's latent drift or intentional (those two may already resolve through another mechanism). Restructuring or adding a strict contract test both risked either a real SPM behavior change or an immediate, unrelated CI failure — both out of scope for a same-day behavior-preserving PR. Flagging as a follow-up rather than guessing.

**Fallout fix**: one pre-existing contract test (`Tests/AudioAutomationCoverageContractTests.swift`) did a raw-text `.contains(...)` check on `run-e2e-smoke.sh` for two files that are now sourced from the shared lib instead of listed directly. Updated it to check the combined text of both files.

**check-build-source-lists.py**: updated to expand `"${SHARED_..._SOURCES[@]}"` references against the shared lib before validating, so it still checks every file each script actually compiles.

## No-functional-diff proof

For each script, expanded the *effective* array (after `source`-ing the shared lib and evaluating the array literal) and diffed the sorted file list against the original array's sorted file list, byte-for-byte:

```
run-tests.sh APP_SOURCES:              185 files before, 185 files after — diff: empty
run-e2e-smoke.sh SWIFT_SOURCES:         48 files before, 48 files after  — diff: empty
run-slow-pasteback-smoke.sh SOURCES:     3 files before, 3 files after   — diff: empty
```

Swift compiles all `.swift` arguments passed to one `swiftc` invocation as a single whole-module unit — argument order does not affect program semantics (only potential diagnostic ordering), and the two scripts already listed their overlapping files in different relative orders before this change, so reordering here changes nothing observable.

For the deps-staleness functions: diffed each of the six function bodies pairwise across `build-deps.sh`, `build.sh`, and `run-integration-smoke.sh` before extracting — all pairs were byte-identical. `build-beta.sh`'s divergent copy was left as-is (see above).

## Test plan

- [x] `python3 scripts/dev/check-build-source-lists.py` — passes
- [x] `bash build.sh --no-open` — builds and launch-smokes clean
- [x] `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12602/12602 passed (one pre-existing flaky failure in `MeetingPromptDetectorTests` reproduced identically on clean `origin/main` HEAD via `git stash`, confirmed unrelated to this change, and did not reproduce on the final run)
- [x] `bash run-e2e-smoke.sh` — `[e2e] OK - deterministic capture contract passed`
- [x] `bash run-slow-pasteback-smoke.sh` — 9/9 checks passed
- [x] `bash run-integration-smoke.sh` — core smoke + wake smoke + recovery-merge package tests all passed
- [x] `bash build-deps.sh --force` — re-run twice (once after editing build-deps.sh itself, a tracked dependency input; once after a `git stash`/`pop` round-trip touched its mtime) — both succeeded
- [x] `SKIP_NOTARIZATION=1 bash build-beta.sh '' redbars` (touched build-beta.sh) — signed DMG built successfully


r3dbars:
Addressed all three P2 guardrail-strength findings from Codex review (commit 2b50211b):

**1. `scripts/dev/check-build-source-lists.py` strictness**
- `${ARRAY[@]}` expansion now fails (not skips) on any unresolved array reference — verified against a typo'd array name (`SHARED_TEST_STORAGE_SOURCEZ`) and a malformed reference (`[*]` instead of `[@]`).
- Fails if a referenced shared array expands to zero entries — verified by emptying `SHARED_PASTEBACK_SUPPORT_SOURCES`.
- Confirmed macOS's bash 3.2 (still the system `/bin/bash`) aborts with `unbound variable` under `set -u` on a bare `"${ARR[@]}"` expansion of an empty array — fixed upstream in bash 4.4+, but macOS never moved past 3.2. Updated all three consumer scripts (`run-tests.sh`, `run-e2e-smoke.sh`, `run-slow-pasteback-smoke.sh`) to use the `${ARR[@]+"${ARR[@]}"}` guard instead of a bare `"${ARR[@]}"`.

Synthetic failure output (typo'd array name):
```
Build source-list validation failed:
- scripts/entrypoints/lib/shared-smoke-sources.sh: missing SHARED_TEST_STORAGE_SOURCEZ=(...) array
```
Synthetic failure output (malformed `[*]` reference):
```
Build source-list validation failed:
- unresolved array reference in source list: '${SHARED_TEST_STORAGE_SOURCES[*]}'
```
Synthetic failure output (zero-entry array):
```
Build source-list validation failed:
- scripts/entrypoints/lib/shared-smoke-sources.sh: SHARED_PASTEBACK_SUPPORT_SOURCES expands to zero entries
```
All synthetic breakages were reverted after confirming the failure; `python3 scripts/dev/check-build-source-lists.py` passes cleanly on the real tree.

**2. `Tests/AudioAutomationCoverageContractTests.swift` precision**
Replaced the concatenated-text search with two assertions per required file: (a) `run-e2e-smoke.sh` actually expands the named shared array (`${SHARED_TEST_STORAGE_SOURCES[@]`), and (b) the file is a member of that specific array's block in `shared-smoke-sources.sh` (via a new `extractShellArrayBlock` helper scoped to one array's text). Verified this catches a synthetic wrong-array migration (moved `LocalMeetingSummarizer.swift` from `SHARED_TEST_STORAGE_SOURCES` to `SHARED_PASTEBACK_SUPPORT_SOURCES`) that both the old concatenated-text search and `check-build-source-lists.py` miss (the file is still compiled somewhere, and still present in the file's raw text, so neither of those catches a migration to the wrong array). Reverted after confirming the failure.

**3. `.agents/test-matrix.yml` coverage**
Added `scripts/entrypoints/lib/shared-smoke-sources.sh` to the `run-tests.sh` / `run-e2e-smoke.sh` / `run-slow-pasteback-smoke.sh` rules, and `scripts/entrypoints/lib/deps-staleness.sh` to the `build-deps.sh` / `build.sh` / `run-integration-smoke.sh` rules. An edit to either shared lib alone now matches all of its consumers' rules, and the existing "run the union of checks" note selects the right verification set (source-list check + all three smokes for the shared source list; build-deps/build/integration-smoke for the deps-staleness functions).

**Verification (scoped, foreground only):**
- `python3 scripts/dev/check-build-source-lists.py` — passes (plus the three synthetic-failure demos above)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12607/12607 passed (5 more tests than before, from the new precise per-file assertions)
- `bash scripts/dev/agent-preflight.sh` — passes, correctly selects the expanded check union from the `.agents/test-matrix.yml` change
- `python3 scripts/dev/test-matrix-checks.py --self-test` — passes

PR #1625, opened 2026-08-04
Factor settings toggle bindings into a shared persist/track helper

r3dbars:
## Summary

- `TranscriptedSettingsView.swift` hand-wrote ~29 `Binding(get:set:)` blocks, most following the same triple: mirror local `@State` -> track settings telemetry -> persist to a preferences helper. Added `persistedSettingsBinding(_:persist:track:sideEffect:)` in `TranscriptedSettingsComponents.swift` and routed every site that matches that shape through it. Behavior-preserving: same persist call, same telemetry label/page, same side effects per site.
- Collapsed the live-meeting Codex sidecar toggle, which existed twice (`TranscriptedSettingsView`'s Beta page and `AgentConnectionSettingsPage`'s Agent page) with byte-identical `prepare*/stop*` workspace functions, into one `LiveMeetingSidecarController` (`Sources/UI/Shared/LiveMeetingSidecarController.swift`) that both toggle sites now call for `prepareWorkspaceForUse()`, `stop(meetingSession:)`, and `disable(enabled:meetingSession:)`. Deleted the duplicated private functions in both files.
- Did not touch `SpeakerPeopleSettingsSection.swift` or restructure any page parameter lists.

## `persistedSettingsBinding` order note

The helper runs `mirror -> track -> persist -> sideEffect`, always in that fixed order. A few original hand-written sites ran `persist` before `track`. In those cases `track` and `persist` are independent side effects — telemetry never reads back the just-written preference value, and the preference write doesn't depend on telemetry having fired — so normalizing the order is behavior-preserving. Call sites where the original had additional/conditional logic beyond a single trailing side-effect closure, **or where the two calls are order-dependent**, were left hand-written rather than forced through the helper (see below) — `anonymousAnalyticsEnabled` is the one order-dependent case, caught in review (see Concerns).

## Site-by-site (all 29 `Binding(get:set:)` sites)

**Converted to `persistedSettingsBinding` (18):**

| Property | Persist | Track |
|---|---|---|
| `showTranscriptedInDock` | `DockVisibilityPreferences.setVisible` | `show_in_dock` |
| `uiSoundsEnabled` | `UISoundPreferences.setEnabled` | `dictation_sounds` |
| `dictationCleanupEnabled` | `DictationCleanupPreferences.setEnabled` | `dictation_cleanup` |
| `dictationOverlayMode` | `DictationOverlayPresentationPreferences.setMode` | action: `change_dictation_overlay_mode` |
| `confirmQuitDuringMeetingEnabled` | `QuitConfirmationPreferences.setConfirmQuitDuringActiveMeetingRecording` | `meeting_quit_confirmation` |
| `autoDetectCallsEnabled` | `AutoCallDetectionPreferences.setEnabled` | `auto_call_detection` |
| `missedCallNudgeEnabled` | `MissedCallNudgePreferences.setEnabled` | `missed_call_nudge` |
| `dictationShortcutsEnabled` | `HotkeyPreferences.setDictationShortcutsEnabled` | `dictation_shortcuts` |
| `keepRecommendedMicrophoneActive` | `DictationPersistentInputPreferences.setEnabled` | `keep_recommended_microphone_active` |
| `preferredDictationInputUID` (Picker) | `DictationPersistentInputPreferences.setPreferredDeviceUID` | action: `change_preferred_dictation_microphone` |
| `autoEnterEnabled` | `DictationAutoSendPreferences.setEnabled` | `auto_send` |
| `autoEnterKey` (Picker) | `DictationAutoSendPreferences.setSendKey` | action: `change_auto_send_key` |
| `meetingMicProcessingMode` (Picker) | `MicrophoneProcessingPreferences.setMode` | `meeting_mic_processing_<mode>` |
| `splitLocalSpeakersEnabled` | `LocalSpeakerPreferences.setEnabled` | `local_speaker_split` |
| `crashReportingEnabled` | `CrashReportingPreferences.setEnabled` | `crash_reporting` (+ sideEffect: `CrashReporter.applySessionTrackingPreference()`, clear `sentryTestStatus`/`diagnosticsActionStatus`) |
| `localMeetingSummariesEnabled` (Beta) | `LocalMeetingSummaryPreferences.setEnabled` | `local_ai_meeting_summaries` (+ sideEffect: `handleLocalMeetingSummaryToggle`) |
| `liveMeetingSidecarEnabled` (Beta) | `LiveMeetingCodexPreferences.setEnabled` | `live_meeting_sidecar` (+ sideEffect: `handleBetaLiveMeetingSidecarToggle`, now backed by `LiveMeetingSidecarController`) |
| `nemotronModelEnabled` (Beta) | `SpeechModelBetaPreferences.setNemotronBetaEnabled` | `nemotron_streaming_model` |

**Left hand-written (11)** — each has branching, multi-step, order-dependent, or fully-delegated logic beyond what a single trailing side-effect closure covers:

- `rootAlertBinding` (~1780) — alert dismissal routing between two separate `@State` alerts, no preference persist.
- `launchAtLoginEnabled` (~2323) — delegates entirely to `updateLaunchAtLogin(_:)`, which has its own do/catch error handling and status refresh.
- `preferredTranscriptionModel` Picker (~2420) — delegates entirely to `updatePreferredTranscriptionModel(_:page:)`.
- "Better speaker matching on calls" toggle (~2467) — branches to a confirmation alert when named speakers exist vs. applying immediately.
- Per-app auto-send toggle (~2637) — delegates entirely to `setAutoEnterApp(_:isAllowed:page:)`; no simple `@State` mirror (backed by a `Set`).
- `anonymousAnalyticsEnabled` (~2733) — **order-dependent, not just independent side effects**: `AnalyticsReporter.trackEvent` drops any event fired while `AnalyticsPreferences` reads disabled, so the toggle's own `"anonymous_analytics"` transition event needs persist-before-track on enable and track-before-persist on disable to avoid silently losing that event. Restored the original hand-written asymmetric ordering with a comment explaining the constraint (caught in Codex review of an earlier revision of this PR, where routing it through `persistedSettingsBinding`'s fixed `track -> persist` order silently dropped the opt-in event).
- `localMeetingSummaryProvider` Picker (~2975) — multi-step conditional: refreshes setup status, and if summaries are enabled, cancels jobs/clears notices/re-prepares the model.
- Correction row `spoken`/`replacement` bindings (~2794, ~2798) — delegate to `updateCorrectionSpoken`/`updateCorrectionReplacement` per row identity, not a `@State` mirror.
- `customDictionaryText` TextEditor (~2868) — delegates entirely to `updateCustomDictionaryText(_:)`.
- `captureLibraryChoicePromptBinding` (~4288) — sheet-presentation binding, no preference persist.

## Test plan

- [x] `bash build.sh --no-open` — clean build (after `bash build-deps.sh --force` to refresh stale `TranscriptedCore` deps)
- [x] `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12607 tests, 12607 passed, 0 failed
- [x] Re-ran both after the review fix-up below — still 12607/12607, clean build

## Concerns

- `persist`/`track` order was normalized to a fixed `mirror -> track -> persist -> sideEffect` sequence across every converted site, even where the original ran persist before track. This is behavior-preserving for every remaining converted site since telemetry and preference writes are independent there, but it's a real (if inert) semantic shift worth a second look.
- **Fixed in review**: `anonymousAnalyticsEnabled` was initially converted to `persistedSettingsBinding` too. Codex review of an earlier revision flagged that the helper's fixed `track -> persist` order means, on opt-in, `trackSettingsToggle("anonymous_analytics", enabled: true)` fires while `AnalyticsPreferences` still reads disabled — and `AnalyticsReporter.trackEvent` silently drops any event sent while analytics reads disabled, so the opt-in transition event was lost. Reverted that one site to a hand-written `Binding` with the original asymmetric ordering (persist-then-track on enable, track-then-persist on disable) and a comment stating the constraint; did not add an ordering knob to the shared helper for this one site. Moved it to the hand-written list above.
- Preference setters like `DockVisibilityPreferences.setVisible` take a defaulted `userDefaults:` parameter, so they can't be passed as bare function references to `persist:` (Swift keeps the full signature) — every `persist:` argument is wrapped in an explicit `{ Foo.setBar($0) }` closure instead.
- `LiveMeetingSidecarController` is `@MainActor` because it calls into `MeetingSessionController` (itself `@MainActor`); both call sites are `View`-conforming settings pages so this should be transparent, but flagging since it's a new cross-file dependency surface.

### DIFF
--- a/Sources/UI/Settings/AgentConnectionSettingsPage.swift
+++ b/Sources/UI/Settings/AgentConnectionSettingsPage.swift
@@ -231,8 +231,7 @@ struct AgentConnectionSettingsPage: View {
                 if enabled {
                     prepareLiveMeetingSidecarWorkspace()
                 } else {
-                    meetingSession?.stopLiveCodexSessionFromSettings()
-                    stopLiveMeetingSidecarPreview()
+                    LiveMeetingSidecarController.stop(meetingSession: meetingSession)
                 }
             }
 
@@ -686,7 +685,7 @@ struct AgentConnectionSettingsPage: View {
         do {
             liveMeetingCodexEnabled = true
             LiveMeetingCodexPreferences.setEnabled(true)
-            let workspaceURL = try prepareLiveMeetingSidecarWorkspaceForUse()
+            let workspaceURL = try LiveMeetingSidecarController.prepareWorkspaceForUse()
             copyText(AgentConnectionGuide.liveMeetingCodexSetupPrompt(workspaceURL: workspaceURL))
             ActivationTelemetry.trackAgentPromptAction(
                 promptKind: .liveMeetingCodexSetup,
@@ -748,7 +747,7 @@ struct AgentConnectionSettingsPage: View {
         liveMeetingCodexSetupErrorDetails = nil
 
         do {
-            _ = try prepareLiveMeetingSidecarWorkspaceForUse()
+            _ = try LiveMeetingSidecarController.prepareWorkspaceForUse()
         } catch {
             disableLiveMeetingSidecarAfterFailure()
             liveMeetingCodexSetupError = AgentSetupFailureCopy.liveMeetingsPrepare
@@ -763,7 +762,7 @@ struct AgentConnectionSettingsPage: View {
         do {
             liveMeetingCodexEnabled = true
             LiveMeetingCodexPreferences.setEnabled(true)
-            let workspaceURL = try prepareLiveMeetingSidecarWorkspaceForUse()
+            let workspaceURL = try LiveMeetingSidecarController.prepareWorkspaceForUse()
             copyText(AgentConnectionGuide.liveMeetingCoworkSetupPrompt(workspaceURL: workspaceURL))
             ActivationTelemetry.trackAgentSetupCTA(
                 setupKind: .liveSidecar,
@@ -798,7 +797,7 @@ struct AgentConnectionSettingsPage: View {
         do {
             liveMeetingCodexEnabled = true
             LiveMeetingCodexPreferences.setEnabled(true)
-            let workspaceURL = try prepareLiveMeetingSidecarWorkspaceForUse()
+            let workspaceURL = try LiveMeetingSidecarController.prepareWorkspaceForUse()
             let previewURL: URL
             if #available(macOS 14.0, *) {
                 previewURL = try LiveMeetingPreviewServer.shared.start(workspaceURL: workspaceURL)
@@ -845,25 +844,8 @@ struct AgentConnectionSettingsPage: View {
         }
     }
 
-    private func prepareLiveMeetingSidecarWorkspaceForUse() throws -> URL {
-        let workspaceURL = try AgentConnectionGuide.ensureLiveMeetingCodexWorkspace()
-        if #available(macOS 14.0, *) {
-            _ = try LiveMeetingPreviewServer.shared.start(workspaceURL: workspaceURL)
-        }
-        return workspaceURL
-    }
-
-    private func stopLiveMeetingSidecarPreview() {
-        if #available(macOS 14.0, *) {
-            LiveMeetingPreviewServer.shared.stop()
-        }
-    }
-
     private func disableLiveMeetingSidecarAfterFailure() {
-        liveMeetingCodexEnabled = false
-        LiveMeetingCodexPreferences.setEnabled(false)
-        meetingSession?.stopLiveCodexSessionFromSettings()
-        stopLiveMeetingSidecarPreview()
+        LiveMeetingSidecarController.disable(enabled: $liveMeetingCodexEnabled, meetingSession: meetingSession)
     }
 
     // MARK: - Small helpers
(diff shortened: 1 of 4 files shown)


r3dbars:
Addressed the P2 from Codex review: `anonymousAnalyticsEnabled` (~2733) is reverted from `persistedSettingsBinding` back to a hand-written `Binding` with the original asymmetric ordering — persist-then-track on enable, track-then-persist on disable — plus a comment stating why (`AnalyticsReporter.trackEvent` drops events fired while `AnalyticsPreferences` reads disabled, so the fixed `track -> persist` order in the shared helper was silently dropping the opt-in transition event). No ordering knob was added to `persistedSettingsBinding` — this stays a one-off hand-written site, now called out in the PR body's hand-written list and the helper's own doc comment.

Verified after the fix (both foreground):
- `bash build.sh --no-open` — clean build
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12607 tests, 12607 passed, 0 failed

Pushed as `9523ff63`. PR body updated to move this site to the hand-written enumeration with the reason.

PR #1627, opened 2026-08-04
Add typed PipelineErrorKind to stop failure classification drift

r3dbars:
## Summary

Failure meaning for the transcription pipeline was re-derived from human-readable `errorMessage` strings by **five independent `.contains(...)` keyword parsers**, even though `TranscriptionTaskManager` has the typed Swift `Error` in hand at the moment a failure occurs. These parsers had already drifted apart once in production (commit `7c70a0c1`, "fix: align pipelineBusy classifier with actual TranscriptionTaskManager error message").

This PR adds a stable, `Codable` `PipelineErrorKind` enum, computes it **once** where the typed error still exists, threads it through `FailedTranscription` as an optional field, and converts the downstream classifiers to prefer it — keeping the existing keyword matching only as an explicit legacy fallback for entries with no typed kind.

## The five parsers (audit)

1. `TranscriptionTaskManager.safeFailureDiagnosticMessage` (Core) — typed `PipelineError` switch + `.contains(...)` text fallback → canonical display string. **Has the typed error, throws the type info away.**
2. `importedAudioFailureDisplayMessage` / `savedAudioRetranscriptionFailureDisplayMessage` (Core) — re-parses the canonical string from #1 via more `.contains(...)` chains into user-facing copy.
3. `FailedTranscription.pipelineError` + `isRetryable` (Core) — re-parses the persisted `errorMessage` back into a `PipelineError` via its own keyword chain + a separate "permanent" keyword list.
4. `MeetingFailureKind.classify(message:)` (Meeting) — a third independent keyword pass, with a much broader taxonomy (permission gates, import prep, stop timeout, speaker-name finalization, ...) since it also classifies failures that never flow through Core's typed `PipelineError` at all.
5. `MeetingFailureExplanation` retryability derivation (Meeting) — on inspection this one **already switches on the typed `MeetingFailureKind` enum**, not raw strings, so nothing needed to change there directly; it benefits automatically once #4 gets typed-aware.

## Bucket enumeration

`PipelineErrorKind` covers the buckets reachable from `TranscriptionTaskManager`'s own classification (its typed `PipelineError` switch, unioned with its `.contains(...)` text fallback — since `FailedTranscription.errorMessage` is often the raw, non-canonicalized `error.localizedDescription`, not the canonical string, so the fallback matters):

| `PipelineErrorKind` | Canonical message | `isRetryable` | Maps to `MeetingFailureKind` |
|---|---|---|---|
| `transcriptionAlreadyInProgress` | "Transcription already in progress" | true | `.pipelineBusy` |
| `missingSystemAudio` | `PipelineError.missingSystemAudio.localizedDescription` | false* | `.systemAudioPermission` |
| `recordingTooShort` | "Recording too short" | false | `.recordingTooShort` |
| `emptyAudioFile` | "Empty audio file" | false | `.emptyAudio` |
| `noSpeechDetected` | "No speech detected" | false | `.noSpeechDetected` |
| `invalidAudioFormat` | "Invalid audio format" | false | `.invalidAudioFormat` |
| `microphoneAudioUnusable` | "Microphone audio was not usable" | false | `.microphoneAudioUnusable` |
| `saveFailed` | "Failed to save transcript" | true | `.saveFailed` |
| `modelNotLoaded` | "Model not loaded" / "\<model\> model not loaded" | true | `.modelNotLoaded` |
| `diarizationFailed` | "Diarization failed" | true | `.diarizationFailed` |
| `transcriptionInferenceFailed` | "Transcription inference failed" / "\<model\> inference failed" | true | `.transcriptionInferenceFailed` |
| `pipelineFailed` | "Pipeline failed" (default fallback) | true | `.pipelineFailed` |

\* `missingSystemAudio` is forced retryable when `systemAudioURL == nil` (mic-only capture can still retry once permission is granted) — preserved from the existing logic on both the typed and legacy paths.

`MeetingFailureKind`'s remaining buckets (`microphonePermission`, `microphoneMissing`, `audioDeviceUnavailable`, `speakerNameFinalizationFailed`, `speakerFinalizationFailed`, `modelDownloadFailed`, `importFileMissing`, `importFileUnreadable`, `importUnsupportedFile`, `importCopyFailed`, `savedBeforeQuit`, `stopTimeout`, `unexpectedError`) come from failure sources that never carry a typed Core `PipelineError` — permission gates, import preparation, stop-timeout bookkeeping, speaker-name finalization. Those stay on the legacy string-matching fallback, since there is no typed error to compute a kind from in the first place; giving `PipelineErrorKind` cases for them would be untestable dead weight (nothing could ever set them, and the contract test below requires every case to have a real canonical message).

## Where it's threaded

- Computed once in `TranscriptionTaskManager`'s primary transcription-failure catch block, the retry-failure catch block, the imported-audio failure path, and the saved-audio-retranscription failure path — everywhere the typed `Error` is directly in hand.
- `TranscriptionTaskManager` gained `lastFailureErrorKind` (published, alongside the existing `lastFailureDiagnosticMessage`) for the live/in-session failure surface.
- `FailedTranscriptionManager.addFailedTranscription` / `updateFailedTranscriptionError` / all internal reconciliation rewrites now carry `errorKind` through persistence instead of dropping it.
- `FailedTranscription.isRetryable` checks `errorKind` first (including the mic-only `missingSystemAudio` special case), falling back to the renamed `legacyIsRetryable` / `legacyPipelineError` only when `errorKind == nil`.
- `MeetingFailureKind` gained `classify(errorKind:message:)` (typed-first, falls back to the pre-existing `classify(message:)` when `errorKind` is nil) and `init(errorKind:)` (the total, non-failable Core→Meeting mapping). `FailedMeetingPresentation` and `MeetingSessionController`'s terminal-failure handling now pass `errorKind` through.
- Import-preparation, permission-gate, and other non-Core failure sources are untouched — they never had a typed kind and still don't.

## Compatibility proof (old persisted `FailedTranscription` rows)

`errorKind` is a new optional `var` on a `Codable` struct with synthesized `Codable` conformance, so old on-disk rows missing the key decode to `errorKind == nil` for free. Proven directly rather than just asserted:

- `testFailedTranscriptionDecodesLegacyJSONMissingErrorKindField` (`Tests/TranscriptedCoreTests/PipelineTests/FailedTranscriptionManagerTests.swift`) — encodes a modern entry, **strips the `errorKind` key entirely** (not just nulls it — matching what's actually on disk for pre-existing rows) via `JSONSerialization`, decodes it back, and asserts `errorKind == nil` plus that `isRetryable` still falls back correctly.
- `testFailedTranscriptionManagerLoadsLegacyQueueFileWithoutErrorKind` — same strip-the-key trick but through the full `FailedTranscriptionManager(paths:)` on-disk load path (the actual persisted-queue-file format), confirming the manager loads legacy rows without failing or defaulting oddly.
- `testFailedTranscriptionErrorKindTakesPrecedenceOverLegacyMessageMatching` — same message text, one entry typed and one not, proving the typed field is actually consulted first rather than just carried as inert metadata.

## Contract test (pins typed vs. legacy so they can't drift again)

`Tests/PipelineErrorKindContractTests.swift` (fast-test runner) asserts, for every `PipelineErrorKind` case, that classifying via the typed path (`MeetingFailureKind.classify(errorKind:message:)`) and via the legacy string path (`classify(message:)` on that kind's canonical message) yield the **same `MeetingFailureKind`** and the **same `isRetryable`**. It also asserts every `PipelineErrorKind` case has a canonical-message fixture, so a new case can't silently skip the contract.

`TranscriptionPipelineErrorPolicyTests.swift` (Core SPM target) complements this with `failureKind(for:)`/`safeFailureDiagnosticMessage(for:)` pinning: both are computed from one shared internal pass (`failureClassification(for:)`), so kind and message literally cannot drift from each other at the source.

(These live in two files because `Sources/Meeting/*` and `Sources/TranscriptedCore/*` are different compilation boundaries — the fast-test runner is a hand-curated flat `swiftc` compile with no module linking, while `TranscriptionTaskManager` is only reachable from the SPM package's `swift test`. `MeetingFailureKind.swift`'s `import TranscriptedCore` is now guarded behind `#if canImport(TranscriptedCore)`, matching the existing pattern used elsewhere in `Sources/Meeting`/`Sources/UI`, and `Sources/TranscriptedCore/Models/FailedTranscription.swift` was added to `run-tests.sh`'s flat source list so `PipelineErrorKind` is visible there too.)

## Verification

All run against a clean `build-deps.sh --force` in this worktree (Core sources changed):

- `bash build-deps.sh --force` — clean
- `bash build.sh --no-open` — clean (two pre-existing, unrelated `[#no-usage]` warnings in `MeetingSessionController.swift`/`DictationSessionController.swift`, not touched by this change)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12636 tests, 12636 passed, 0 failed
- `bash run-integration-smoke.sh` — core smoke + wake-recovery smoke + recovery-merge package tests all pass
- `swift test` — 884 tests, 13 skipped, 0 failures (includes the new contract + legacy-decode tests above)

## Concerns / follow-ups

- The synthetic hardcoded-string early-return paths in `TranscriptionTaskManager` (e.g. rejecting a second concurrent transcription, or a too-short recording caught before the pipeline even starts) still don't set `errorKind` — they're known-constant strings rather than typed errors, so they keep relying on the (unchanged, still-tested) legacy string fallback. Threading explicit kinds through those too would be a reasonable, low-risk follow-up but was left out to keep this diff focused on the actual typed-error path.
- `MeetingFailureExplanation` needed no code changes — it already switched on the typed `MeetingFailureKind` enum rather than parsing strings, so it inherits the fix once its callers pass a typed-aware `MeetingFailureKind` in.

### DIFF
--- a/Sources/Meeting/FailedMeetingPresentation.swift
+++ b/Sources/Meeting/FailedMeetingPresentation.swift
@@ -6,7 +6,7 @@ enum FailedMeetingPresentation {
         from failed: FailedTranscription,
         isRetrying: Bool
     ) -> MeetingSessionController.FailedMeetingItem {
-        let failureKind = MeetingFailureKind.classify(message: failed.errorMessage)
+        let failureKind = MeetingFailureKind.classify(errorKind: failed.errorKind, message: failed.errorMessage)
         let availableAudioURLs = audioURLs(for: failed)
         let hasRetryableAudioFiles = failed.audioFilesExist()
         let copy = MeetingFailureCopy.make(
--- a/Sources/Meeting/MeetingFailureKind.swift
+++ b/Sources/Meeting/MeetingFailureKind.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if canImport(TranscriptedCore)
+import TranscriptedCore
+#endif
 
 enum MeetingFailureKind: String {
     case systemAudioPermission = "system_audio_permission"
@@ -36,6 +39,53 @@ enum MeetingFailureKind: String {
         }
     }
 
+    /// Maps a typed Core pipeline failure onto this taxonomy. `PipelineErrorKind`
+    /// only covers failures that flow through `TranscriptionTaskManager`'s typed
+    /// error surface, so this is a total, non-failable mapping over that subset —
+    /// it never produces `.unexpectedError`. Meeting-only categories (permission
+    /// gates, import preparation, stop timeouts, speaker-name finalization, ...)
+    /// have no `PipelineErrorKind` counterpart and stay on the legacy string path.
+    init(errorKind: PipelineErrorKind) {
+        switch errorKind {
+        case .transcriptionAlreadyInProgress:
+            self = .pipelineBusy
+        case .missingSystemAudio:
+            self = .systemAudioPermission
+        case .recordingTooShort:
+            self = .recordingTooShort
+        case .emptyAudioFile:
+            self = .emptyAudio
+        case .noSpeechDetected:
+            self = .noSpeechDetected
+        case .invalidAudioFormat:
+            self = .invalidAudioFormat
+        case .microphoneAudioUnusable:
+            self = .microphoneAudioUnusable
+        case .saveFailed:
+            self = .saveFailed
+        case .modelNotLoaded:
+            self = .modelNotLoaded
+        case .diarizationFailed:
+            self = .diarizationFailed
+        case .transcriptionInferenceFailed:
+            self = .transcriptionInferenceFailed
+        case .pipelineFailed:
+            self = .pipelineFailed
+        }
+    }
+
+    /// Preferred entry point once a typed `errorKind` is available (e.g. from
+    /// `FailedTranscription.errorKind`). Falls back to legacy string matching
+    /// via `classify(message:)` only when `errorKind` is nil — persisted
+    /// entries from before typed errors existed, or failures that never had
+    /// a `PipelineError` in the first place.
+    static func classify(errorKind: PipelineErrorKind?, message: String) -> MeetingFailureKind {
+        if let errorKind {
+            return MeetingFailureKind(errorKind: errorKind)
+        }
+        return classify(message: message)
+    }
+
     static func isRecordingTooShortMessage(_ message: String) -> Bool {
         let normalized = message.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
         let mentionsAudioMinimum = (
@@ -57,6 +107,11 @@ enum MeetingFailureKind: String {
             || mentionsAudioMinimum
     }
 
+    /// Legacy fallback: classifies by keyword matching over a free-form message.
+    /// Prefer `classify(errorKind:message:)` when a typed `PipelineErrorKind` is
+    /// available. This stays byte-equivalent to the pre-typed-error behavior so
+    /// old persisted `FailedTranscription` rows (with `errorKind == nil`) keep
+    /// classifying the same way.
     static func classify(message: String) -> MeetingFailureKind {
         let normalized = message.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
 
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -2751,7 +2751,11 @@ final class MeetingSessionController: ObservableObject {
             activeStoppedAudioRecovery = nil
             let transcriptionTrigger = activeTranscriptionTrigger
             let diagnosticMessage = taskManager.lastFailureDiagnosticMessage ?? message
-            let failureKind = MeetingFailureKind.classify(message: diagnosticMessage)
+            // Only trust the typed kind when it was captured alongside the diagnostic
+            // message actually being classified below — not when this fell back to
+            // the raw overlay `message`, which never had a typed kind computed for it.
+            let errorKind = taskManager.lastFailureDiagnosticMessage != nil ? taskManager.lastFailureErrorKind : nil
+            let failureKind = MeetingFailureKind.classify(errorKind: errorKind, message: diagnosticMessage)
             if failureKind.shouldReportAsSkippedTranscript {
                 finishLiveCodexSessionForCurrentTranscriptionFailureIfNeeded(
                     failedJobID: activeQueuedTranscriptionJobID
--- a/Sources/TranscriptedCore/Models/FailedTranscription.swift
+++ b/Sources/TranscriptedCore/Models/FailedTranscription.swift
@@ -1,5 +1,49 @@
 import Foundation
 
+/// Stable, persistable classification of a pipeline failure.
+///
+/// This is the single source of truth for "what kind of failure was this"
+/// once a typed `PipelineError` (or an error `TranscriptionTaskManager` can
+/// otherwise classify) is in hand. It replaces re-deriving the same meaning
+/// from `errorMessage` strings in multiple independent places.
+///
+/// Cases mirror the buckets `TranscriptionTaskManager.safeFailureDiagnosticMessage`
+/// already distinguishes (see its switch over `PipelineError` plus its text
+/// fallback), because that is the one place a typed Swift `Error` is still in
+/// hand when a failure is classified. Downstream consumers such as
+/// `MeetingFailureKind` fold these into their own broader taxonomy — which
+/// also covers failure sources (permission gates, import preparation, stop
+/// timeouts, ...) that never produce a `PipelineError` and so never populate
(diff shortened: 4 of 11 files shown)


r3dbars:
## Fixed: P1 from Codex review (broad text net leaking into persisted `errorKind`)

Codex's review caught a real regression: `TranscriptionTaskManager.failureKind(for:)` was sourcing the persisted `errorKind` from the same broad `.contains(...)` text-fallback net that `safeFailureDiagnosticMessage` uses to pick a *display* message. That net is deliberately broader than `FailedTranscription`'s row-level legacy parser (`legacyPipelineError` / the `legacyIsRetryable` permanent-keyword list), which only classifies retry-eligibility from a narrower set of exact phrases.

Concretely: a raw AVAudioFile `NSError` whose `localizedDescription` contains `"com.apple.coreaudio.avfaudio"`/`"coreaudio error"` would get classified by the broad net as `.invalidAudioFormat`, and — because I was persisting that as `errorKind` at the throw site — `FailedTranscription.isRetryable` would consult `errorKind.isRetryable` first and return `false`, hiding "Try Again" for what was, on `main`, a transient CoreAudio device error (`legacyPipelineError` returns `nil` for that text, so it fell to the permanent-keyword list, which doesn't match, so it stayed retryable and bucketed as `unexpected_error` in `MeetingFailureKind`).

**Fix (`ba45274c`):** `failureKind(for:)` now only returns non-nil for a genuinely typed `PipelineError` case — via its own switch, completely independent of the text-fallback classification pass — and explicitly excludes `PipelineError.unknown(underlying:)`, since that case just wraps free text rather than representing a real typed classification. Every non-`PipelineError` `Error`, and every `.unknown` case, now yields `errorKind == nil`, so those rows fall through to the unchanged legacy string fallback — exactly matching `main`'s behavior. The display-message classification (`safeFailureDiagnosticMessage`) is untouched; only what gets *persisted* as `errorKind` changed.

**Tests added** (`Tests/TranscriptedCoreTests/PipelineTests/TranscriptionPipelineErrorPolicyTests.swift`):
- `testFailureKindReturnsNilForNonPipelineErrors` — the full text-routed table now asserts `nil` kind (previously asserted broad-net kinds).
- `testFailureKindReturnsNilForUnknownPipelineError` — `.unknown(...)` also yields `nil`.
- Three explicit regression tests using the reviewer's examples verbatim: `testFailureKindIsNilForCoreAudioDeviceErrorText` (the `-50` NSError), `testFailureKindIsNilForBroadEmptyAudioVariantText` ("Empty audio was captured..." — broader than "empty audio file"), `testFailureKindIsNilForScreenRecordingVariantText` ("...Screen Recording access was revoked..." — doesn't match the legacy permanent list's exact "System Audio Recording" phrase). Each constructs a real `FailedTranscription` with the computed `errorKind` and asserts `isRetryable == true`, matching `main`.

**Bucket verification** (`Tests/MeetingFailureKindTests.swift`): since `errorKind` is now `nil` for all three examples, `MeetingFailureKind.classify(errorKind: nil, message:)` always defers to the untouched `classify(message:)` — I checked what each message actually buckets to today (unchanged from `main`) rather than assuming: the CoreAudio message legitimately falls through to `.unexpectedError` (confirmed), but the "empty audio" and "screen recording" variants each already match `MeetingFailureKind`'s *own* independent `"empty audio"` / `"screen recording"` keywords (`.emptyAudio` and `.systemAudioPermission` respectively) — that's pre-existing, correct, `main` behavior, not something this PR touches or regresses. The three new `MeetingFailureKindTests` assertions pin those actual (verified, not assumed) values.

## P3: `modelInferenceFailed` → `transcriptionInferenceFailed` model-name divergence

Added a comment at the `.modelInferenceFailed` case in `failureKind(for:)` acknowledging the divergence: legacy text classification would bucket an underlying message naming a diarization model (e.g. "PyAnnote") as `.diarizationFailed`, while the typed switch always maps to `.transcriptionInferenceFailed` regardless of model name. Confirmed there are zero throw sites for `PipelineError.modelInferenceFailed` in the codebase today (grepped `Sources/TranscriptedCore/`), so this is documented as a latent, currently-unreachable divergence rather than fixed with model-name branching — if a diarization engine ever starts throwing `.modelInferenceFailed`, the comment flags that the typed switch should special-case the model name the way the legacy text path does.

## Re-verified (foreground, clean `build-deps.sh --force`)

- `bash build-deps.sh --force` — clean
- `bash build.sh --no-open` — clean
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12639 tests, 12639 passed, 0 failed
- `bash run-integration-smoke.sh` — clean
- `swift test` — 886 tests, 13 skipped, 0 failures

PR #1629, opened 2026-08-04
Replace pipeline cancellation param-threading with a rollback registry

r3dbars:
## Summary

`transcribeMultichannelPipeline` (`Sources/TranscriptedCore/Pipeline/TranscriptionPipelineRunner.swift`) called `checkCancellationAfterTranscriptSideEffects(...)` at six checkpoints as the pipeline progressed, re-passing a growing hand-assembled "what to roll back" parameter list at each one (`savedURL`, `retainedAudioDirectory`, `retainedAudioURLs`, `speakerClipURLs`, `queuedSpeakerRequestTranscriptId`, `deleteSavedTranscriptOnCancellation`, `replacementTranscriptRollback`). Every new side effect had to be manually added to every subsequent call site, or a late cancellation would leak that artifact.

This replaces the threaded-parameter approach with `PipelineRollbackRegistry`, a small local accumulator (final class, `@unchecked Sendable` only to satisfy the compiler across the pipeline's `nonisolated` ⇄ `@MainActor` hops — never shared across concurrent tasks). Each side effect registers its own undo closure the moment it happens; every checkpoint collapses to `try await rollback.checkCancellation()`, which runs registered undos **in reverse (LIFO) order** and rethrows exactly as before.

Same fix applied to `transcribeMicrophoneOnlyPipeline`, which shared the same `checkCancellationAfterTranscriptSideEffects` / `cleanupCancelledTranscriptSideEffects` / `commitSavedTranscriptSideEffectsUnlessCancelled` helpers (that's why it wasn't split off — the plumbing is genuinely shared).

## Parameter → registration mapping

| Old parameter | New registration point |
|---|---|
| `savedURL` + `deleteSavedTranscriptOnCancellation` + `replacementTranscriptRollback` | `TranscriptionTaskManager.registerSavedTranscriptRollback(savedURL:replacementTranscriptRollback:into:)`, called once right after `TranscriptSaver.saveTranscript` succeeds. `deleteSavedTranscriptOnCancellation` was always derived as `replacementTranscriptRollback == nil` at every call site, so that derivation is now folded directly into the branch instead of being recomputed and re-threaded everywhere. |
| `retainedAudioDirectory` + `retainedAudioURLs` | `registerRetainedAudioRollback(directory:urls:into:)`, called once right after `archiveRecordingAudioIfConfigured` returns. Bundles "remove files, then remove the directory if that leaves it empty" into one closure, preserving the one real ordering dependency in the original cleanup. |
| `speakerClipURLs` (system) | `registerSpeakerClipsRollback(_:channel:"system",into:)`, called right after system-speaker clip extraction, scoped to only the clips just extracted (not the whole growing `namingEntries` list the old code re-passed). |
| `speakerClipURLs` (mic) | `registerSpeakerClipsRollback(_:channel:"mic",into:)`, called right after mic-speaker clip extraction, same scoping. |
| `queuedSpeakerRequestTranscriptId` | An inline `rollback.register { … cancelSpeakerNamingRequest(transcriptId:) … }` closure registered immediately after `enqueueSpeakerNamingRequest` succeeds. |

`commitSavedTranscriptSideEffectsUnlessCancelled`'s non-cancellation "was this task's side effects superseded?" branch (`canCommitTaskSideEffects` returning false) now also routes through `rollback.rollbackAll()` instead of a parallel hand-rolled cleanup call — same set of undos, same effect, one code path instead of two.

## Behavior notes / things I checked

- **Conditional rollback (the old `deleteSavedTranscriptOnCancellation` flag):** always tied 1:1 to `replacementTranscriptRollback == nil` at every original call site, so folding it into `registerSavedTranscriptRollback`'s branch is behavior-preserving, not just a simplification.
- **Ordering:** the original cleanup ran in a fixed order (transcript → retained-audio-files → retained-audio-dir → clips) regardless of when cancellation was detected. The new registry runs undos LIFO (most-recently-registered first), so the *order* of independent cleanup categories now differs from before. This is safe because all of them operate on disjoint filesystem paths with only one real dependency — retained-audio files must be removed before checking whether the directory is empty — which is preserved by keeping that as one atomic closure. The one *behaviorally* order-sensitive part (dequeue the speaker-naming request before other cleanup) is preserved too: the dequeue closure is registered last (after enqueue), so LIFO naturally runs it first, exactly matching the original code's explicit "dequeue first, then cleanup" sequencing.
- **Non-cancellation reuse:** `checkCancellationAfterTranscriptSideEffects`/`cleanupCancelledTranscriptSideEffects` were also invoked from the "superseded task" branch inside `commitSavedTranscriptSideEffectsUnlessCancelled` (not a `Task.checkCancellation()` throw). That path now calls `rollback.rollbackAll()` directly — same registry, same undos, explicitly documented in a code comment at that call site.
- Exposed `TranscriptionTaskManager.ReplacementTranscriptRollback` and the three `register*Rollback` static helpers as internal (dropped `private`) so the new test can drive them directly against real temp-file artifacts instead of re-implementing their logic — still invisible outside the module.

## Test plan

- [x] Added `Tests/TranscriptedCoreTests/PipelineTests/PipelineRollbackRegistryTests.swift`: generic registry mechanics (LIFO order, non-cancelled no-op, `rollbackAll()` idempotency), each `register*Rollback` helper exercised individually against real files/directories, and a composed "checkpoint sequence" simulation that registers exactly what a real run would have registered by cancellation stages 1/2/3/4/6 and asserts only those artifacts are cleaned (and that artifacts from stages never reached are left untouched).
- [x] `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12607 tests, 0 failed
- [x] `swift test` — 891 tests (13 pre-existing skips), 0 failures

## Concerns

- No existing tests exercised this rollback path before this PR (confirmed via grep for the old helper names in `Tests/`), so there's no way to diff against a "before" baseline test run — correctness rests on the parameter-mapping table above plus the new tests.
- The rollback-ordering change (fixed-forward → LIFO-by-registration) is a real, deliberate behavior change from the literal old code, though not from its *effective* filesystem outcome for the reasons above. Flagging it explicitly in case reviewers want the exact original ordering preserved instead.

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipelineRunner.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipelineRunner.swift
@@ -737,10 +737,15 @@ extension TranscriptionTaskManager {
                     }
                 ))
             }
-            rollback.register { [weak self] in
-                await MainActor.run {
-                    self?.cancelSpeakerNamingRequest(transcriptId: transcriptId)
-                }
+            // Snapshot a strong local before the actor hop instead of capturing `[weak self]`
+            // into this closure: `self` isn't Sendable, so a weak capture that then crosses into
+            // a nested `@Sendable` MainActor closure warns today and is a hard error under
+            // Swift 6. The registry only ever lives for the duration of this call, so there is no
+            // retain-cycle risk in capturing strongly here — unlike `onComplete` above, which is
+            // handed off to a request object that can outlive this function.
+            let manager = self
+            rollback.register {
+                await manager.cancelSpeakerNamingRequest(transcriptId: transcriptId)
                 AppLogger.pipeline.info("Rollback: cancelled queued speaker naming request", [
                     "transcriptId": transcriptId.uuidString
                 ])
@@ -956,7 +961,11 @@ extension TranscriptionTaskManager {
         (statsStore ?? StatsDatabase.shared).recordSession(metadata)
     }
 
-    private func commitSavedTranscriptSideEffectsUnlessCancelled(
+    /// Not `private`: the "superseded task" branch below (`canCommitTaskSideEffects` returning
+    /// false without any `Task.checkCancellation()` throw) is exercised directly by
+    /// `PipelineRollbackRegistryManagerTests` so that scenario is tested against the real
+    /// method instead of only inferred from `PipelineRollbackRegistry`'s generic behavior.
+    func commitSavedTranscriptSideEffectsUnlessCancelled(
         taskId: UUID,
         savedURL: URL,
         result: TranscriptionResult,
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionPipelineRunner.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionPipelineRunner.swift
@@ -197,25 +197,25 @@ extension TranscriptionTaskManager {
         ) else {
             throw PipelineError.saveFailed(detail: "Could not write mic-only transcript to \(outputFolder.lastPathComponent)")
         }
-        let deleteSavedTranscriptOnCancellation = replacementTranscriptRollback == nil
-        try await checkCancellationAfterTranscriptSideEffects(
+        let rollback = PipelineRollbackRegistry()
+        Self.registerSavedTranscriptRollback(
             savedURL: savedURL,
-            deleteSavedTranscriptOnCancellation: deleteSavedTranscriptOnCancellation,
-            replacementTranscriptRollback: replacementTranscriptRollback
+            replacementTranscriptRollback: replacementTranscriptRollback,
+            into: rollback
         )
+        try await rollback.checkCancellation()
 
         let archiveOutcome = await archiveRecordingAudioIfConfigured(
             micURL: micURL,
             systemURL: nil,
             savedURL: savedURL
         )
-        try await checkCancellationAfterTranscriptSideEffects(
-            savedURL: savedURL,
-            retainedAudioDirectory: archiveOutcome.retainedAudioDirectory,
-            retainedAudioURLs: archiveOutcome.retainedAudioURLs,
-            deleteSavedTranscriptOnCancellation: deleteSavedTranscriptOnCancellation,
-            replacementTranscriptRollback: replacementTranscriptRollback
+        Self.registerRetainedAudioRollback(
+            directory: archiveOutcome.retainedAudioDirectory,
+            urls: archiveOutcome.retainedAudioURLs,
+            into: rollback
         )
+        try await rollback.checkCancellation()
 
         try await commitSavedTranscriptSideEffectsUnlessCancelled(
             taskId: taskId,
@@ -225,10 +225,7 @@ extension TranscriptionTaskManager {
             meetingTitle: meetingTitle,
             transcriptDate: transcriptDate,
             notifier: notifier,
-            retainedAudioDirectory: archiveOutcome.retainedAudioDirectory,
-            retainedAudioURLs: archiveOutcome.retainedAudioURLs,
-            deleteSavedTranscriptOnCancellation: deleteSavedTranscriptOnCancellation,
-            replacementTranscriptRollback: replacementTranscriptRollback
+            rollback: rollback
         )
 
         if archiveOutcome.didArchiveRecordingAudio && removeSourceAudioAfterArchive && sourceFailedTranscriptionId == nil {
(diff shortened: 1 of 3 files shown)


r3dbars:
Addressed all three points from the review:

**1. `[weak self]` across the nested `@Sendable` MainActor.run boundary (P2)**

`registerSavedTranscriptRollback`'s dequeue closure at what was line 740 captured `[weak self]` into `rollback.register { [weak self] in await MainActor.run { self?.cancelSpeakerNamingRequest(...) } } }` — a weak `Optional<TranscriptionTaskManager>` crossing into a nested `@Sendable` closure, which warns today and is a hard error under Swift 6. Fixed by snapshotting a strong local (`let manager = self`) before the actor hop and calling `await manager.cancelSpeakerNamingRequest(transcriptId:)` directly (no need for the extra `MainActor.run` wrapper — `await` on an actor-isolated method already does the hop). Safe because the registry never outlives this single pipeline call, unlike `onComplete`, which is handed to a request object that can outlive the function and genuinely needs `weak self`.

**2. Test fidelity — harness reimplementation vs. real pipeline code (P2)**

- Exposed `commitSavedTranscriptSideEffectsUnlessCancelled` as internal (dropped `private`, doc comment explains why) so the superseded/non-cancellation branch can be driven directly.
- Added `Tests/TranscriptedCoreTests/PipelineTests/PipelineRollbackRegistryManagerTests.swift` (extension on `TranscriptionTaskManagerMetadataTests`, reusing its `makeManager()`/`tempDirectory` fixtures) with three tests against a **real** `TranscriptionTaskManager`:
  - `testRealDequeueRollbackCancelsQueuedSpeakerNamingRequestAndItsOwnClipCleanup` — real `enqueueSpeakerNamingRequest` → real `cancelSpeakerNamingRequest` round trip, using the exact strong-local-capture pattern from the fix above, asserting the request is genuinely dequeued and `cancelSpeakerNamingRequest`'s own clip cleanup fires.
  - `testRealCheckpointSixSequenceCleansEveryArtifactAndDequeuesTheRequest` — full six-checkpoint shape (transcript, retained audio, both channels' clips, queued request) against a live manager and real files.
  - `testSupersededTaskRollsBackThroughRealCommitSavedTranscriptSideEffectsUnlessCancelled` — drives the real superseded-task branch (`taskId` deliberately absent from `activeTasks`, ambient `Task` never cancelled) directly against the now-internal method.
- Also added `testMicOnlyPipelineShapeOnlyEverRegistersTranscriptAndRetainedAudio` in `PipelineRollbackRegistryTests.swift` covering `transcribeMicrophoneOnlyPipeline`'s distinct two-checkpoint shape (no clips, no naming request), and `testRollbackAllRunsEveryUndoEvenWhileTheAmbientTaskIsAlreadyCancelled` proving `rollbackAll()`'s undo loop runs to completion even when the ambient task is genuinely cancelled throughout (it must not bail early via an implicit `Task.checkCancellation()`).

**Residual gap, stated honestly:** these tests still don't drive the actual `transcribeMultichannelPipeline`/`transcribeMicrophoneOnlyPipeline` async functions end-to-end through a real STT/diarization pipeline cancelled at each checkpoint — that would need heavy scaffolding (blocking stubs injected at each of the six checkpoints) disproportionate to this fix. What's covered now: every registration helper is exercised as the actual production code (not reimplemented), including the one that needs a live manager (`cancelSpeakerNamingRequest`) and the one that needs the superseded-task branch (`commitSavedTranscriptSideEffectsUnlessCancelled`). The remaining gap is purely "does the pipeline call these helpers at the right point in the right order," which is now a smaller, more auditable surface (six call sites, all shown in the parameter-mapping table in the PR description) rather than a large parameter-threading surface.

**3. Harness didn't prove its stated checkpoint boundary (P3)**

`runCheckpointSequence` previously created every stage's files unconditionally (only *registration* was gated by `stage`), so an early-checkpoint test never proved later-stage artifacts didn't exist — it just happened not to check them. Rewrote it with early returns per stage, so requesting `upTo: 1` genuinely never creates the audio/clips directories at all, `upTo: 2` never creates the clips directory, etc. Added explicit `XCTAssertFalse(fileExists(...))` assertions in the stage 1/2/3 tests proving the later-stage paths are absent, not just unregistered.

**Verification (all green, foreground, in order):**
1. `bash build-deps.sh --force`
2. `bash build.sh --no-open`
3. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12607 tests, 0 failed
4. `swift test` — 895 tests (13 pre-existing skips), 0 failures — includes 16 tests in `PipelineRollbackRegistryTests` and 3 new tests in `PipelineRollbackRegistryManagerTests`

Not merged, per instructions.

PR #1630, opened 2026-08-04
Unify speaker rename/merge/discard behind SpeakerIdentityMutationService

r3dbars:
## The problem

"Write a speaker identity change to both the speaker DB and saved transcripts" was implemented three different ways, each patched independently over five hardening commits (1816ff34, 09d7a9b2, 12a69280, 5412512e, 34730890):

1. **`SpeakerNamingCoordinator.handleNamingComplete`** (per-meeting naming flow) — rewrites the just-finalized transcript first (it has the `TranscriptionResult` in memory), then applies a DB mutation batch inside a real SQL transaction, and rolls the transcript back to its pre-rewrite snapshot if the DB step throws. **This was the only one of the three with any rollback at all.**
2. **`SpeakerPeopleSettingsViewModel.rename`** (Settings > People) — wrote the DB change first, then scanned every saved transcript on disk and best-effort rewrote each one, logging and continuing past per-file failures. A DB write that succeeded while a later transcript write failed left the two stores permanently disagreeing about the speaker's name.
3. **The merge path** (`SpeakerPeopleSettingsViewModel.merge` + `SpeakerProfileMergeSideEffectCoordinator`) — DB merge, then clip promotion/deletion, then a separate un-rolled-back transcript scan (`TranscriptSaver.retroactivelyMergeSpeaker`). Its own ordering, its own subset of side effects.

## The fix

New Core service: `Sources/TranscriptedCore/Speaker/SpeakerIdentityMutationService.swift`, with three intents — `.rename`, `.merge`, `.discard` — and one canonical sequence:

**plan affected transcripts → snapshot their content → rewrite them (all-or-nothing per call) → commit the DB change inside `SpeakerStore.performMutationBatch`'s real SQL transaction → restore every rewritten transcript to its snapshot if that transaction throws → run clip side effects only after both stores agree.**

### Deliberate ordering deviation from the literal task spec

The task description ordered the canonical steps as "DB txn → rewrite transcripts → ... → roll back DB (or restore snapshots) on failure." I implemented the opposite order — **transcripts first, DB transaction last** — deliberately, because it's what `SpeakerNamingCoordinator` (explicitly cited as "the only one with rollback" / the model to follow) already does, and because it's the only ordering that gets a *free, exact* rollback out of the primitives that exist today:

- `SpeakerStore.performMutationBatch` already wraps its closure in `BEGIN EXCLUSIVE ... COMMIT`/`ROLLBACK`, so a DB step that throws is undone by SQLite itself — nothing extra to build.
- A transcript rewrite has no equivalent undo once a DB **merge** has *committed*: there is no way to losslessly reverse "the source profile's row was deleted and its embedding blended into the target" (even `unmergeMostRecent` only approximately reconstructs it from stored history). Restoring a transcript from an in-memory string snapshot, by contrast, is always exact.

So the step with a free, exact rollback (the DB write) goes last; the step whose rollback has to be hand-rolled (the transcript write) goes first, using a snapshot that's always safe to restore. This is not a new design — it's `SpeakerNamingCoordinator`'s existing pattern, generalized from one transcript to N scanned transcripts and extended to cover merge's clip side effects.

### Caller migrations

- **`SpeakerPeopleSettingsViewModel.rename`** now calls `SpeakerIdentityMutationService.apply(.rename(...))` instead of writing the DB directly then calling `TranscriptSaver.retroactivelyUpdateSpeaker`.
- **`SpeakerPeopleSettingsViewModel.merge`** now calls `SpeakerIdentityMutationService.apply(.merge(...), clipSideEffects:)` instead of `SpeakerProfileMergeSideEffectCoordinator.merge` + a separate `TranscriptSaver.retroactivelyMergeSpeaker` scan. `SpeakerProfileMergeSideEffectCoordinator` (and its test) is deleted — fully superseded, would otherwise be dead code.
- **`SpeakerNamingCoordinator`** (path 1) is **intentionally not re-routed** onto the new service. It operates on one already-open transcript with the `TranscriptionResult` in memory, applying several heterogeneous update kinds (`named`/`confirmed`/`merged`/`corrected`/`collapsedToMe`/`discardedFromDatabase`) inside *one* atomic transcript+DB transaction. Forcing that shape onto a service designed around "one intent → scan the saved-transcript library" would either break its existing atomicity guarantee or require a second, riskier refactor of an already deeply-tested 900-line file, for no behavior change (it's already correct — it's the model, not a bug to fix). Its DB-mutation vocabulary (`setDisplayName`+`resetDisputeCount`, `mergeProfiles`+`resetDisputeCount`, `restoreProfile`+`incrementDisputeCount`) intentionally matches the new service's vocabulary exactly, so a future full unification has no semantic gap to close.

### Engine census (which "speaker rename inside saved transcript Markdown" engines exist, and what happened to each)

| Engine | Used by | Disposition |
|---|---|---|
| `TranscriptSaver.applyRetroactiveRename` (frontmatter-row-scoped, name-text-based, skips ambiguous channels but still writes YAML + non-conflicting channels) | rename (path 2) + merge (path 3) — already shared before this PR | **Kept as the one library-scan engine.** Now called from `SpeakerIdentityMutationService` instead of directly from `TranscriptSaver.retroactivelyUpdateSpeaker`/`retroactivelyMergeSpeaker` (both of which still exist as thinner wrappers for `TranscriptStorage` protocol conformance and are no longer called by any in-app UI path). Also gained a companion `discardSpeakerDatabaseLink(in:dbId:)` for the new discard intent. |
| `TranscriptSaver.updateSpeakerNames(transcriptURL:updates:transcriptionResult:)` (utterance-position-based, fails closed entirely on ambiguity) | `SpeakerNamingCoordinator` only | **Kept, not unified.** Requires the in-memory `TranscriptionResult` from the transcription that just finished — data that does not exist for historical transcripts being scanned from disk. Not interchangeable with the library-scan engine; "pick the strictest" is scoped to engines that could actually replace each other, and these two structurally can't swap. |
| `TranscriptSaver.updateDeferredSpeakerName` (single-transcript, scoped by diarizer id + channel + db id) | `SpeakerPeopleSettingsViewModel.namePendingReviewItem` (deferred-review "voices to name" naming — a related but distinct flow, not one of the 3 audited call sites) | **Left as-is, out of scope.** Cut documented here rather than attempted, per the task's own scope-cut allowance. |
| `TranscriptSaver.updateSpeakerNames(transcriptURL:updates:speakerStoreForIndex:)` ("generic" overload) | **No production callers found** (grep across `Sources/`, `Tools/`) — only exercised by its own test | **Left as dead code, flagged here.** Deleting it is a separate, low-risk cleanup outside this PR's scope. |

I did **not** change `applyRetroactiveRename`'s existing same-channel-shared-name ambiguity handling (it already fails closed for the ambiguous *body* text while still correcting the unambiguous YAML row — this exact behavior is pinned by `testRetroactivelyUpdateSpeakerFailsClosedForSameChannelSharedName`, a deliberate, already-strict design decision, not the looser one the task's audit note was pointing at).

### Bug fix along the way

`SpeakerDatabase.deleteSpeaker` was the only DB mutator that didn't check `isExecutingOnQueue` before calling `queue.sync` — every sibling (`setDisplayName`, `restoreProfile`, `mergeProfiles`, `incrementDisputeCount`, ...) does. Calling it from inside `performMutationBatch`'s closure (needed for the discard-without-snapshot case) deadlocked/trapped on a nested `queue.sync` onto the same serial queue. Fixed by adding the same guard every other mutator has. Existing callers (`SpeakerNamingCoordinator`, `SpeakerPeopleSettingsViewModel.delete`) call it outside any transaction, so this is additive — it makes `deleteSpeaker` composable inside a batch without changing its behavior for anyone who wasn't already doing that.

## Tests

`Tests/TranscriptedCoreTests/SpeakerTests/SpeakerIdentityMutationServiceTests.swift` (SpeakerTests SPM target):

- `testRenameHappyPathUpdatesDatabaseAndTranscript`
- `testRenameWithOneTranscriptWriteFailureRestoresSnapshotsAndLeavesDatabaseUntouched` — pins the new atomicity: forces a real I/O failure (a read-only directory) on the second of two affected transcripts, asserts the DB was never mutated and the first (already-rewritten) transcript is restored to its exact snapshot.
- `testMergeAppliesClipSideEffectsOnlyAfterDatabaseCommit` — asserts clip side effects fire exactly once, after the DB commit, with the right ids.
- `testMergeRestoresTranscriptsWhenDatabaseMergeFails` — merges into a nonexistent target id, asserts the transcript is restored and clip side effects never ran.
- `testDiscardRestoresMatchedProfileSnapshotAndClearsTranscriptLink`
- `testDiscardWithoutSnapshotDeletesProfileAndInvokesClipCleanup`

All pass, alongside the full existing `RetroactiveSpeakerUpdaterTests` and `SpeakerNamingCoordinatorTests` suites (untouched behavior, confirmed green).

## Verification (foreground, in order, all green on this branch)

1. `bash build-deps.sh --force`
2. `bash build.sh --no-open`
3. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12605 tests, 0 failures
4. `bash run-integration-smoke.sh` — core + wake + package smoke green
5. `swift test` — 882 tests, 13 skipped (pre-existing hardware-gated skips), 0 failures

`scripts/entrypoints/run-tests.sh`'s `APP_SOURCES` manifest was updated to drop the deleted `SpeakerProfileMergeSideEffectCoordinator.swift` entry (the fast-test runner's source list is curated, unlike `build.sh`'s `find`-based `APP_SOURCE_FILES`).

## Deviations / concerns

- `SpeakerNamingCoordinator` (path 1) is not re-routed onto the new service — see "Caller migrations" above for the reasoning. Its DB-mutation vocabulary now matches the service's exactly, so this is a clean seam for a future follow-up if full unification is still wanted.
- Two of the ~4 "speaker rename inside transcript Markdown" engines identified in the audit (`updateDeferredSpeakerName`, and the unused `speakerStoreForIndex` overload) are intentionally left untouched — see the engine census table.
- No production caller currently exercises the new `.discard` intent (the only existing discard flow is `SpeakerNamingCoordinator`'s in-flow discard, which stays on its own path per above); it's covered by tests as a general-purpose primitive per the task's requested test matrix, ready for a future Settings-side "discard" affordance if one is added.


r3dbars:
Addressed all 5 findings from the Codex review in `7c7de8e7`.

**P1a** (planning silently skips unreadable-but-flagged transcripts) — `planAffectedTranscripts` now tracks files where the byte pre-scan couldn't rule out a match but the full read failed, and throws `MutationError.transcriptPlanningReadFailed(fileCount:)` instead of `continue`-ing past them. Nothing is mutated at that stage, so this is a clean abort. New test: `testPlanningReadFailureAbortsBeforeAnyMutation` (chmods a flagged transcript to `0o000`, asserts the throw and that the DB is untouched).

**P1b** (`deleteSpeaker` swallows SQL failures inside a mutation batch) — `deleteSpeakerImpl` now calls `recordMutationFailure` on both the prepare and step failure branches, matching every sibling mutator (`setDisplayNameImpl`, `restoreProfileImpl`, ...). A delete failure now correctly propagates through `performMutationBatch`'s throw path, which the service's existing catch-and-rollback logic already handles. New test: `testDiscardDeleteFailureRollsBackTheBatch` (denies `DELETE` on `speakers` via `sqlite3_set_authorizer`, asserts the row survives and the transcript's DB link is restored).

**P2a** (restore failures are log-only) — added `restoreOrThrow`: retries each failed restore once, and if any file is still unrestorable, throws `MutationError.transcriptRestoreFailed(fileCount:)` (a count only, never a path or name) instead of returning silently. Both Settings callers (`rename`, `merge`) now `do/catch` around `apply(_:)` and forward this specific case to `EventReporter.shared.capture(...)` with just a profile id (opaque UUID) and file count — no transcript text, names, or paths. The doc comment on `restoreOrThrow` calls out the residual disk-full/permanently-broken-mount reality explicitly, since no in-process retry can fix that.

A true end-to-end filesystem repro of "the rewrite to a path succeeded, then the later restore to that identical path fails" turned out not to be constructible with static POSIX permissions — both operations have identical requirements on the identical path, so there's no way to make one succeed and the other fail without code running *between* them, which a single synchronous `apply()` call doesn't allow. I loosened `PlannedRewrite` and `restoreOrThrow` from `private` to internal (visible to `@testable import` only) so `testRestoreOrThrowSurfacesADistinctErrorWhenRestoreFailsAfterRetry` and `testRestoreOrThrowSucceedsWhenEveryFileCanBeRestored` exercise the retry/throw mechanism directly instead. Flagging this in case a different repro approach is preferred.

**P2b** (stale-name race against Coordinator) — the `getSpeaker` reads in `applyMerge` already run inside the same `serializeTranscriptFileUpdate` lock the whole `apply(_:)` call holds, which is the same lock `SpeakerNamingCoordinator` holds around its own transcript-write-then-DB-mutation block — so those two can't interleave; one fully precedes or fully follows the other. Added a doc comment at the read site being explicit about the one residual gap I did *not* close: Coordinator's collapse/discard DB side effects (`restoreCollapsedMatchedSpeakers`, `applyDiscardedSpeakerActions`, the collapsed-mic-profile delete loop) run *after* that lock is released, on a background queue — a merge landing in that specific window could still read a display name Coordinator is about to overwrite or restore. Closing that fully means moving those DB writes inside Coordinator's lock, which is a change to a separately deeply-tested 900-line file I don't think belongs in this PR (per the same reasoning as the "SpeakerNamingCoordinator not re-routed" scope cut already in the PR description). No test requested for this one per your list; happy to add a targeted one if you want the specific gap pinned rather than just documented.

**P2c** (same-id merge) — `apply(_:)` now checks `sourceId == targetId` up front and throws `MutationError.sameSourceAndTarget(profileId:)` before acquiring the transcript lock or touching anything. The Settings `merge` caller's `source.id != target.id` guard is now explicitly documented as defense-in-depth, not the only guard. New test: `testMergeWithSameSourceAndTargetThrowsBeforeAnyMutation`.

**API change**: `SpeakerIdentityMutationService.apply(_:)` is now `throws`. Both Settings callers (`rename`, `merge`) updated to `do/catch`; ordinary recoverable failures (a transcript write failed and was cleanly rolled back, or the DB transaction threw and rollback succeeded) still surface as `Outcome(succeeded: false, ...)` as before — only the four failure modes above throw.

Verification (foreground, all green on `claude/w1-speaker-mutation-service`):
1. `bash build-deps.sh --force`
2. `bash build.sh --no-open` — signs cleanly
3. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12605 tests, 0 failures
4. `bash run-integration-smoke.sh` — green
5. `swift test` — 887 tests (+5 net new), 13 pre-existing skips, 0 failures

PR #1632, opened 2026-08-04
Make meeting capture attempt bookkeeping and local summary telemetry compiler-checked MainActor

r3dbars:
## Summary

Two guard types were safe only by convention (no lock, no actor annotation). This makes both compiler-checked.

### 1. `MeetingCaptureAttempt<Output>` (Sources/Meeting/MeetingCaptureSupport.swift)

Marked `@MainActor`. Every real caller is `MeetingCaptureBridge` (`@MainActor`), including its `deinit`, which was **not** MainActor-isolated before this change — a latent hazard if the bridge ever deinited off-main.

**Waiter coalescing** (updated after review — see PR comment for the full story): `MeetingCaptureAttempt` now supports multiple waiters on one in-flight attempt via `joinIfActive(_ continuation:) -> Bool`. `reset()`/`resetIfCurrent(_:)` return `[CheckedContinuation<Output, Never>]` (every continuation that must resolve to the *same* value) instead of a single optional. The start path's `begin()`/`reset()` single-waiter semantics are untouched; only `completionAttempt`'s stop path ever calls `joinIfActive`.

**Teardown paths and their continuation guarantees** (both `startAttempt` and `completionAttempt` are `MeetingCaptureAttempt` instances; every path below now resumes *every* returned continuation, not just a primary one):

- `MeetingCaptureBridge.deinit` (now `isolated deinit`) — unconditionally resumes every pending `startAttempt` continuation (`false`) and every pending `completionAttempt` continuation (best-effort `CaptureStopResult`), cancels any outstanding timed-out-stop expiry tasks. `isolated deinit` (Swift 6.1+, available on this toolchain — `swiftc --version` shows Swift 6.3.3 — without any extra language-mode flag; verified with a standalone `swiftc -typecheck` repro before adopting it) keeps this on MainActor instead of relying on every release site happening to run there.
- `startRecording()` — resets a stale `completionAttempt` left over from a previous session at the top; installs a fresh `startAttempt` guarded by a MainActor timeout `Task` that resumes+resets via `resetIfCurrent` (token-guarded, so it can't double-resume if `finishPendingStartAttemptIfPossible()` already won the race).
- `finishPendingStartAttemptIfPossible()` (driven by Combine subscriptions on `audio.$isRecording`/`$error`/etc.) — resumes+resets `startAttempt` on ready/failed.
- `stopAndAwaitFiles()` — if a stop is already in flight, **joins** it via `joinIfActive(_:)` instead of starting a new one (no new `begin()`, no new `audio.stop()`, no new token/generation). Otherwise installs `completionAttempt`, guarded by a MainActor timeout `Task` (`resetIfCurrent`-guarded, resumes every waiter with the same timed-out result) and by `wireCallbacks()`'s `onRecordingCompleteWithGeneration` handler (resumes every waiter with the same real result on `.expectedStop` disposition). The join check runs *before* consulting `audio.isRecording`, because `Audio.stop()` flips `isRecording` false from a `DispatchQueue.main.async` block queued before WAV finalization even starts — `completionAttempt`'s own active/inactive state, not `audio.isRecording`, is the source of truth for "is a stop already outstanding."
- **Original P1 (fixed)**: a first pass at this fix resumed a displaced `completionAttempt` continuation with a fabricated `currentStopResult()`, which could report still-finalizing audio as complete and cause the real completion to be misrouted once a second `begin()` advanced the attempt's own token. Replaced with waiter coalescing so two overlapping stop callers always observe the one true result of the one underlying `audio.stop()` call. `Tests/MeetingCaptureAttemptTests.swift` covers coalescing via both `resetIfCurrent` (real-completion path) and `reset()` (every teardown/timeout path), plus a regression test documenting that `begin()` itself still silently displaces if a caller calls it without checking `joinIfActive` first (by design — `joinIfActive` is the guard, not `begin()`).

**Primitive adoption**: re-expressed `MeetingCaptureAttempt`'s `attemptID: UUID?` + `continuation` pair using `SupersessionEpoch` + `ClaimSlot` from PR #1631 (`Sources/TranscriptedCore/Utilities/SupersessionEpoch.swift`). Verified this is a clean drop-in, not a semantic change: `attemptID` was only ever minted by `begin()` and compared against `self.attemptID` on the *same* instance — never logged, persisted, or compared across bridge instances — so the UUID's global uniqueness was never load-bearing.

### 2. `LocalSummaryAttemptTelemetry` (Sources/UI/Settings/LocalSummaryAttemptTelemetry.swift)

Verified every real usage site first (grep): only `TranscriptedSettingsView.generateLocalSummary(...)`. It's constructed synchronously, `.intent()` is called synchronously right after, and every `.terminal()` call is inside `Task { @MainActor in ... }` — the `Task.detached` background work only touches the already-NSLock-protected sibling `LocalSummaryTerminalOutcomeArbiter`, never this type. All MainActor — annotated `@MainActor` to match, giving it the same compiler-checked safety story as its sibling's NSLock (different mechanism, same guarantee, since the sibling actually is touched from the detached task).

This broke the standalone subprocess test harness in `Tests/LocalSummaryAttemptTelemetryTests.swift`, whose `private static` helper functions (called from `Harness.main()`, not inlined into it) lost the implicit MainActor isolation `main()` itself gets. Fixed by marking the harness `struct Harness` `@MainActor`.

## Test plan

- [x] `bash build-deps.sh --force` (stamp mismatch on fresh worktree)
- [x] `bash build.sh --no-open` — clean build, no new warnings
- [x] `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12684 tests, 0 failed
- [x] `bash run-integration-smoke.sh` — pass
- [x] `swift test` — 946 tests, 13 skipped, 0 failures
- [x] `Tests/MeetingCaptureAttemptTests.swift` covers begin/reset/resetIfCurrent semantics, cross-token rejection, timeout-task cancellation on reset, `joinIfActive` coalescing (both the resetIfCurrent and reset() paths), the false-when-inactive contract, and the begin()-without-reset displacement hazard

Not merging per instructions.

### DIFF
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -76,7 +76,17 @@ final class MeetingCaptureBridge: ObservableObject {
         wireSubscriptions()
     }
 
-    deinit {
+    // `isolated deinit` (available on this toolchain without any extra
+    // language-mode flag — verified with `swiftc -typecheck` against a
+    // standalone repro) keeps this teardown compiler-checked on MainActor
+    // instead of relying on every release site happening to run there.
+    // `MeetingCaptureAttempt` is itself @MainActor now, so a plain
+    // (nonisolated) deinit could no longer touch `startAttempt`/
+    // `completionAttempt` at all; isolating the whole deinit is simpler than
+    // splitting teardown into an explicit `invalidate()`/`shutdown()` call
+    // for a bridge whose only owner (`MeetingSessionController`) doesn't
+    // need bespoke teardown sequencing before releasing it.
+    isolated deinit {
         timedOutStopCompletionExpiryTasks.values.forEach { $0.cancel() }
         startAttempt.reset()?.resume(returning: false)
         completionAttempt.reset()?.resume(returning: CaptureStopResult(
@@ -159,6 +169,16 @@ final class MeetingCaptureBridge: ObservableObject {
         guard audio.isRecording else {
             return currentStopResult()
         }
+        // A second overlapping stop call (e.g. two callers racing to stop the
+        // same recording before Core's completion callback fires) must not
+        // silently displace a still-pending completion's continuation —
+        // `MeetingCaptureAttempt.begin()` overwrites unconditionally, so the
+        // caller is responsible for resolving any outstanding attempt first.
+        // `startRecording()` already does the equivalent reset for a stale
+        // completion attempt left over from a previous recording; mirror it
+        // here so a concurrent stop can't leak a continuation and hang its
+        // caller forever.
+        completionAttempt.reset()?.resume(returning: currentStopResult())
         let stopTimeout = TranscriptedConstants.meetingStopTimeout(
             forRecordingDuration: max(recordingDuration, audio.recordingDuration)
         )
--- a/Sources/Meeting/MeetingCaptureBridge.swift
+++ b/Sources/Meeting/MeetingCaptureBridge.swift
@@ -88,12 +88,17 @@ final class MeetingCaptureBridge: ObservableObject {
     // need bespoke teardown sequencing before releasing it.
     isolated deinit {
         timedOutStopCompletionExpiryTasks.values.forEach { $0.cancel() }
-        startAttempt.reset()?.resume(returning: false)
-        completionAttempt.reset()?.resume(returning: CaptureStopResult(
+        for continuation in startAttempt.reset() {
+            continuation.resume(returning: false)
+        }
+        let stopResult = CaptureStopResult(
             micURL: audio.micAudioFileURL,
             systemURL: audio.systemAudioFileURL,
             didTimeOut: false
-        ))
+        )
+        for continuation in completionAttempt.reset() {
+            continuation.resume(returning: stopResult)
+        }
         // Combine cancellables auto-release. Audio's own deinit tears down CoreAudio.
     }
 
@@ -103,7 +108,10 @@ final class MeetingCaptureBridge: ObservableObject {
     /// active until `stopAndAwaitFiles()` is called.
     func startRecording() async -> Bool {
         expectedStopGeneration = nil
-        completionAttempt.reset()?.resume(returning: currentStopResult())
+        let staleStopResult = currentStopResult()
+        for continuation in completionAttempt.reset() {
+            continuation.resume(returning: staleStopResult)
+        }
         if audio.isRecording { return true }
 
         // Keep the immediately preceding timed-out stop across this start so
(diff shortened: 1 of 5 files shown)


r3dbars:
Addressed the P1 from the Codex review.

## The fix

Replaced displacement with waiter coalescing, as requested:

- `MeetingCaptureAttempt<Output>` gained `joinIfActive(_ continuation:) -> Bool`, which appends an additional waiter to an already-in-flight attempt instead of starting a new one. `reset()`/`resetIfCurrent(_:)` now return `[CheckedContinuation<Output, Never>]` (every continuation that must resolve to the *same* value — primary plus every joined waiter) instead of a single optional. The start path's `begin()`/`reset()` single-waiter semantics are untouched — only `completionAttempt`'s stop path ever calls `joinIfActive`.
- `MeetingCaptureBridge.stopAndAwaitFiles()` now joins an in-flight attempt when one exists; it only calls `begin()`/`audio.stop()` when starting fresh. Because a joined caller never re-invokes `begin()`/`audio.stop()`, there's no second `stopGeneration`/token to race — the "real completion classified as stale" failure mode you flagged is eliminated structurally, not just guarded against.

## One thing I caught while implementing this that the original review didn't call out

The join check has to run **before** consulting `audio.isRecording`, not after. `Audio.stop()` flips `isRecording = false` from a `DispatchQueue.main.async` block that's queued *before* the slow CoreAudio teardown/WAV finalization even starts (see `Audio.stop()`'s "unfreeze the UI immediately" comment) — so a second `stopAndAwaitFiles()` call landing in that (real, if narrow) window would see `isRecording == false` and fall through the pre-existing early-return path (`guard audio.isRecording else { return currentStopResult() }`), skipping the join check entirely and hitting the same premature-success bug through a different door. I moved the `audio.isRecording` guard inside the `withCheckedContinuation` closure, after the `joinIfActive` check, so `completionAttempt`'s own active/inactive state — not `audio.isRecording` — is the source of truth for "is a stop already outstanding."

## Tests

Added to `Tests/MeetingCaptureAttemptTests.swift`:
- `joinIfActive coalesces overlapping callers onto the same result via resetIfCurrent` — mirrors the real-completion path (bridge's `.expectedStop` branch)
- `joinIfActive coalesces overlapping callers onto the same result via reset` — mirrors every teardown/timeout path
- `joinIfActive returns false when nothing is in flight` — the not-active contract
- `begin() clears any stale joined waiters from a prior attempt` — defensive coverage
- Updated all pre-existing suites for the new `[CheckedContinuation]`-returning API

I could not add a bridge-level integration test for "the late-completion path still runs (not classified stale)" — there's no existing mock/fake for `TranscriptedCore.Audio` in this repo's test infra (it's a real CoreAudio-backed class, and `MeetingCaptureBridge` isn't in `run-tests.sh`'s flat-compile list since it needs the full `@MainActor`/AppKit app target). The claim is structural rather than test-proven: since a joined caller never mints a second token/generation, there is nothing left that *could* be classified stale. Flagging this so a reviewer can judge whether that structural argument is sufficient or whether it's worth building bridge-level test scaffolding in a follow-up.

## Verification (all foreground)
- `bash build.sh --no-open` — clean, no new warnings
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12684 tests, 0 failed (12 new vs. previous push)
- `bash run-integration-smoke.sh` — pass

Not merging.

PR #1634, opened 2026-08-04
Collapse meeting session state to one source of truth

r3dbars:
## Summary

`MeetingSessionController` answered "what is the meeting doing right now" with five overlapping signals that could disagree with each other during real windows (mid-start, mid-stop):

1. `state` (the `State` enum) — written from ~24 sites in the controller plus ~6 sites in `TranscriptionQueueCoordinator` via a `setState`/`setDisplayStatus` seam.
2. `@Published isRecording` — a Combine mirror of `capture.$isRecording`, read by `TranscriptedApp.swift` (menubar/quick actions/force-quit gating) while `MeetingOverlayController` gated on `state` — two consumers reading two fields meant to mean the same thing.
3. `isCaptureSessionActive` — `if case .recording = state { true } else { isRecording }`, an OR of the two.
4. `isStartingRecording`/`isFinishingRecording` — out-of-band phase booleans bolted on because `State` had no transitional cases; `shouldConfirmQuitForActiveCapture`, `shouldBlockDictationForActiveMeetingCapture`, `canShareMicWithDictation`, `hasRuntimeDiagnosticsWork`, and the mic-boost-prompt gate each independently hand-ORed these onto `isCaptureSessionActive`.
5. `displayStatus` — a Combine mirror of `taskManager.$displayStatus` plus ~4 manual assignments for phases Core doesn't know about.

This PR collapses all five onto `state` as the single source of truth, with two new transitional states (`.startingRecording`, `.stoppingRecording`) closing the gaps that made the old booleans necessary in the first place.

## The design

- **`State` extracted to its own file.** `Sources/Meeting/MeetingSessionState.swift` is a dependency-free (Foundation-only) top-level `MeetingSessionState` enum; `MeetingSessionController.State` is now a `typealias` onto it. `Sources/Meeting/MeetingSessionStateMachine.swift` holds the pure legal-transition table and two query helpers (`isCaptureSessionActive`, `isSteadyStateRecording`), covered by `Tests/MeetingSessionStateMachineTests.swift`. Case names are additive/stable (state is logged by name).
- **`isRecording` is computed from `state`** (`MeetingSessionStateMachine.isSteadyStateRecording`). The old `capture.$isRecording` sink no longer writes it — it's an event handler now: it closes the `.startingRecording -> .recording` gap the instant capture confirms the mic engaged (idempotent — the synchronous `transition(to: .recording, ...)` in `startRecording()` usually wins the race, this is a safety net either way), and still runs cleanup (dismiss boost prompt/warnings) when capture stops. It deliberately does **not** drive the failure transition for an unexpected stop — that stays in `handleUnexpectedCaptureStop(_:)`, driven by `capture.onUnexpectedRecordingComplete`, because only that callback carries the real `CaptureStopResult` (audio URLs) needed to preserve a failed meeting.
- **One transition writer.** `MeetingSessionController.transition(to:reason:)` is now the only place `state` is assigned (mechanical migration of every prior `state = ...` site). In DEBUG builds it *logs* (does not crash) a transition `MeetingSessionStateMachine.isLegalTransition` doesn't recognize — logging rather than asserting because the real hand-written machine is permissive by design (error reporting must never be blocked by a table miss).
- **Coordinator reports outcomes, doesn't drive state.** `TranscriptionQueueCoordinator`'s `setState`/`setDisplayStatus` calls are gone, replaced by four narrow controller methods: `transcriptionJobDidStart()`, `transcriptionJobFailedToPrepare(message:)`, `transcriptionWorkContinues()`, `transcriptionQueueSettled()`. The controller's `transition`/`updateDisplayStatus` own the actual writes.
- **`displayStatus` has one writer**, `updateDisplayStatus(_:source:)`. `source` only distinguishes the two effectively-different callers that already existed: the `taskManager.$displayStatus` mirror also runs `handleDisplayStatusChange` (the only place that ever did), controller/coordinator-driven phase updates don't. Same effective behavior (last write wins, in call order), one function instead of three write sites.
- **`MeetingMicBoostPromptPolicy` simplified.** It used to take `isRecording` (capture mirror), `isFinishingRecording`, and `sessionStateIsRecording` as three separate params specifically to catch a stale cue mid-stop when the mirror and state machine disagreed. Now that `isRecording` *is* `state == .recording`, those three collapse into one boolean — the policy takes just `isRecording` now. Updated its fast test (`Tests/MeetingMicBoostPromptPolicyTests.swift`) to match.

## Signal migration table

| Old signal | Replacement |
|---|---|
| `state` (24 sites in controller) | `transition(to:reason:)` — single writer |
| `state` (6 `setState` sites in `TranscriptionQueueCoordinator`) | `transcriptionJobDidStart()` / `transcriptionJobFailedToPrepare(message:)` / `transcriptionWorkContinues()` / `transcriptionQueueSettled()` |
| `@Published isRecording` (Combine mirror of `capture.$isRecording`) | computed `isRecording` (`state == .recording`) |
| `isCaptureSessionActive` (`state==.recording \|\| isRecording`) | `MeetingSessionStateMachine.isCaptureSessionActive(state)` — true for `.startingRecording`/`.recording`/`.stoppingRecording` |
| `isStartingRecording` (stored bool, zero external readers) | kept as `startRecordingCallInFlight`, a narrower, explicitly-scoped reentrancy guard — see "Not migrated" below |
| `isFinishingRecording` (stored bool, ~15 sites) | fully deleted; replaced by `.stoppingRecording` state (persists through the entire stop/cancel/termination window, no nested state churn to fight with) |
| `shouldConfirmQuitForActiveCapture` = `isCaptureSessionActive \|\| isFinishingRecording` | `isCaptureSessionActive` alone (the `isFinishingRecording` OR was always redundant — under the old design `state` stayed `.recording` for the *entire* stop window, so `isCaptureSessionActive` was already true whenever `isFinishingRecording` was) |
| `shouldBlockDictationForActiveMeetingCapture` = same OR | `isCaptureSessionActive` alone, same reasoning |
| `canShareMicWithDictation` = `isCaptureSessionActive && !isFinishingRecording` | `isRecording` (`state == .recording` alone already excludes starting/stopping) |
| `hasRuntimeDiagnosticsWork` = `isCaptureSessionActive \|\| isFinishingRecording \|\| hasBackgroundTranscriptionWork` | `isCaptureSessionActive \|\| hasBackgroundTranscriptionWork` |
| `displayStatus` (3 write sites: taskManager mirror, controller manual assignments, coordinator `setDisplayStatus`) | `updateDisplayStatus(_:source:)` — single writer |

### `$isRecording` Combine subscribers (task explicitly called these out)

Grepped every `\$isRecording` subscriber across `Sources/`. Two, both in `TranscriptedApp.swift` (status-item recording indicator, activation-policy force-quit gating) — migrated to `appState.meetingSession.$state.map { MeetingSessionStateMachine.isSteadyStateRecording($0) }.removeDuplicates()`.

## Legal-transition table

(Full comment lives in `MeetingSessionStateMachine.swift`; two blanket rules — any state to itself, and any state to `.error` — sit on top of this table because the current code has no state-specific guard before surfacing a fatal error, and several callers are written to be safe idempotent no-ops.)

```
idle            -> loadingModels
loadingModels   -> ready
ready           -> idle | loadingModels | startingRecording | transcribing
startingRecording -> recording
recording       -> stoppingRecording
stoppingRecording -> transcribing | ready
transcribing    -> ready
error           -> idle | loadingModels | ready | transcribing
(any state -> itself, any state -> error)
```

## Behavior changes (intentional, small, documented)

The bar was behavior preservation; these are the specific, narrow spots where closing the desync window changes an edge case, each low-risk:

- **Force-quit gating now covers the mic-engage window.** Previously `isStartingRecording` was never read by `shouldConfirmQuitForActiveCapture` at all, so quitting during the (usually sub-second) permission-check-to-mic-engage window wasn't gated. `isCaptureSessionActive` now includes `.startingRecording`, closing that gap.
- **`MeetingMicBoostPromptPolicy`'s three-way stale-cue check collapses to one boolean**, because the specific desync it was defending against (`isRecording: true` but `sessionStateIsRecording: false`) can no longer happen — that *is* the point of this PR.

## Not migrated (documented exception)

`startRecording()`'s whole-function reentrancy guard (renamed `startRecordingCallInFlight`, still a private stored `Bool`) is **not** derived from `state`. The permission-check + model-prep preamble inside `startRecording()` legitimately cycles `state` through `.loadingModels`/`.ready`/`.error` via the shared `prepareModels()` path — a second concurrent `startRecording()` call would see one of those "free" values and slip past a `state`-only guard, defeating the reentrancy check (or worse, letting two calls both reach `capture.startRecording()`). `.startingRecording` is used for the narrower, unambiguous window right before `capture.startRecording()` instead. `isStartingRecording` had zero external readers before this change (confirmed via grep), so nothing else needed migrating.

## Verification

All run foreground, in order, against this branch:

- `bash build-deps.sh --force` — green
- `bash build.sh --no-open` — green (two pre-existing warnings confirmed present on `origin/main` unchanged — `MeetingSessionController.swift:2712` unused `remove(for:)` result, `DictationSessionController.swift:1010` unused `Bool` — neither touched by this change)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12720/12720 passed (one source-text regression-guard test, `FailedMeetingPresentationTests.swift`, updated to match the new `transition(to:reason:)` call shape it was literally grepping the source for)
- `bash run-integration-smoke.sh` — green

### DIFF
--- a/Sources/Meeting/MeetingMicBoostPromptPolicy.swift
+++ b/Sources/Meeting/MeetingMicBoostPromptPolicy.swift
@@ -18,20 +18,20 @@ enum MeetingMicBoostPromptPolicy {
     /// the user already enabled Apple voice processing in Settings.
     ///
     /// Invariant: the prompt flag is never true while nothing is recording.
-    /// A late cue can land mid-stop — capture teardown awaits file handles
-    /// while the published recording flags are still settling — so
-    /// presentation also requires that no stop/cancel/termination teardown is
-    /// in flight and that the session state machine still says recording.
+    /// A late cue can land mid-stop; `isRecording` here is expected to be the
+    /// caller's single steady-state-recording signal (state == .recording),
+    /// so it already reads false once a stop/cancel/termination teardown
+    /// starts — see MeetingSessionController.isRecording. Before the 2026-08
+    /// state collapse this took three separate signals (a capture-level
+    /// mirror, an `isFinishingRecording` flag, and the session state check)
+    /// because those could disagree with each other; now that the session
+    /// controller has one source of truth, they're the same boolean.
     static func shouldPresent(
         isRecording: Bool,
-        isFinishingRecording: Bool,
-        sessionStateIsRecording: Bool,
         voiceProcessingPreferenceEnabled: Bool,
         currentOutcome: MeetingMicBoostPromptOutcome
     ) -> Bool {
         isRecording
-            && !isFinishingRecording
-            && sessionStateIsRecording
             && !voiceProcessingPreferenceEnabled
             && currentOutcome == .notShown
     }
@@ -41,13 +41,8 @@ enum MeetingMicBoostPromptPolicy {
     /// preference or record a prompt outcome for a dead recording.
     static func shouldApplyPromptAction(
         isPromptVisible: Bool,
-        isRecording: Bool,
-        isFinishingRecording: Bool,
-        sessionStateIsRecording: Bool
+        isRecording: Bool
     ) -> Bool {
-        isPromptVisible
-            && isRecording
-            && !isFinishingRecording
-            && sessionStateIsRecording
+        isPromptVisible && isRecording
     }
 }
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -286,6 +286,30 @@ final class MeetingSessionController: ObservableObject {
         state = newState
     }
 
+    /// Reports a failure that is NOT about this session's own live capture
+    /// pipeline — a rejected user action (import/retranscribe blocked by
+    /// something unrelated to capture), a background job's outcome (queued
+    /// model-prep, a different meeting's queued transcript finishing with no
+    /// speech) — as a visible `.error`, but ONLY when capture isn't
+    /// currently live. Before the 2026-08 state collapse this guard wasn't
+    /// needed: `isCaptureSessionActive` OR'd in a capture-bridge-derived
+    /// `isRecording` mirror independent of `state`, so even a call site that
+    /// stomped `state` to `.error` for an unrelated reason left the gates
+    /// reading true because the mirror still reflected the real, physically
+    /// -still-running capture. Now that `isRecording`/`isCaptureSessionActive`
+    /// are both derived purely from `state`, an unguarded transition here
+    /// would silently clear every gate that reads them (quit-confirm,
+    /// dictation-block, mic-share, menubar, force-quit) for a capture that
+    /// is still actually running — see
+    /// `MeetingSessionStateMachine.mayReportUnrelatedFailureAsError`. When
+    /// capture is live, this is a silent no-op: the diagnostics event that
+    /// led here already ran at the call site, and the recording lifecycle
+    /// keeps driving `state` normally.
+    private func reportUnrelatedFailure(_ message: String, reason: StaticString) {
+        guard MeetingSessionStateMachine.mayReportUnrelatedFailureAsError(while: state) else { return }
+        transition(to: .error(message), reason: reason)
+    }
+
     /// `displayStatus`'s single writer. `source` only distinguishes the two
     /// callers for behavior that was already conditional on the caller: the
     /// `taskManager.$displayStatus` mirror also runs `handleDisplayStatusChange`
@@ -322,9 +346,15 @@ final class MeetingSessionController: ObservableObject {
     }
 
     /// A queued job failed before it could start (bounded model-recovery
-    /// retry gave up).
+    /// retry gave up). This job is not necessarily the active recording's
+    /// own job — a completely different meeting can still be capturing live
+    /// while an earlier queued transcript fails in the background — so this
+    /// must not force `state` to `.error` while capture is live; see
+    /// `MeetingSessionStateMachine.mayReportUnrelatedFailureAsError`.
+    /// `displayStatus` still updates unconditionally: it is a separate,
+    /// capture-independent "background work progress" signal.
     func transcriptionJobFailedToPrepare(message: String) {
-        transition(to: .error(message), reason: "transcription_job_prepare_failed")
+        reportUnrelatedFailure(message, reason: "transcription_job_prepare_failed")
         updateDisplayStatus(.failed(message: message), source: .controllerPhase)
     }
 
@@ -1477,7 +1507,21 @@ final class MeetingSessionController: ObservableObject {
     @discardableResult
     func importAudioFile(from sourceURL: URL) async -> Bool {
         guard !isCaptureSessionActive else {
-            transition(to: .error("Stop the current meeting before importing an audio file."), reason: "import_blocked_capture_active")
+            // A meeting is actively capturing (starting/recording/stopping).
+            // Rejecting the import must not force `state` to `.error` — that
+            // would silently clear the capture-active gates (quit-confirm,
+            // dictation-block, mic-share, menubar) for a recording that is
+            // still physically running; see
+            // MeetingSessionStateMachine.mayReportUnrelatedFailureAsError.
+            // The recording lifecycle keeps driving `state` normally; this
+            // is only traceable via diagnostics, not a visible session error.
+            DiagnosticsTrail.record(
+                level: .warning,
+                engine: "meeting",
+                event: "meeting_file_import_blocked_capture_active",
+                message: "Imported meeting transcription request rejected because a meeting capture is active",
+                context: baseDiagnosticsContext(extra: ["trigger": StartTrigger.fileImport.rawValue])
+            )
             return false
         }
 
@@ -1587,7 +1631,10 @@ final class MeetingSessionController: ObservableObject {
                 failureKind: failureKind,
                 modelState: state.diagnosticName
             )
-            transition(to: .error(displayMessage), reason: "import_preparation_failed")
+            // A separate recording can be actively capturing by the time
+            // this detached preparation task finishes (this call is not
+            // gated the way the entry guard above is) — do not stomp it.
+            reportUnrelatedFailure(displayMessage, reason: "import_preparation_failed")
             Self.runtimeDiagnosticsRecorder?.clearSession(kind: "meeting", outcome: "file_import_failed")
             return false
         }
@@ -1618,7 +1665,7 @@ final class MeetingSessionController: ObservableObject {
             let message = ImportedAudioQueuePersistenceFailureCopy.displayMessage(
                 preservedForRelaunch: preservedForRelaunch
             )
-            transition(to: .error(message), reason: "import_queue_persist_failed")
+            reportUnrelatedFailure(message, reason: "import_queue_persist_failed")
             updateDisplayStatus(.failed(message: message), source: .controllerPhase)
             Self.runtimeDiagnosticsRecorder?.clearSession(kind: "meeting", outcome: "import_queue_persist_failed")
             return false
@@ -1724,7 +1771,14 @@ final class MeetingSessionController: ObservableObject {
         }
         activeQueuedTranscriptionJobID = nil
         activeStoppedAudioRecovery = nil
-        transition(to: .ready, reason: "transcription_cancelled")
+        // The cancelled work is background transcription (queued/importing
+        // audio), not necessarily this recording — the Home "cancel current
+        // activity" button is gated on `displayStatus`, which reflects
+        // background work that can be visible while a different meeting is
+        // actively recording live. Must not stomp that capture's `state`.
+        if !isCaptureSessionActive {
+            transition(to: .ready, reason: "transcription_cancelled")
+        }
         DiagnosticsTrail.record(
             level: .warning,
             engine: "meeting",
@@ -1740,6 +1794,20 @@ final class MeetingSessionController: ObservableObject {
         var recordingTrigger = activeRecordingTrigger
         var stoppedFiles: (micURL: URL?, systemURL: URL?) = (nil, nil)
         var stopTimedOut = false
+
+        if isStartingRecording {
+            // Quit-confirm now fires while capture is only engaging the mic
+            // (isCaptureSessionActive covers .startingRecording, correctly
+            // widening force-quit gating to that window), so "Save Audio and
+            // Quit" can land here before startRecording()'s own
+            // `await capture.startRecording()` has resolved. Join it
+            // (bounded) instead of quitting out from under it: otherwise a
+            // recording that was about to succeed would be neither saved
+            // nor visibly recorded, breaking the dialog's promise. Once this
+            // resolves, `state` is .recording (success) or .error (failure)
+            // — startRecording() itself owns that transition synchronously.
+            await waitForRecordingStartBeforeTermination()
+        }
 
         if isStoppingRecording {
             await waitForRecordingFinishBeforeTermination()
@@ -1849,6 +1917,19 @@ final class MeetingSessionController: ObservableObject {
         }
     }
 
+    /// Bounded wait for an in-flight `startRecording()` call to resolve
+    /// before termination decides what to preserve. `capture.startRecording()`
+    /// has its own internal timeout (`TranscriptedConstants.meetingStartTimeout`,
+    /// 12s) that guarantees `state` leaves `.startingRecording` well before
+    /// this generous outer deadline; the outer bound only guards against an
+    /// unexpected hang so quit is never blocked forever.
+    private func waitForRecordingStartBeforeTermination() async {
+        let deadline = Date().addingTimeInterval(TranscriptedConstants.meetingTerminationFinishWaitTimeout)
+        while isStartingRecording && Date() < deadline {
+            try? await Task.sleep(nanoseconds: 100_000_000)
+        }
+    }
+
     private func handleUnexpectedCaptureStop(_ stopResult: CaptureStopResult) {
         // `state` alone now distinguishes this from an app-initiated stop:
         // stopRecording()/cancelRecording()/prepareForTermination() move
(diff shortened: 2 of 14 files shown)


r3dbars:
Addressed all findings from the Codex review (commit 8534d0ee). Verified each independently against the actual bridge/coordinator code before fixing — details below.

## P1a — gap-closer race (sink premature promotion)

Confirmed real: `MeetingCaptureBridge.finishPendingStartAttemptIfPossible()` only resolves `capture.startRecording()`'s continuation once `AudioCaptureStartState.meetingCaptureOutcome` sees BOTH mic and system streams actually streaming (file URL assigned + `micAudioStreaming`/`systemAudioStreaming` true), while `capture.$isRecording` mirrors `audio.isRecording`, which flips true as soon as `audio.start()` runs — a real, meaningfully-sized window where the sink's old promotion could fire before the original `startRecording()` call's own await was anywhere near resolving.

**Fix:** removed the sink's `.startingRecording -> .recording` promotion entirely. `startRecording()` alone now owns that transition, synchronously, right after its own `await capture.startRecording()` returns `true`. The sink is back to being pure cleanup (audio-inactivity bookkeeping, boost-prompt/warning reset on stop) — no state-machine writes.

## P1b — error transitions stomping live capture ownership

Confirmed, and confirmed it's a systemic pattern, not just the two examples in the review. Audited every `transition(to: .error(...))` (and the one unconditional `.ready` write) in the controller and classified each as either legitimate (about *this* session's own capture lifecycle — start/stop failing, capture dying underneath the controller) or unrelated (a rejected user action or a background job's outcome that can fire while a *different* meeting is actively capturing live). Added `MeetingSessionStateMachine.mayReportUnrelatedFailureAsError(while:)` (pure, tested) and a `reportUnrelatedFailure(_:reason:)` controller helper built on it, applied to every unrelated-failure site:

- `transcriptionJobFailedToPrepare` (the one you flagged)
- `importAudioFile`'s entry guard (the one you flagged — capture-active rejection no longer transitions state at all now, just logs a diagnostic; there's no safe message to show without stomping the very capture the guard exists to protect)
- `importAudioFile`'s two failure paths after the entry guard (preparation failure, queue-persist failure) — both run after an `await`, so a different recording can have started in the gap
- `retranscribeSavedMeeting`'s four early guards (dictation busy / capture active / background work / speaker review pending) — same shape as `importAudioFile`'s guard
- `applyModelPreparationResult`'s failure branch — `prepareModels()` is awaited from import/retranscribe/start-recording flows that can race a different concurrent recording
- `handleDisplayStatusChange`'s skipped-transcript branch — `taskManager` runs one job at a time, but that job can belong to an earlier meeting queued behind a live one
- `cancelActiveTranscription` — same unconditional-write bug (`transition(to: .ready, ...)` with no guard at all), not explicitly in your list but the identical root cause: Home's "cancel current activity" button is gated on `displayStatus`, which reflects background work visible while a different meeting records live

`displayStatus` still updates unconditionally at the sites that had it — it's a separate, capture-independent signal and was never part of the gate-stomping bug.

## P1c — prepareForTermination doesn't handle .startingRecording

Confirmed: `shouldConfirmQuitForActiveCapture` now covers `.startingRecording`, so "Save Audio and Quit" can be chosen while a start is mid-flight, but `prepareForTermination()` only checked `case .recording = state`.

**Fix:** added `waitForRecordingStartBeforeTermination()` (bounded by the same generous outer deadline `waitForRecordingFinishBeforeTermination()` already uses; the real bound is `capture.startRecording()`'s own internal 12s timeout, which guarantees `state` leaves `.startingRecording` well before that). `prepareForTermination()` joins it first; afterward `state` is `.recording` (falls into the existing save path) or `.error` (falls through to a normal quit) — no duplicated save-path logic needed.

## P2a — missing legal-transition-table edges

Added all three: `transcribing -> startingRecording` (start-while-transcribing), `idle -> transcribing` (a recovered/imported job can start before `prepareModels()` has ever run once), `idle -> ready` (a concurrent flow — e.g. a model-selection reset — can move `state` back to idle while an import cancellation is resolving). Also had to flip one existing test that asserted `transcribing -> startingRecording` was illegal — it was wrong given `startRecording()`'s own entry switch explicitly allows starting from `.transcribing`. Added regression tests for all three new edges plus the `mayReportUnrelatedFailureAsError` rule.

## P2b — .startingRecording presentation

Confirmed my original comment was wrong — I'd conflated the starting and stopping windows. `.stoppingRecording` genuinely preserves old behavior (state stayed `.recording` for the whole teardown window pre-collapse), but `.startingRecording` doesn't: pre-collapse, `state` during mic-engage was whatever it was *before* the start began (`.ready` in the common case), never `.recording`. Fixed both `MeetingOverlayController` (`presentationState`/`applySessionState`) and `HomeTranscriptionActivityPresentation` to group `.startingRecording` with `.loadingModels` ("preparing"/"getting ready") instead of `.recording`.

## Verification

Foreground, in order, all green: `build-deps.sh --force`, `build.sh --no-open` (same two pre-existing unrelated warnings as before, confirmed present on `origin/main`), `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 run-tests.sh` (12730/12730 passed — 10 new tests), `run-integration-smoke.sh`.

r3dbars:
Fixed all 4 residual findings (commit 90437691).

## P1d — "Stop and Transcribe" silently no-ops during .startingRecording

Confirmed: `stopRecording()`'s entry guard is `case .recording = state`, so a bare call during `.startingRecording` does nothing, `applicationShouldTerminate` still returns `.terminateCancel` (quit cancelled), and the pending start goes on to resolve to `.recording` — the meeting keeps recording despite the user explicitly choosing "Stop and Transcribe."

**Fix:** added `MeetingSessionController.stopRecordingJoiningPendingStart(reason:)`, which joins a pending start first (reusing the same bounded wait `prepareForTermination()` already used — renamed `waitForRecordingStartBeforeTermination` → `waitForPendingRecordingStartToResolve` since it's now shared by both callers) and then calls `stopRecording(reason:)`. `applicationShouldTerminate`'s `.stopAndTranscribe` branch now calls this instead of `stopRecording()` directly.

## P1e — import-cancellation forces every non-.transcribing state to .ready

Confirmed — this is the same P1b bug class, reintroduced in a spot that round didn't touch: `if case .transcribing = state {} else { transition(to: .ready, ...) }` only exempted `.transcribing`, so a cancellation landing after a *different* recording started concurrently during the import's async window would force that live capture's state to `.ready`, clearing every `isCaptureSessionActive` gate while the mic was still physically running.

**Fix:** exempt the whole capture-active window, not just `.transcribing`:
```swift
if case .transcribing = state {
    // Leave as-is: a background job is actively transcribing.
} else if !isCaptureSessionActive {
    transition(to: .ready, reason: "import_cancelled")
}
```

### Full sweep (as requested)

Grepped every `transition(to:` (and `reportUnrelatedFailure`) call site in `MeetingSessionController.swift` post-fix and classified each:

| Site (reason) | Target | Classification |
|---|---|---|
| `model_preparation_started` (in `prepareModels()`) | `.loadingModels` | **Found unguarded in this sweep, now fixed.** `prepareModels()` is reused by 3 callers; `importAudioFile`/`retranscribeSavedMeeting` synchronously check `state` immediately before calling (no true race — MainActor doesn't preempt without a real suspension point in between), but `FailedMeetingStore.prepareModelsForRetry` calls it unconditionally after its own async retry steps, with no such check. Guarded with `if !isCaptureSessionActive`. |
| `capture_start_requested`/`capture_start_confirmed`/`capture_start_failed` | `.startingRecording`/`.recording`/`.error` | Capture-lifecycle-owned (`startRecording()`'s own flow, reentrancy-guarded) |
| `stop_requested`/`stop_completed`/`stop_timeout`/`stop_missing_mic_audio` | `.stoppingRecording`/`.transcribing`/`.error` | Capture-lifecycle-owned (`stopRecording()`'s own synchronous flow) |
| `cancel_requested`/`cancel_completed*` | `.stoppingRecording`/`.transcribing`/`.error`/`.ready` | Capture-lifecycle-owned (`cancelRecording()`'s own flow — reentrancy-guarded via its own `case .recording = state` entry check, so no other flow can touch `state` mid-cancel) |
| `unexpected_capture_stop` | `.error` | Capture-lifecycle-owned (`handleUnexpectedCaptureStop`, driven by the real capture-stop callback) |
| `prepare_for_termination`/`prepare_for_termination_preserved_work` | `.stoppingRecording`/`.ready` | Capture-lifecycle-owned (single quit-path invocation, already joins pending start/stop first) |
| `transcription_job_started`/`transcription_work_continues`/`transcription_queue_settled_*` | `.transcribing`/`.ready`/`.error` | Guard-routed — each has its own `guard !isCaptureSessionActive else { return }` |
| `transcription_job_prepare_failed`/`import_preparation_failed`/`import_queue_persist_failed`/`retranscribe_blocked_*`/`model_preparation_failed`/`transcript_skipped` | `.error` | Guard-routed via `reportUnrelatedFailure` |
| `transcription_cancelled` (`cancelActiveTranscription`) | `.ready` | Guard-routed via inline `if !isCaptureSessionActive` |
| `import_cancelled` | `.ready` | **This round's fix** — now guard-routed |
| `model_preparation_succeeded` | `.ready` | Guard-routed — `switch state { case .recording, .transcribing, .startingRecording, .stoppingRecording: break; default: transition(...) }` |
| `speech_model_selection_changed` (`resetPreparedSpeechModelIfNeeded`) | `.idle` | Guard-routed — same function has its own switch excluding all 4 capture-related states above this line; function is fully synchronous (no `async`), so no interleaving is possible within it |
| `retranscribe_started` | `.transcribing` | Pre-existing guard, unrelated to this PR: `guard case .ready = state else { return false }` immediately precedes it, which structurally excludes every non-`.ready` state including all three capture-live ones |

No direct `state =` writes exist outside `MeetingSessionController.swift` (`state` is `private(set)`), and `TranscriptionQueueCoordinator.swift` only reaches `state` through the four guarded outcome-report methods.

## P2c — silent import rejection

Confirmed: `importAudioFileFromSettings()` discarded `importAudioFile(from:)`'s `Bool` return with `_ =`. Added an `NSAlert` at the call site (checked via the already-public `isCaptureSessionActive` before even calling `importAudioFile`), matching the existing `confirmQuitDuringActiveMeeting()` alert pattern in the same file rather than building new UI surface.

## P2d — activation-policy publisher misses starting/stopping

Confirmed: `wireActivationPolicy`'s meeting-recording publisher mapped `isSteadyStateRecording` into `ActivationPolicyController.setMeetingRecording`, which exists specifically so the app "stays visible in Cmd+Option+Esc for recovery" (its own header comment). With Show in Dock off, that left the app invisible in the force-quit dialog for the ~12s mic-engage window and the up-to-~120s stop/cancel teardown window. Switched to `MeetingSessionStateMachine.isCaptureSessionActive`, the same broad signal `shouldConfirmQuitForActiveCapture` already uses.

## Tests

`ActivationPolicyControllerTests.swift` already fully covers `ActivationPolicyDecision`/`ActivationPolicyController`'s pure boolean-in behavior — the P2d bug was entirely in which upstream Combine signal fed that boolean (app-delegate wiring, not independently pure-testable without a live NSApplication/Combine harness), so there's no new pure surface to add a test for there. The import-cancel fix (P1e) reuses the already-tested `isCaptureSessionActive` directly rather than a new named rule, so no new test needed on that front either — flagging both explicitly rather than silently skipping the "add tests" ask.

## Verification

Foreground, in order, all green: `build-deps.sh --force`, `build.sh --no-open` (same two pre-existing unrelated warnings, unchanged), `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 run-tests.sh` (12730/12730 passed), `run-integration-smoke.sh`.

r3dbars:
Fixed all 3 round-3 findings (commit 521483ed).

## 1 — Task body still discarded the import result

Confirmed: the up-front `isCaptureSessionActive` guard closes the common case, but `Task { _ = await importAudioFile(from: url) }` still swallowed the boolean, so a meeting starting in the (small, real) gap between the guard and the Task body actually running would make the in-controller rejection silent again. Fixed:

```swift
Task {
    let started = await appState.meetingSession.importAudioFile(from: url)
    guard !started, appState.meetingSession.isCaptureSessionActive else { return }
    presentImportBlockedByActiveCaptureAlert()
}
```

Extracted the alert into `presentImportBlockedByActiveCaptureAlert()` so both the up-front guard and this Task-body re-check show identical UI. Re-checking `isCaptureSessionActive` after the call (rather than alerting on any `false`) keeps this specific to the capture-active rejection — `importAudioFile` returns `false` for other reasons too (model prep failing, etc.), and those already surface through `state` correctly.

## 2 — canRetry gated on the narrow isRecording

Confirmed: `makeFailedMeetingStore()`'s `canRetry` closure checked `!self.isRecording` (steady-state-only), so a retry could pass the gate and reach `prepareModelsForRetry` → `prepareModels()` during `.startingRecording`/`.stoppingRecording` — exactly the window `prepareModels()`'s own internal guard (added in the round-2 fix) was defending against. Switched to `!self.isCaptureSessionActive`.

## 3 — table gap: `start_blocked_permission`

Added. Full, now-exhaustive table (supersedes the round-2 one — one addition, one entry corrected to note it resumes after an `await`):

| Site (reason) | Target | Classification |
|---|---|---|
| `start_blocked_permission` (~657-672) | `.error` | **Added this round.** Capture-lifecycle-owned: it's `startRecording()`'s own permission-check failing, before `.startingRecording` is ever reached — resumes after `await resolveStartRecordingPermissionDecision(...)` (a real suspension point, e.g. an OS permission-prompt round trip), but `state` at that point is still whatever it was at `startRecording()`'s entry (the top switch already excluded `.recording`/`.startingRecording`/`.stoppingRecording` as valid entry states, and `startRecordingCallInFlight` blocks a second overlapping call from this same session), so no unrelated concurrent capture can be live here. Distinct from `capture_start_failed` (~744), which is the later capture-engage failure once `.startingRecording` is already set. |
| `model_preparation_started` (in `prepareModels()`) | `.loadingModels` | Guard-routed (round 2): `if !isCaptureSessionActive` |
| `capture_start_requested`/`capture_start_confirmed`/`capture_start_failed` | `.startingRecording`/`.recording`/`.error` | Capture-lifecycle-owned (`startRecording()`'s own flow, reentrancy-guarded) |
| `stop_requested`/`stop_completed`/`stop_timeout`/`stop_missing_mic_audio` | `.stoppingRecording`/`.transcribing`/`.error` | Capture-lifecycle-owned (`stopRecording()`'s own synchronous flow) |
| `cancel_requested`/`cancel_completed*` | `.stoppingRecording`/`.transcribing`/`.error`/`.ready` | Capture-lifecycle-owned (`cancelRecording()`'s own flow, reentrancy-guarded via its own `case .recording = state` entry check) |
| `unexpected_capture_stop` | `.error` | Capture-lifecycle-owned (`handleUnexpectedCaptureStop`, driven by the real capture-stop callback) |
| `prepare_for_termination`/`prepare_for_termination_preserved_work` | `.stoppingRecording`/`.ready` | Capture-lifecycle-owned (single quit-path invocation, joins pending start/stop first) |
| `transcription_job_started`/`transcription_work_continues`/`transcription_queue_settled_*` | `.transcribing`/`.ready`/`.error` | Guard-routed — each has its own `guard !isCaptureSessionActive else { return }` |
| `transcription_job_prepare_failed`/`import_preparation_failed`/`import_queue_persist_failed`/`retranscribe_blocked_*`/`model_preparation_failed`/`transcript_skipped` | `.error` | Guard-routed via `reportUnrelatedFailure` |
| `transcription_cancelled` (`cancelActiveTranscription`) | `.ready` | Guard-routed via inline `if !isCaptureSessionActive` |
| `import_cancelled` | `.ready` | Guard-routed (round 2 fix): exempts the whole capture-active window, not just `.transcribing` |
| `model_preparation_succeeded` | `.ready` | Guard-routed — `switch state { case .recording, .transcribing, .startingRecording, .stoppingRecording: break; default: transition(...) }` |
| `speech_model_selection_changed` (`resetPreparedSpeechModelIfNeeded`) | `.idle` | Guard-routed — same function's switch excludes all 4 capture-related states above this line; function is fully synchronous (no `async`), so no interleaving is possible within it |
| `retranscribe_started` | `.transcribing` | Pre-existing guard, unrelated to this PR: `guard case .ready = state else { return false }` immediately precedes it |

Still true: no direct `state =` writes exist outside `MeetingSessionController.swift` (`state` is `private(set)`), and `TranscriptionQueueCoordinator.swift` only reaches `state` through the four guarded outcome-report methods.

## Verification

Foreground: `build.sh --no-open` (same two pre-existing unrelated warnings, unchanged), `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 run-tests.sh` (12730/12730 passed), `run-integration-smoke.sh` — all green.

PR #1636, opened 2026-08-04
Collapse TranscriptionTaskManager's lockstep task-state collections into one enum

r3dbars:
## Summary

`TranscriptionTaskManager` tracked one task's identity across five parallel collections keyed by the same `UUID`:

- `activeTasks: [UUID: Task<Void, Never>]`
- `activeTaskAudio: [UUID: ActiveTaskAudio]`
- `preservedTaskIdsForShutdown: Set<UUID>`
- `intentionallyCancelledTaskIds: Set<UUID>`
- `committedTranscriptTaskIds: Set<UUID>`

Functions like `finishCancelledTaskIfNeeded`, `cancelAll()`, and `preserveActiveTranscriptionsForShutdown()` had to consult/mutate several of these in a specific, commented-as-fragile order.

This PR replaces **four of the five** with one `private var tasks: [UUID: TaskLifecycleState]`, where each task is in exactly one state by construction. `activeTasks` (the raw `Task<Void, Never>` handle map) stays a separate stored property — see "Why `activeTasks` stays separate" below.

## Combination analysis

Read every call site that touched `activeTaskAudio`, `preservedTaskIdsForShutdown`, `intentionallyCancelledTaskIds`, or `committedTranscriptTaskIds` (all four are `private`, so this is a closed set: `startTranscription`, `startImportedTranscription`, `startSavedAudioRetranscription`, `retryFailedTranscription`/`performRetry`, `markTaskTranscriptCommitted`, `canCommitTaskSideEffects`, `finishCancelledTaskIfNeeded`, `cancelAll`, `preserveActiveTranscriptionsForShutdown`, `handleTaskCompletion`, and the three imported-scratch-cleanup helpers).

Of the 2^4 = 16 theoretical combinations of {hasAudio, preserved, cancelled, committed}, **5 are reachable**, each mapped to one enum case:

| Reachable combination (audio / preserved / cancelled / committed) | Enum case | How it's reached |
|---|---|---|
| audio-or-nil / — / — / — | `.active(audio:)` | Normal start. `audio` is non-nil for live/imported captures, `nil` for saved-audio-retranscription and failed-row retries (which reuse already-retained files — the old code never entered those into `activeTaskAudio` either). |
| audio-or-nil / — / — / committed | `.committed(audio:)` | `markTaskTranscriptCommitted` runs mid-flight (side effects saved) while the task is still finishing speaker-naming/scratch-cleanup work. |
| — / — / cancelled / — | `.cancelling` | `cancelAll()` marks every currently-occupied task cancelled and unconditionally discards its audio. |
| — / — / cancelled / committed | `.cancellingCommitted` | **The combination the audit flagged as "old code allowed cancelled AND committed simultaneously."** Verified reachable, not merely theoretical: `cancelAll()` iterates `activeTasks` and marks every key cancelled *without checking commit state first* — a task that already committed can still receive the cancel marker if `cancelAll()` runs before its async body returns. Old code's `finishCancelledTaskIfNeeded` checked `committedTranscriptTaskIds` **before** `intentionallyCancelledTaskIds`, so committed always won and the stray cancel marker was silently dropped. This PR models that explicitly instead of leaving it implicit: `.cancellingCommitted` collapses back to `.committed` the next time `finishCancelledTaskIfNeeded` runs for that id. |
| — / preserved / — / — | `.preservedForShutdown` | `preserveActiveTranscriptionsForShutdown()` already synchronously persisted the task's audio into the failed queue and evicted it from `activeTasks`/occupancy counters. Only this marker remains so the still-running (uncooperative CoreML) task body can recognize its own outcome should be suppressed when it eventually returns. |

Combinations that are **not** reachable (and why): `preserved` never coexists with `cancelled` (`preserveActiveTranscriptionsForShutdown` and `cancelAll` are two different shutdown/cancel paths — `cancelAll` unconditionally clears any leftover `.preservedForShutdown` markers at the end, mirroring the old `preservedTaskIdsForShutdown.removeAll()`); `preserved` never coexists with `committed` as an *observed* state because both `finishCancelledTaskIfNeeded` (success path) and the failure-path's preserved-check run generically enough that a `.preservedForShutdown` marker is always consumed/cleared before anything would need to inspect a co-occurring commit flag — see `consumePreservedForShutdownMarker`.

## Why `activeTasks` stays separate

The design goal was one map for everything, including the `Task` handle. In practice, `activeTasks` (`internal`, not `private`) is directly poked by test code in five files as an occupancy-guard seam — e.g. `manager.activeTasks[failedId] = sentinel`, `manager.activeTasks[activeTaskId]?.cancel()`, `manager.activeTasks.removeValue(forKey:)` — across `TranscriptionTaskManagerStopTimeoutTests.swift`, `TranscriptionTaskManagerImportedAudioTests.swift`, `PipelineRollbackRegistryManagerTests.swift`, `TranscriptionTaskManagerFailedQueueTests.swift`, and `TranscriptionTaskManagerRetryTests.swift` (all extensions of the 64-test `TranscriptionTaskManagerMetadataTests` class named in this item's brief). Folding it into `TaskLifecycleState` would mean either rewriting those direct pokes (for no behavior change — `activeTasks` genuinely is a different *kind* of thing, a concurrency handle, not state data) or building a computed-property bridge with diffing semantics, which is more moving parts for the same outcome. `activeTaskAudio`, `preservedTaskIdsForShutdown`, `intentionallyCancelledTaskIds`, and `committedTranscriptTaskIds` are all `private` — never touched directly by any test — so collapsing those four was unambiguous.

## Per-collection migration table

| Old | New | Notes |
|---|---|---|
| `activeTaskAudio[id]` | `tasks[id]?.audio` | `.active`/`.committed` carry audio; `.cancelling`/`.cancellingCommitted`/`.preservedForShutdown` never do (matches old code, which unconditionally stripped `activeTaskAudio` in both `cancelAll()` and after computing the shutdown snapshot). |
| `preservedTaskIdsForShutdown.contains`/`.remove` | `case .preservedForShutdown = tasks[id]` / `consumePreservedForShutdownMarker(taskId:)` | New helper mirrors the old `Set.remove(_:) != nil` check-and-remove-in-one-step idiom. |
| `intentionallyCancelledTaskIds.contains` | `tasks[id]?.isCancelling ?? false` | True for both `.cancelling` and `.cancellingCommitted`. |
| `committedTranscriptTaskIds.contains` | `tasks[id]?.isCommitted ?? false` | True for both `.committed` and `.cancellingCommitted`. |
| `handleTaskCompletion` clearing all 4 sets | `tasks.removeValue(forKey:)` | One removal instead of four. |

## Incidental fix (documented, not silent)

`performRetry`'s success and failure branches did their own manual partial cleanup (`activeTasks.removeValue` + counters) instead of calling `handleTaskCompletion`. If a retry task had been marked `.committed` (or raced a `cancelAll()` into `.cancellingCommitted`) before reaching that point, the old code left a permanently stale entry behind in `committedTranscriptTaskIds`/`intentionallyCancelledTaskIds` — harmless in practice (both were `private`, never observed again, and UUIDs don't recur) but a real leak. This PR's equivalent manual-cleanup branches now also clear `tasks[failedId]`, closing it. No observable behavior changes — flagged inline in the diff with a comment at both call sites.

## Test changes

No existing test was modified. `TranscriptionTaskManagerMetadataTests` (82 tests across its extension files) passes unmodified, including the ones that directly poke `activeTasks`.

Added `Tests/TranscriptedCoreTests/PipelineTests/TranscriptionTaskLifecycleStateTests.swift` (4 new tests, another extension on the same class, matching the established pattern in `TranscriptionTaskManagerImportedAudioTests.swift` / `PipelineRollbackRegistryManagerTests.swift`):

- `testCancelDuringRunningSuppressesLateResultFromAnUncooperativeEngine` — the literal "CoreML calls are not guaranteed to observe cancellation immediately" scenario from the old `cancelAll()` comment: a stub STT engine that never checks `Task.isCancelled` still gets suppressed correctly because `.cancelling` is a state flag, not a race with the engine.
- `testCancelAllRacingAnAlreadyCommittedTaskDropsQuitConfirmationAndRejectsALaterCommit` — drives `commitSavedTranscriptSideEffectsUnlessCancelled` for real (same pattern as the existing "superseded task" test), then `cancelAll()`, proving `.cancellingCommitted` is reached and that `canCommitTaskSideEffects` correctly rejects a second commit attempt afterward.
- `testShutdownPreservationIgnoresAnAudioLessRetranscriptionTask` — proves `preserveActiveTranscriptionsForShutdown()` is a true no-op (0 preserved, task left running) for a saved-audio-retranscription task, since it never owned preservable scratch audio.
- `testShutdownPreservationEvictsOccupancySynchronouslyBeforeTheEngineNoticesCancellation` — proves eviction from `activeTasks`/`tasks` happens synchronously inside `preserveActiveTranscriptionsForShutdown()`, before the still-blocked underlying engine call returns.

## Part B (app-side `MeetingSessionController` job mirror)

**Deferred**, per this item's own fallback instruction ("if B conflicts badly or would balloon scope, ship A+C alone and document B as follow-up"). Given the size of the Core-side combination analysis and the amount of verification budget already spent proving it byte-for-byte behavior-preserving, I did not start B in this PR. `MeetingSessionController.swift` (~3.1k lines, already flagged as a hotspot) and `TranscriptionQueueCoordinator.swift` still mirror queue-job identity in scalar fields correlated to the coordinator's queue state by UUID equality — that's real follow-up work, ideally its own PR so it can get equally careful before/after behavior verification against the post-#1634 `transition(to:reason:)` design.

## Verification (all foreground, all green)

1. `bash build-deps.sh --force` — clean
2. `bash build.sh --no-open` — clean (2 pre-existing warnings in `MeetingSessionController.swift`/`DictationSessionController.swift`, unrelated to this change, not touched by this diff)
3. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12730 tests, 12730 passed, 0 failed
4. `bash run-integration-smoke.sh` — core smoke + wake-recovery smoke both OK
5. `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 swift test` — 950 tests, 0 failures, 13 skipped (pre-existing, environment-gated, unrelated to this change)

No pre-existing failures to diff against — everything is green against clean `origin/main` (verified via `git log --oneline -5` showing PR #1634's merge commit before branching).

### DIFF
--- a/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
+++ b/Sources/TranscriptedCore/Pipeline/TranscriptionTaskManager.swift
@@ -69,7 +69,20 @@ public class TranscriptionTaskManager: ObservableObject {
         /// returns. Deliberately has no associated `Task<Void, Never>` — by
         /// the time this case exists, `activeTasks` no longer has an entry for
         /// this id.
-        case preservedForShutdown
+        ///
+        /// `wasCommitted` carries forward whatever `isCommitted` was true at the
+        /// moment of preservation. This is required, not decorative: on the old
+        /// five-collection code, `committedTranscriptTaskIds` and
+        /// `preservedTaskIdsForShutdown` were independent `Set`s, so a task that
+        /// committed *before* shutdown kept its committed membership even after
+        /// `cancelAll()` later wiped `preservedTaskIdsForShutdown` with its own
+        /// unconditional `.removeAll()` — `finishCancelledTaskIfNeeded` would
+        /// still see `committedTranscriptTaskIds.contains(taskId) == true` and
+        /// give the committed outcome precedence over the task's own
+        /// `CancellationError`. Collapsing to a payload-less `.preservedForShutdown`
+        /// would silently lose that precedence the moment `cancelAll()` clears the
+        /// marker. See `cancelAll()`'s handling of this case.
+        case preservedForShutdown(wasCommitted: Bool)
 
         var audio: ActiveTaskAudio? {
             switch self {
@@ -84,7 +97,9 @@ public class TranscriptionTaskManager: ObservableObject {
             switch self {
             case .committed, .cancellingCommitted:
                 return true
-            case .active, .cancelling, .preservedForShutdown:
+            case .preservedForShutdown(let wasCommitted):
+                return wasCommitted
+            case .active, .cancelling:
                 return false
             }
         }
@@ -1744,13 +1759,21 @@ public class TranscriptionTaskManager: ObservableObject {
                     failedTranscriptionManager.deleteFailedTranscription(id: failedId)
                 }
                 self.activeTasks.removeValue(forKey: failedId)
-                // NOTE: also clears `tasks[failedId]` here (unlike the pre-refactor code,
-                // which left stale entries behind in `committedTranscriptTaskIds` /
-                // `intentionallyCancelledTaskIds` for a retry that reached this success
-                // path after being marked committed — a latent, purely-internal leak: those
-                // collections were `private` and never observed again for a UUID that
-                // will not recur). Removing it here is a strict improvement with no
-                // observable behavior change.
+                // NOTE: also clears `tasks[failedId]` here — a real behavior fix, not just
+                // internal bookkeeping. When `waitingForSpeakerNames` is true above, the failed
+                // row for `failedId` is deliberately kept, so the *same* `failedId` can be
+                // retried again later. On the pre-refactor code, this success path never cleared
+                // `committedTranscriptTaskIds` for `failedId` — if THIS retry had already been
+                // marked committed before reaching here, that stale membership would survive
+                // into a later retry of the same id. If that later retry was then cancelled via
+                // `cancelAll()` before it ever committed, `finishCancelledTaskIfNeeded` would
+                // still see the stale committed marker and give it precedence over the later
+                // retry's own `CancellationError` — incorrectly publishing "Retry failed" for a
+                // retry that was actually just cleanly cancelled. Clearing `tasks[failedId]` here
+                // closes that: each retry of the same id now starts from a clean
+                // `.active(audio: nil)` state (set in `retryFailedTranscription`), so a stale
+                // commit from an earlier retry of the same id can never leak into a later one.
+                // See `testSecondRetryOfTheSameFailedIdIsCleanlySuppressedWhenCancelledBeforeCommit`.
                 self.tasks.removeValue(forKey: failedId)
                 self.activeCount = max(0, self.activeCount - 1)
                 self.backgroundTaskCount = max(0, self.backgroundTaskCount - 1)
@@ -1769,7 +1792,8 @@ public class TranscriptionTaskManager: ObservableObject {
 
                 self.activeTasks.removeValue(forKey: failedId)
                 // See the matching NOTE in the success branch above: clears `tasks[failedId]`
-                // too, closing the same latent (purely-internal) leak on the failure path.
+                // here too, so a failed retry of `failedId` also can't leave behind a stale
+                // commit marker for a later retry of the same id to trip over.
                 self.tasks.removeValue(forKey: failedId)
                 self.activeCount = max(0, self.activeCount - 1)
                 self.backgroundTaskCount = max(0, self.backgroundTaskCount - 1)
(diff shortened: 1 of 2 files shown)


r3dbars:
Program paused — P1 fix in progress, do not merge; remaining: confirm the two new regression tests actually pass (and don't hang) under `swift test`, then re-run the full verification suite (`build-deps.sh --force`, `build.sh --no-open`, `run-tests.sh`, `run-integration-smoke.sh`, `swift test`).

Status of the two Codex review findings as of this push (commit 24cb0ce2):

**P1** (missed reachable state: preserved+committed): Fixed. `.preservedForShutdown` now carries `wasCommitted: Bool` instead of being payload-less, `preserveActiveTranscriptionsForShutdown()` captures that bit at transition time, and `cancelAll()`'s cleanup pass downgrades a committed marker to plain `.committed(audio: nil)` instead of dropping it outright — matching the old code's independent-`Set`s behavior where `committedTranscriptTaskIds` survived `cancelAll()`'s unconditional `preservedTaskIdsForShutdown.removeAll()`. Source change compiles clean (`swift build`).

**P2** (comment was wrong; this is a real fix): Fixed. Updated both comments at the `performRetry` success/failure cleanup sites to correctly describe why clearing `tasks[failedId]` there is a genuine behavior fix (a kept-alive failed row can be retried again under the same id; without this fix a stale commit marker from an earlier retry could incorrectly win over a later retry's real cancellation).

**Not yet verified**: added `testShutdownPreservationThenCancelAllKeepsCommittedPrecedenceOverTheTasksOwnCancellation` and `testSecondRetryOfTheSameFailedIdIsCleanlySuppressedWhenCancelledBeforeCommit` to pin both fixes. A `swift test` run for the full `TranscriptionTaskLifecycleStateTests` + `TranscriptionTaskManagerMetadataTests` filter was still running past 300s (everything through the alphabetically-earlier `testSavedAudio*` group had passed normally when it was interrupted; the two new tests run later alphabetically and hadn't been reached yet, given prior full-suite runs of this same filter took ~3s). This needs a follow-up pass to isolate whether one of the two new tests hangs and fix it before this branch is done — do not merge until that's confirmed green.

r3dbars:
Resumed and completed after the pause: committed-precedence fix (preservedForShutdown carries wasCommitted; cancelAll restores it), corrected performRetry comments, and the hanging regression test fixed (stub now accounts for per-channel transcribeSegment calls — the hang was the first retry's mic-channel call blocking). Full verification green: build, 12735/12735 fast tests, SPM suite, integration smoke. Rebased onto current main.

PR #1639, opened 2026-08-05
Bring system-audio (SCK) recovery to parity with the mic path

r3dbars:
## Summary

System-audio capture (SCK, macOS 26+ ScreenCaptureKit) was a second-class citizen in recovery/health compared to the mic path. This PR closes the gap through one reporting seam, without changing either path's tuning values:

1. **Health visibility** — `SCKAudioCapture` now reports device-switch/gap events through a new `SystemAudioCaptureEngine.recoveryEventPublisher`. `Audio` subscribes to it (in `wireSystemAudioStatusPublisher`) and feeds the events into the SAME `deviceSwitchCount` / `recordingGaps` the mic path already uses, so `RecordingHealthInfo.captureQuality` now reflects system-audio dropouts instead of being blind to them.
2. **Wake hook** — `Audio.installWorkspaceSleepWakeObservers()`'s existing wake handler now also calls `systemAudioCapture?.recoverAfterSystemWake()` (a new protocol method, default no-op) right alongside the existing mic-path proactive kick. No new `NSWorkspace` observer was added — this routes through the one that already exists.
3. **Constants** — mic and system-audio recovery tuning constants now live in one named surface, `AudioRecoveryTuning` (`Sources/TranscriptedCore/Audio/AudioDeviceRecovery.swift`), with `Mic` and `SystemAudio` sub-namespaces. **Values are unchanged** — this is mechanism unification, not a tuning change.

### Design decisions

- **Protocol-based seam, not a downcast.** Added `SystemAudioRecoveryEvent` (`.deviceSwitch` / `.gap(duration:)`) and two new `SystemAudioCaptureEngine` requirements (`recoveryEventPublisher`, `recoverAfterSystemWake()`), both with default (empty/no-op) implementations via a protocol extension. This means the legacy CoreAudio `SystemAudioCapture` and existing test stubs need zero changes — only `SCKAudioCapture` overrides them with real behavior. Mirrors the existing `errorMessagePublisher` pattern exactly.
- **Reused the existing counters, not new fields.** `RecordingHealthInfo` was NOT changed. System-audio gaps/switches feed straight into `Audio.deviceSwitchCount` / `Audio.recordingGaps` — the exact same storage `RecordingHealthInfo.from(...)` already reads for the mic path. This keeps the blast radius to `TranscriptedCore/Audio/*` — no `TranscriptFormatter`, `TranscriptSaver`, frontmatter, `docs/capture-format.md`, or `Tools/TranscriptedCaptureKit` parser changes needed. System-audio gaps are labeled with `reason: "System audio reconnect"` in the gap description so they're distinguishable in saved transcripts.
- **Wake hook reuses `handleMidRecordingFailure`, not a second recovery path.** `SCKAudioCapture.recoverAfterSystemWake()` just calls the existing bounded-recovery method with a synthetic message, guarded by "only if currently capturing." Same one-attempt cap, same epoch/token ownership — no new state machine.
- **`SystemAudio.maxRecoveryAttempts` stays at 1.** Explicitly _not_ raised to match the mic path's 5. SCK's bounded recovery restarts an entire ScreenCaptureKit stream (`stopSync` → `prepare` → `start`, including a live `SCShareableContent` fetch) — much heavier than reinstalling a mic tap. The wake hook spends that one attempt proactively on every wake during an active recording (mirroring the mic path's existing unconditional post-wake kick) — see the budget-reset rule below for why this can't starve a later genuine failure. Raising the attempt cap, and/or gating the wake kick on some cheaper liveness signal first, are both deliberate follow-ups — flagging here rather than changing behavior in a parity PR.

### Review round: 3 fixes

Codex review flagged three P2s on the initial version of this PR. All three are fixed on this branch:

1. **Wake kick could exhaust the single recovery attempt.** `recoverAfterSystemWake()` reuses `handleMidRecordingFailure`'s bounded recovery, and that recovery never reset its attempt counter on success — only a brand-new (non-recovery) `start()` did. Since the wake kick runs unconditionally on every wake during a recording (mirroring the mic path) and almost always succeeds (nothing was actually broken), it would have permanently spent SCK's one-attempt budget for the rest of the recording, so a LATER genuine dropout would find zero attempts left and fail immediately where pre-PR it would have recovered.
   **Rule implemented:** `SCKAudioCapture.recoveryAttempts` resets to `0` on every confirmed successful recovery (reactive failure or proactive wake kick alike) — exactly mirroring the mic path's `Audio.finalizeMicRecoveryArtifacts` resetting `recoveryAttemptCount = 0` on success — so a successful wake kick can never reduce the recovery capacity available to a subsequent genuine failure in the same recording. (If the ONE attempt is instead spent on a wake kick that *fails*, system audio for that recording is over regardless — that's the existing, unchanged 1-attempt-cap behavior, not something this PR changes.)
   New test: `SCKAudioCaptureInterleavingTests.testRecoveryAttemptsResetToZeroAfterSuccessfulRecovery` exercises the real `resetRecoveryAttemptsAfterSuccess()` production method (via a thin `...ForTesting` seam, since driving a full success needs live ScreenCaptureKit) to prove a spent attempt returns to `0`.

2. **Recovery-gap duration was measured from the wrong start point.** `failureDetectedAt` was captured at the top of `handleMidRecordingFailure` — i.e. AFTER the ≥5s buffer-stall watchdog had already been silently waiting, so the published gap omitted the largest part of the real outage. **Fixed:** the gap is now measured from `SCKAudioCapture`'s tracked last-real-buffer timestamp (`lastKnownBufferTime()`, the same value the buffer-stall watchdog itself compares against), captured once before any recovery step can reset it, exactly mirroring how the mic path measures its gap from `lastBufferTime` captured before `recoverFromDeviceChange` begins.
   **Residual, intentionally-not-fixed divergence from mic semantics:** unlike the mic path's `waitForMicBuffer`, SCK's recovery declares success as soon as ScreenCaptureKit's `start()` callback returns without error — it does not wait for a real audio frame to actually flow before publishing the `.gap` event. A stream that "successfully" restarts but silently never resumes delivering buffers would therefore still publish a (too-short) gap here; the next real stall would only be caught by the buffer-stall watchdog on its next cycle. Adding a post-start buffer-wait to SCK (mirroring `waitForMicBuffer`) is a larger, riskier change to an already-intricate lock/epoch state machine and is left as a deliberate follow-up rather than bolted on here — see the code comment at the `.gap` publish site.
   New tests: `testLastKnownBufferTimeReflectsLastRealBufferNotWhenAskedForIt` and `testLastKnownBufferTimeAdvancesOnlyWhenARealBufferArrives` exercise the real `lastKnownBufferTime()` / `markBufferReceived()` / `resetBufferWatchdogState()` production methods directly.

3. **`deviceSwitchCount` lost-update race.** The mic path increments this counter from its background recovery queue; the new SCK-path subscription increments it from `DispatchQueue.main`. The old property did a `get` (lock, read, unlock) then a `set` (lock, write, unlock) as two SEPARATE lock acquisitions, so `deviceSwitchCount += 1` from two genuinely concurrent callers could interleave and drop an increment. **Fixed:** added `Audio.incrementDeviceSwitchCount()`, which does the read-modify-write inside ONE lock acquisition, and routed both increment sites (`Audio.recordSystemAudioDeviceSwitch()` and the mic path's `recoverFromDeviceChange` in `AudioDeviceRecovery.swift`) through it. Checked `recordingGaps`' append path too — `appendRecordingGap` already mutates inside a single lock acquisition, so it wasn't exposed to this race and needed no change.
   New tests: `testIncrementDeviceSwitchCountSurvivesHighConcurrencyWithoutLostUpdates` (8 queues × 500 increments each) and `testConcurrentMicAndSystemAudioDeviceSwitchesBothLand` (mirrors the real shape — one background queue + main queue, concurrently). Verified both tests actually catch the bug: temporarily reverting `incrementDeviceSwitchCount()` to the old `+=` pattern made the first test fail with `1601 != 4000` (real lost updates observed), confirming the tests aren't vacuous.

### Constants table (unchanged values, now named in one place)

| | Mic (`AudioRecoveryTuning.Mic`) | System audio (`AudioRecoveryTuning.SystemAudio`) |
|---|---|---|
| Max recovery attempts | 5 | 1 |
| Stall detection threshold | 3.0s (watchdog `timeSinceLastBuffer`) | 5s (buffer watchdog) |
| Recovery cooldown | 5.0s (min seconds between attempts) | n/a (single-attempt bound, no repeat cooldown needed) |

### Files changed

- `Sources/TranscriptedCore/Audio/SystemAudioCaptureEngine.swift` — new `SystemAudioRecoveryEvent` enum, two new protocol requirements with default (no-op/empty) implementations
- `Sources/TranscriptedCore/Audio/SCKAudioCapture.swift` — publishes `.deviceSwitch` at recovery start and `.gap(duration:)` on recovery success (now measured from `lastKnownBufferTime()`); new `recoverAfterSystemWake()`; new `resetRecoveryAttemptsAfterSuccess()` (called on every successful recovery); constants now sourced from `AudioRecoveryTuning`; new `...ForTesting` seams for `resetRecoveryAttemptsAfterSuccess()` / `resetBufferWatchdogState()` / `markBufferReceived()` / `lastKnownBufferTime()`
- `Sources/TranscriptedCore/Audio/AudioDeviceRecovery.swift` — new `AudioRecoveryTuning` enum; mic watchdog's stall-threshold literal now reads from it; mic-path device-switch increment routed through `incrementDeviceSwitchCount()`
- `Sources/TranscriptedCore/Audio/Audio.swift` — new `recordSystemAudioDeviceSwitch()` / `recordSystemAudioGap(duration:)`; new atomic `incrementDeviceSwitchCount()`; `wireSystemAudioStatusPublisher` also subscribes to `recoveryEventPublisher`; wake handler calls `systemAudioCapture?.recoverAfterSystemWake()`; `maxRecoveryAttempts` / `recoveryCooldown` now source from `AudioRecoveryTuning.Mic`
- `Tests/TranscriptedCoreTests/AudioTests/SystemAudioRecoveryParityTests.swift` (new) — counter parity, wiring through an injected backend, wake-hook reachability (recording vs. not recording), plus 2 concurrency tests for the `deviceSwitchCount` race fix
- `Tests/TranscriptedCoreTests/AudioTests/SCKAudioCaptureInterleavingTests.swift` — `.deviceSwitch` publishes before the background recovery closure runs (proven not to touch live ScreenCaptureKit APIs by cancelling recovery synchronously from inside the event handler), `recoverAfterSystemWake()` no-ops when idle and triggers bounded recovery when capturing, budget-reset-on-success, and last-known-buffer-time behavior

## Hardware checklist (REQUIRED before merge — this PR does not merge on CI/unit tests alone)

Everything above is verified at the unit level only. `SCKAudioCapture`'s real recovery path (`prepare()` → live `SCShareableContent` fetch → `SCStream` construction) cannot be exercised in CI/SPM tests without real ScreenCaptureKit/TCC state — this matches the existing test suite's convention (no prior test exercises that path either). The following must be proven on real hardware (macOS 26+, Apple Silicon) before this merges:

- [ ] Start a meeting recording with system-audio capture enabled and granted (System Audio Recording permission).
- [ ] **Wake hook — recording through sleep:** Start a recording, put the Mac to sleep (or lid-close if a laptop), wait >10s, wake it. Confirm in `logs/debug.log` / Console that `SCKAudioCapture: proactive recovery after system wake` fires within ~1.5s of wake (alongside the existing mic-path recovery log line), and that system-audio capture resumes (system audio level meter moves again, no `"System audio unavailable"` banner).
- [ ] **Wake hook — not recording:** Put the Mac to sleep/wake with no active recording. Confirm no SCK recovery log lines fire and nothing crashes.
- [ ] **Health visibility — real dropout:** Force a system-audio interruption during a live recording (e.g., toggle System Audio Recording permission off and back on, or another repro that has triggered SCK's watchdog/`didStopWithError` before). After the meeting ends, confirm the saved transcript's YAML frontmatter (`capture_quality`, `audio_gaps`, `device_switches`) reflects the interruption — before this PR, an SCK-only dropout with no mic-side issue would have shown `capture_quality: excellent` / `device_switches: 0`.
- [ ] **No regression — mic path:** Trigger the existing mic-side device-change recovery (unplug/replug a USB or Bluetooth mic mid-recording) and confirm behavior/timing is unchanged (5 attempts, 3s stall threshold, 5s cooldown) — the constants got a new named home but the values and logic are untouched.
- [ ] **No regression — SCK reactive recovery:** With the Mac awake the whole time, trigger SCK's existing buffer-stall/`didStopWithError` recovery independent of wake (whatever repro was used to validate that path before) and confirm it still recovers the same way (still capped at 1 in-flight attempt).
- [ ] **Budget-reset rule:** In one recording, trigger TWO separate system-audio dropouts that each successfully recover (e.g., wake-then-recover, then later a real buffer-stall repro that also recovers). Confirm the second recovery is NOT skipped/terminal — proving the attempt budget really resets to 0 after each success instead of being permanently spent by the first (wake) recovery.
- [ ] Confirm no duplicate/extra `NSWorkspace` sleep or wake observers were registered (only the one `Audio.installWorkspaceSleepWakeObservers()` sets up, per the audit's existing observer-count concern).

## Test plan

- [x] `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh`
- [x] `swift test`
- [ ] Hardware checklist above (blocks merge)

### DIFF
--- a/Sources/TranscriptedCore/Audio/Audio.swift
+++ b/Sources/TranscriptedCore/Audio/Audio.swift
@@ -356,6 +356,31 @@ public class Audio: ObservableObject, @unchecked Sendable {
         _recordingGaps.append(gap)
     }
 
+    /// Records a system-audio-side bounded-recovery attempt (SCK restarting
+    /// its ScreenCaptureKit stream after a buffer stall or stream stop) into
+    /// the SAME counter mic-path device switches use, so system-audio
+    /// dropouts count toward `RecordingHealthInfo.captureQuality` instead of
+    /// being invisible to it. Wired from `wireSystemAudioStatusPublisher`'s
+    /// `recoveryEventPublisher` subscription, which already runs on main.
+    func recordSystemAudioDeviceSwitch() {
+        guard isRecording else { return }
+        deviceSwitchCount += 1
+    }
+
+    /// Records a system-audio recovery gap (bounded SCK recovery succeeded
+    /// after `duration` seconds of stalled/stopped capture), mirroring the
+    /// mic path's `AudioGap` entries into the SAME `recordingGaps` array so
+    /// system-audio interruptions show up in saved transcript health
+    /// metadata the same way mic-side gaps already do.
+    func recordSystemAudioGap(duration: TimeInterval) {
+        guard isRecording else { return }
+        appendRecordingGap(AudioGap(
+            start: Date(timeIntervalSinceNow: -duration),
+            duration: duration,
+            reason: "System audio reconnect"
+        ))
+    }
+
     /// Commit recovery artifacts only while the recovery still owns the
     /// current recording generation. The generation lock also blocks a
     /// concurrent stop/start from advancing the session between the check and
@@ -639,8 +664,8 @@ public class Audio: ObservableObject, @unchecked Sendable {
             _recordingSessionGeneration = newValue
         }
     }
-    let maxRecoveryAttempts = 5
-    let recoveryCooldown: TimeInterval = 5.0  // Min seconds between recovery attempts
+    let maxRecoveryAttempts = AudioRecoveryTuning.Mic.maxRecoveryAttempts
+    let recoveryCooldown: TimeInterval = AudioRecoveryTuning.Mic.recoveryCooldownSeconds  // Min seconds between recovery attempts
 
     func withAudioGraphLock<T>(_ body: () throws -> T) rethrows -> T {
         audioGraphLock.lock()
@@ -931,6 +956,10 @@ public class Audio: ObservableObject, @unchecked Sendable {
 
     // System audio status observation
     private var systemAudioCancellable: AnyCancellable?
+    // Recovery/health event observation (device-switch + gap parity with the
+    // mic path). Separate from `systemAudioCancellable` so re-wiring one
+    // subscription on a new capture attempt doesn't need to touch the other.
+    private var systemAudioRecoveryEventCancellable: AnyCancellable?
     // Protected by systemSilenceLock — written from callback thread, reset on main thread
     private var _systemAudioSilenceStart: Date?
     private let systemSilenceLock = NSLock()
@@ -1105,6 +1134,17 @@ public class Audio: ObservableObject, @unchecked Sendable {
             .sink { [weak self] errorMessage in
                 self?.updateSystemAudioStatus(fromError: errorMessage)
             }
+        systemAudioRecoveryEventCancellable = capture.recoveryEventPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] event in
+                guard let self else { return }
+                switch event {
+                case .deviceSwitch:
+                    self.recordSystemAudioDeviceSwitch()
+                case .gap(let duration):
+                    self.recordSystemAudioGap(duration: duration)
+                }
+            }
     }
 
     func installWorkspaceSleepWakeObservers() {
@@ -1150,6 +1190,14 @@ public class Audio: ObservableObject, @unchecked Sendable {
                 DispatchQueue.global(qos: .userInitiated).asyncAfter(deadline: .now() + 1.0) { [weak self] in
                     guard let self = self, self.isRecording else { return }
                     self.recoverFromDeviceChange(sessionGeneration: sessionGeneration)
+
+                    // Give system audio the same proactive post-wake recovery
+                    // opportunity as the mic path, instead of only relying on
+                    // SCK's own buffer-stall watchdog (which can take up to
+                    // `AudioRecoveryTuning.SystemAudio.stallTimeoutSeconds`
+                    // to notice). No-ops for backends without bounded
+                    // recovery (see `SystemAudioCaptureEngine`'s default).
+                    self.systemAudioCapture?.recoverAfterSystemWake()
                 }
             }
         }
@@ -2069,6 +2117,7 @@ public class Audio: ObservableObject, @unchecked Sendable {
         timer?.invalidate()
         watchdogTimer?.invalidate()
         systemAudioCancellable?.cancel()
+        systemAudioRecoveryEventCancellable?.cancel()
         replaceSystemAudioMonitoringAttempt(with: nil)?.cancel()
         systemAudioCapture?.stopSync()
 
--- a/Sources/TranscriptedCore/Audio/Audio.swift
+++ b/Sources/TranscriptedCore/Audio/Audio.swift
@@ -364,7 +364,7 @@ public class Audio: ObservableObject, @unchecked Sendable {
     /// `recoveryEventPublisher` subscription, which already runs on main.
     func recordSystemAudioDeviceSwitch() {
         guard isRecording else { return }
-        deviceSwitchCount += 1
+        incrementDeviceSwitchCount()
     }
 
     /// Records a system-audio recovery gap (bounded SCK recovery succeeded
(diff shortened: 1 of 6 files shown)


r3dbars:
Addressed all 3 P2s from review:

1. **Wake-kick budget exhaustion** — `SCKAudioCapture` now resets `recoveryAttempts` to `0` on every confirmed successful recovery (reactive failure or proactive wake kick alike), mirroring `Audio.finalizeMicRecoveryArtifacts` resetting `recoveryAttemptCount` on the mic path. **Rule: a successful recovery — wake-triggered or reactive — always restores the full attempt budget, so a wake kick can never reduce the recovery capacity available to a later genuine failure in the same recording.**

2. **Gap under-measurement** — the gap duration is now measured from `SCKAudioCapture`'s tracked last-real-buffer timestamp (`lastKnownBufferTime()`, the same value the buffer-stall watchdog compares against), captured before any recovery step resets it — not from when `handleMidRecordingFailure` happened to be called (which was already after the ≥5s watchdog silence). The residual "success" ≠ "real frame resumed" divergence from mic semantics (SCK has no `waitForMicBuffer` equivalent) is documented in a code comment and the PR body rather than silently claimed as fixed.

3. **`deviceSwitchCount` lost-update race** — added `Audio.incrementDeviceSwitchCount()`, a single-lock-acquisition read-modify-write, and routed both the mic-path and SCK-path increment sites through it. Checked `recordingGaps`' append path — already atomic (single lock acquisition), no change needed there.

New tests for all three (budget-reset, last-buffer-time semantics, and two concurrency tests for the race). Verified the concurrency test actually catches the bug: temporarily reverted the atomic increment locally and confirmed the test failed with real lost updates (`1601 != 4000`) before restoring the fix.

Verification (foreground, all green):
- `bash build.sh --no-open`
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12735/12735 passed
- `bash run-integration-smoke.sh` — passed
- `swift test` — 964/964 passed, 13 skipped, 0 failures

Not merged — pushed to `claude/w3-sck-recovery-parity`.

PR #1640, opened 2026-08-05
Unify the three default-input-device listeners into one monitor

r3dbars:
**Update (commit 60d50073):** fixed two P2s from codex review — the monitor no longer globally swallows self-write notifications (it now delivers every notification to every subscriber with an `isSelfWrite` flag, and each consumer decides for itself), and a `MicActivityMonitor` start→stop→start subscription race that could leak an observer. Details in the [review-fix PR comment](https://github.com/r3dbars/transcripted/pull/1640#issuecomment-5186173639) and reflected below (design bullet 2, migration table, verification counts).

## Summary

Three subsystems independently registered `AudioObjectAddPropertyListenerBlock` for `kAudioHardwarePropertyDefaultInputDevice`, with no CoreAudio-guaranteed ordering across the registrations:

- `PersistentDictationInputController.installDefaultInputListener`
- `ParakeetEngine.installInputDeviceChangeListenerIfNeeded` (`ParakeetDeviceRecovery.swift`)
- `MicActivityMonitor.installListeners`

`PersistentDictationInputController.applyCurrentPreference` also **writes** the same property (`CoreAudioInputDeviceLookup.setDefaultInputDeviceID`), which re-fires all three listeners, including itself. Each consumer had grown its own defensive workaround for that feedback loop (a recovery-marker short-circuit, `ParakeetEngine`'s `ignoreInputSelectionConfigChangesUntil`, `MicActivityMonitor`'s tolerance for a harmless extra re-scan) — audit-verified, `ParakeetDeviceRecovery.swift` even carried a comment documenting a past feedback loop.

This PR adds `Sources/Speech/DefaultInputDeviceMonitor.swift`: the single `@MainActor` facade that owns the one CoreAudio registration, applies self-write suppression centrally, and delivers to subscribers in a defined (registration) order. All three consumers now subscribe instead of registering their own listener; the per-consumer registration code is deleted.

## Design

- **Single registration**: `DefaultInputDeviceMonitor.start()` registers `AudioObjectAddPropertyListenerBlock` for `kAudioHardwarePropertyDefaultInputDevice` directly against `DispatchQueue.main` (idempotent — safe for every subscriber to call from its own startup path).
- **Self-write classification, not global suppression**: `setDefaultInputDevice(_:)` is the only sanctioned way to write the property through the monitor. It records the written device ID and a deadline (`TranscriptedConstants.defaultInputDeviceMonitorSelfWriteWindow`, 500ms — see rationale in that constant's doc comment). The classification is single-use: the pending write is consumed by the first notification observed after it, whether or not that notification's current device matches, so a stale pending write can never mislabel a later, unrelated change. **Every notification is still delivered to every subscriber** — `addObserver`'s handler receives `isSelfWrite: Bool` and each consumer decides for itself whether to react (see the migration table). An earlier version of this PR had the monitor drop self-write notifications globally before delivery, which silently starved `MicActivityMonitor` (which never writes the property and relied on seeing *every* change) — fixed per codex review.
- **Ordered delivery**: `addObserver`/`removeObserver` maintain a simple append-ordered array (`DefaultInputDeviceObserverRegistry`) and deliver strictly in registration order. Documented known runtime order: `PersistentDictationInputController` (registered in `applicationDidFinishLaunching`, before anything else touches audio) → `ParakeetEngine` (registered the first time dictation warmup or recording calls `prewarm()`) → `MicActivityMonitor` (registered when meeting-prompt detection is enabled, can start/stop repeatedly across a session).
- **Threading**: CoreAudio delivers the listener callback on whatever queue it's registered against; the monitor registers directly on `DispatchQueue.main`, so there's exactly one hop system-wide. `MicActivityMonitor`'s documented contract confines CoreAudio work to its own private serial queue, so its subscription handler re-hops from the monitor's main-actor delivery back onto that queue before touching anything — a second, consumer-owned hop, not one the monitor itself performs.
- **Fast-test isolation**: the pure observer-registry and self-write-tracker types live in `Sources/Speech/DefaultInputDeviceMonitorSupport.swift` (no `EventReporter`, no `CoreAudioInputDeviceLookup`) so `scripts/entrypoints/run-tests.sh`'s raw-`swiftc` fast-test compile unit can include them directly. `MicActivityMonitor.swift` is already in that fast-test source list (for its pure attribution-helper tests); rather than pull `DefaultInputDeviceMonitor.swift`'s `EventReporter` dependency graph into that compile unit, `MicActivityMonitor` takes the monitor via an injected `DefaultInputDeviceSubscribing` protocol property (`defaultInputDeviceMonitor`), wired up in `TranscriptedApp.swift` (not fast-tested) before `start()` is called — same "assign before start()" contract as its existing `onChange`/`onOutputChange` closures.

## Per-consumer migration table

| Consumer | Before | After | isSelfWrite policy | Why |
|---|---|---|---|---|
| `PersistentDictationInputController` | Owned its own `AudioObjectAddPropertyListenerBlock`/`AudioObjectRemovePropertyListenerBlock` pair on `DispatchQueue.main`. Wrote the property directly via `CoreAudioInputDeviceLookup.setDefaultInputDeviceID` (4 call sites: apply, restore, and two unclean-exit recovery paths). Its only defense against its own write was a recovery-marker short-circuit in `applyCurrentPreference` that skips *reapplying* an already-satisfied preference — it did not stop the notification from firing or from reaching the other two listeners. | Subscribes via `DefaultInputDeviceMonitor.shared.addObserver`. All 4 write call sites now go through `DefaultInputDeviceMonitor.shared.setDefaultInputDevice(_:)`. | **Ignore** (`guard !isSelfWrite else { return }`) | This is the write source. Reacting to its own echo is exactly the write-triggers-notification-triggers-reconcile feedback loop this migration exists to fix. Net effect vs. before: one fewer wasted debounced reconcile pass per preference-apply (previously the notification *did* fire and reach `scheduleTopologyRefresh`, just did no-op work once inside `applyCurrentPreference` thanks to the recovery-marker check). That recovery-marker short-circuit is untouched — it still exists for the unclean-exit-recovery case, where there's no in-memory pending write to classify. |
| `ParakeetEngine` / `ParakeetDeviceRecovery.swift` | Owned `installInputDeviceChangeListenerIfNeeded`/`removeInputDeviceChangeListener`, registering on `.main` directly. Read the CoreAudio selection off-main via `Task.detached(priority: .utility)` before hopping back to `@MainActor` for `handleDefaultInputDeviceChange`. | Subscribes via the monitor; the same `Task.detached` → `handleDefaultInputDeviceChange` sequencing is preserved unchanged, just triggered by the monitor's callback instead of a raw listener block. | **Ignore** | Never writes the property itself, so `isSelfWrite == true` only ever means "`PersistentDictationInputController` just reasserted its preference" — this engine has no reason to run device-recovery machinery in reaction to that, matching its pre-migration behavior of not reacting to that controller's writes. Its *own* default-input overrides during recording start are a separate concern entirely: they don't go through `setDefaultInputDevice` at all, so they always arrive as `isSelfWrite == false` and are guarded instead by this engine's existing `ignoreInputSelectionConfigChangesUntil` window (checked in `handleAudioConfigChange`), unchanged — that window spans the whole recording-start sequencing around the override, not just the CoreAudio round trip, so it was intentionally kept per-consumer rather than centralized. |
| `MicActivityMonitor` | Owned a listener via its generic `addSystemListener(_:extra:)` helper, registered directly on its own private serial `queue` (not main), with an `extra` closure calling `reattachDeviceListener()`. | Subscribes via the injected `defaultInputDeviceMonitor` protocol property; the handler re-hops onto `queue` before calling `reattachDeviceListener()` + `scheduleDebouncedScan()` (mirrors the old `extra` + fallthrough `scheduleDebouncedScan()` behavior). | **Handle unconditionally** (ignores the flag) | Never wrote the property, and pre-migration saw *every* notification — including `PersistentDictationInputController`'s self-reassertion writes — always running `reattachDeviceListener()` in reaction. An earlier version of this PR had the monitor globally swallow self-writes, which silently starved this consumer of events it relied on and degraded call detection to the multi-second backstop poll; fixed per codex review by delivering every notification with the flag instead of dropping it. Reacting to a self-write here is, and always was, harmless read-only housekeeping — it just re-points the "running somewhere" listener at whatever the current default input is. The remaining behavior change vs. pre-migration is just an extra dispatch-queue turnaround (sub-millisecond) between "CoreAudio noticed" and "MicActivityMonitor reacts," ahead of its existing 1s `scheduleDebouncedScan` debounce — not observable. |

`MicActivityMonitor` also picked up a fix for a start→stop→start subscription race: its subscribe call captures a `queue`-confined `startGeneration` before hopping to the main actor, and re-checks `isStarted && currentGeneration == capturedGeneration` (pulled into a pure, unit-tested `MicActivityMonitor.SubscriptionOutcome.decide`) before storing the resulting token — a stale subscription from a superseded `start()` cycle is unsubscribed immediately instead of risking overwriting the live one and leaking an observer that fires while conceptually stopped.

Also deleted: the now-dead `unregisterDefaultInputDeviceListener` helper in `ParakeetAudioEngineSupport.swift` (only caller was the old `ParakeetEngine` deinit path, now migrated). `ParakeetEngine`'s `deinit` unregisters via a fire-and-forget `Task { @MainActor in ... }` hop since `deinit` is nonisolated even for an `@MainActor` class — safe because the observer closure captures `self` weakly, so any notification landing before the hop completes is already a no-op.

## Verification (synchronous, foreground, in this worktree)

- `bash build-deps.sh --force` (stamp was stale after `git reset --hard origin/main`) — succeeded
- `bash build.sh --no-open` — succeeded, no new warnings introduced by this change (confirmed by grepping build output for the touched files/symbols; the two pre-existing warnings in `MeetingSessionController.swift`/`DictationSessionController.swift` are unrelated)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — **12757/12757 passed**, including `Tests/DefaultInputDeviceMonitorTests.swift` (observer-registration ordering/removal, isSelfWrite pass-through to every observer regardless of policy, and self-write classification: match-inside-window, no-match-after-window, no-match-for-a-different-device, single-use consumption on both match and mismatch, cancellation on a failed write, no-pending-write baseline) and `Tests/MicActivityMonitorTests.swift`'s `MicActivityMonitor.SubscriptionOutcome` cases (common case, stop-before-hop-resolves, the exact start→stop→start race, and the live cycle's own subscription still storing after unrelated churn)
- `bash run-integration-smoke.sh` — passed (app/core linkage, `MicRecordingFileMergerTests`, wake recovery)

## Hardware checklist — NOT YET DONE, do not merge before this

This PR **must not merge** until a local hardware pass on real hardware confirms no regression in the three migrated flows. None of the above verification exercises live CoreAudio device-change notifications end-to-end (the fast-test suite is pure-logic only; `run-integration-smoke.sh` doesn't drive a real device switch). Needed before merge:

- [ ] Switch the system default input device (e.g. built-in mic ↔ USB interface) while dictation is idle/prewarmed — confirm `ParakeetEngine` recovers and dictation still starts cleanly
- [ ] Switch the system default input device while a dictation recording is in progress — confirm recording recovers on the new device (or fails gracefully per the existing retry budget)
- [ ] Connect/disconnect AirPods (or another Bluetooth headset) during dictation — confirm the existing HFP/A2DP route-settling behavior is unchanged, and that `ignoreInputSelectionConfigChangesUntil` still suppresses `ParakeetEngine`'s own override-triggered notification
- [ ] Enable the "faster Bluetooth dictation" preference (`PersistentDictationInputController`) and confirm it pins the recommended input, that the resulting notification is swallowed (no redundant reconcile pass, no visible flicker), and that disabling the preference (or an unclean exit) still restores the prior input correctly
- [ ] With `AutoCallDetectionPreferences` enabled, start a call in a supported app while switching the default input device — confirm `MicActivityMonitor` still re-points its "running somewhere" listener and prompts correctly
- [ ] Toggle `AutoCallDetectionPreferences` off/on (which start/stops `MicActivityMonitor` repeatedly) a few times in a row and confirm no duplicate-subscription buildup on `DefaultInputDeviceMonitor` (no double-firing of `reattachDeviceListener`/`scheduleDebouncedScan` per real device change)

## Sequencing note

Per instructions: **do not merge this before PR #1636 and the SCK-parity PR**, to avoid rebase churn on `Sources/TranscriptedCore/Audio/Audio.swift`. This PR does not touch that file — `Sources/TranscriptedCore/Audio/CoreAudioUtils.swift` only *reads* `kAudioHardwarePropertyDefaultInputDevice` (no listener registration), so there's no direct overlap — but flagging in case either of those PRs touches related CoreAudio device-change plumbing.


r3dbars:
Addressed both P2s from the codex review in commit 60d50073.

## 1. Global self-write suppression hid a transition MicActivityMonitor relied on

`DefaultInputDeviceMonitor` no longer decides who cares about a self-write. `handlePropertyChanged` now calls `registry.notifyAll(isSelfWrite:)` unconditionally — every subscriber sees every change — and `addObserver`'s handler signature changed from `() -> Void` to `(Bool) -> Void` so each consumer gets told whether the notification was caused by `PersistentDictationInputController`'s own `setDefaultInputDevice(_:)` write.

Per-consumer `isSelfWrite` policy (also documented inline at each subscription site):

| Consumer | Policy | Why |
|---|---|---|
| `PersistentDictationInputController` | Ignore (`guard !isSelfWrite else { return }`) | This is the write source; reacting to its own echo is exactly the feedback loop the original migration exists to prevent. |
| `ParakeetEngine` / `ParakeetDeviceRecovery` | Ignore | Never writes the property itself, and has no reason to run device-recovery machinery in reaction to `PersistentDictationInputController` reasserting its own preference — matches its pre-migration behavior of not reacting to that controller's writes. Its own default-input overrides during recording start are unrelated (they don't go through `setDefaultInputDevice` at all, so they always arrive as `isSelfWrite == false`, and are still guarded by its existing `ignoreInputSelectionConfigChangesUntil` window, unchanged). |
| `MicActivityMonitor` | Handle unconditionally (ignore the flag) | Never wrote the property and pre-migration saw *every* notification, always running `reattachDeviceListener()` in reaction. Reacting to a self-write here is harmless, read-only housekeeping (re-pointing the "running somewhere" listener), and dropping it silently starved call detection down to the multi-second backstop poll — the exact regression flagged. |

`DefaultInputDeviceSelfWriteTracker.consumeSuppression` renamed to `consumeIsSelfWrite` to match — it now classifies rather than gates.

## 2. MicActivityMonitor start→stop→start token race

Added a `queue`-confined `startGeneration` counter, bumped on every `start()`/`stop()` transition. `subscribeToDefaultInputDeviceChanges` captures the generation before hopping to the main actor to call `DefaultInputDeviceMonitor.addObserver`; when the hop resolves back on `queue`, it re-checks `isStarted && currentGeneration == capturedGeneration` (pulled into a pure `MicActivityMonitor.SubscriptionOutcome.decide` for testability) before storing the token. A generation mismatch means a `stop()` (and possibly a subsequent `start()`) ran while the hop was in flight, so the subscription is stale — it's unsubscribed immediately instead of stored, closing the leak where a superseded subscription could win the race and overwrite the live one.

## Tests

- `Tests/DefaultInputDeviceMonitorTests.swift`: added isSelfWrite pass-through coverage (both `true`/`false` reach every observer; a mixed policy — one consumer ignoring, one always reacting — behaves as intended) plus renamed the tracker assertions to match `consumeIsSelfWrite`.
- `Tests/MicActivityMonitorTests.swift`: added 4 cases for `MicActivityMonitor.SubscriptionOutcome.decide` covering the common case, a plain stop-before-hop-resolves case, the exact start→stop→start race (`isStarted: true, currentGeneration: 3, capturedGeneration: 1` → `.dropStale`), and confirming the live cycle's own subscription still stores after unrelated earlier churn.

## Verification (foreground)

- `bash build.sh --no-open` — succeeded, no new warnings
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — **12757/12757 passed**
- `bash run-integration-smoke.sh` — passed

Still not merging — the hardware checklist in the PR description stands, and per the original instructions this branch is not to be merged before PR #1636 and the SCK-parity PR to avoid `Audio.swift` rebase churn.

Issue #1624, opened 2026-08-04
Meeting recording always fails on 1.1.51: mic route reports device_mismatch even when sample rates match


MyaLongmire:
## Summary

Since 1.1.51, meeting recording never starts on my Mac. Every attempt fails right away with:

> Recording failed to start: The microphone route did not become ready. Check your input device and try again.. Try quitting and reopening Transcripted.

I have 20 failed starts in `events.jsonl` and zero successful ones. The last time a recording worked was 2026-07-15 on 1.1.50 / macOS 26.5.2.

The microphone itself is fine. People hear me in Google Meet without any problem, and other apps record fine.

## Environment

- Transcripted 1.1.51 (build `10b6eac5c0f3`, release channel)
- macOS 26.6 (build 25G72)
- MacBook Pro, Apple Silicon
- Model: `parakeet-tdt-v3`
- Default output/input: AirPods Pro (Bluetooth, 24 kHz) in most tests
- Third-party CoreAudio HAL drivers installed: `ARK.driver` 13.1.3 (Rogue Amoeba), `eqMac.driver` 2.6.0, `MSTeamsAudioDevice.driver`

## What the logs show

Every failure looks the same. `mic_raw_peak` and `mic_processed_peak` are exactly `0.00000`, and `voice_processing_active` stays `false` while `voice_processing` is `true`.

From `logs/app.jsonl`, with AirPods Pro as output:

```
"Meeting input pinned to selected microphone"
  selectedTransport=builtIn  reason=preferredBuiltInForBluetoothHeadset
"Voice processing enabled on meeting mic"
  ducking=min  agc=true  reason=issue_500_safari_firefox_vpio_contention
"Meeting microphone route did not settle"
  outcome=device_mismatch  selectedNominalRate=48000.0  capturedRate=24000.0
"Meeting microphone graph attempt failed"
  error=The microphone route did not become ready. Check your input device and try again.
```

At first this looked like a sample rate problem: the app pins the built-in mic at 48 kHz, but Apple Voice Processing is a combined input + output unit, so the graph runs at the Bluetooth output rate of 24 kHz.

But then I switched system output to MacBook Pro Speakers (48 kHz) and tried again. Now the rates match, and it still fails the same way:

```
"Voice processing enabled on meeting mic"
  ducking=min  agc=true  reason=issue_500_safari_firefox_vpio_contention
"Meeting microphone route did not settle"
[...]

PR #1641, opened 2026-08-05
Thread VPIO bind result into meeting route readiness

r3dbars:
## Verify-first: what remained vs. what yesterday's fix already covered

This is the final new-code PR of the simplification program, targeting the
`Audio.voiceProcessingEnabled` ambient-cache audit finding. Before writing
any diff I re-read `Sources/TranscriptedCore/Audio/Audio.swift`'s VPIO code
in full and the two same-day commits from **release 1.1.52 / issue #1624**
(`c373bdc4` "Fix voice-processed meeting mic readiness" and `6137020d` "Fix
meeting VPIO route readiness", merged yesterday as PR #1616).

**Already fixed by 1.1.52 (verified, untouched here):** the field-critical
bug — `routeReadiness` comparing device identity *after* VPIO's wrap, which
broke meeting recording on machines with third-party HAL drivers — is fixed.
`routeReadiness` now takes `boundInputDeviceIDBeforeVoiceProcessing` as an
explicit parameter and never re-derives identity from the post-wrap node.
That parameter, its guard logic, and `MeetingInputDeviceSelectionPolicy`'s
public shape are **unchanged** in this PR.

**What remained:** the fix landed the pre-wrap device-ID capture as two
separate reads sitting right next to the `armVoiceProcessing(on:)` call —
```swift
let boundInputDeviceIDBeforeVoiceProcessing = freshInputNode.auAudioUnit.deviceID
armVoiceProcessing(on: freshInputNode)
...
voiceProcessingEnabled: voiceProcessingEnabled   // ambient re-read
```
This is exactly the "ambient cache read immediately after a call that just
computed the same thing" pattern the audit flagged — correct today only
because nothing runs between the two reads under `withAudioGraphLock`, but
an implicit invariant rather than an enforced one.

The broader audit finding — `armVoiceProcessing`/`disarmVoiceProcessing`'s
double-condition "distrust our own cache" checks (`if voiceProcessingEnabled,
inputNode.isVoiceProcessingEnabled`, `wasMarkedEnabled` vs.
`wasActuallyEnabled`) — **still exists** but is not a bug: it's a defensive
OR of both signals that never causes a wrong device/route decision, and arm
and disarm are genuinely temporally separated (arm at start, disarm at stop,
reached through different call chains — monitoring stop, discard-on-retry,
device recovery). Collapsing that into a single "one true value" would mean
either threading a bind result across an async gap (i.e., re-inventing the
ambient cache under a new name) or reworking arm/disarm's decision logic in
the most field-sensitive, currently-unconfirmed part of the codebase. Given
the explicit zero-behavior-change bar and pending field confirmation for
1.1.52, I did **not** touch that part — doing so would be manufacturing a
bigger diff than the remaining problem justifies. Noted here for whoever
picks this up next.

## What this PR does

`Audio.armVoiceProcessing(on:)` now returns a `VoiceProcessingBindResult`
(`boundInputDeviceIDBeforeWrap`, `enabled`) instead of `Void`. The device ID
is captured as the function's first statement — the same value the caller
used to capture immediately before calling it — and `enabled` is set at
every return point to the same value the caller used to read from
`voiceProcessingEnabled` immediately after the call returned. Because
nothing executes between those paired reads in either the old or new code
(same synchronous, lock-held scope), **this is a provably value-identical
refactor**: no scenario changes its computed values.

`makeReadyMeetingInputGraph` now threads that one return value into both
`recordingFormat(for:voiceProcessingEnabled:)` (new optional override
parameter, defaults to the existing ambient-read behavior for other
callers) and `MeetingInputDeviceSelectionPolicy.routeReadiness(...)` —
`routeReadiness`'s only two VPIO/device-identity inputs — instead of
separately capturing the device ID beforehand and re-reading the ambient
cache after.

### Reader migration table

| Reader of `voiceProcessingEnabled` | Migrated? | Why |
|---|---|---|
| `makeReadyMeetingInputGraph` → `routeReadiness(...)` | **Yes** | The exact call site 1.1.52 touched; now sourced from `armVoiceProcessing`'s return value |
| `makeReadyMeetingInputGraph` → `recordingFormat(for:)` | **Yes** | Same synchronous window as the arm call; new optional param, zero behavior change |
| `disarmVoiceProcessing` (`wasMarkedEnabled`/`wasActuallyEnabled`) | No | Genuinely cross-time (disarm runs later, in a different call chain); the dual-signal check is defensive, not buggy |
| `armVoiceProcessing`'s own early-return check (`voiceProcessingEnabled, inputNode.isVoiceProcessingEnabled`) | No | Same reasoning; not the site of the field bug |
| `startMonitoring` → `recordingFormat(for:)` | No | Monitoring never arms VPIO — no bind result exists in scope; correctly still ambient |
| `AudioFileManager.startAudioCapture` log line (`"voiceProcessing": "\(voiceProcessingEnabled)"`) | No | Runs after `makeReadyMeetingInputGraph` returns; would require widening `PreparedMeetingInputGraph`'s public shape for a log string, no behavior change to justify it |
| `AudioPipelineDiagnosticsSnapshot.voiceProcessingActive` (health/events) | No | Deliberately reads *current live* state for diagnostics/Sentry context, not bind-time state — ambient is the correct source here |
| `AudioFileManager`/other "shared-mic paths" | N/A | Grepped the whole repo; no external readers exist outside `Audio.swift` itself + these two files + tests |

No public API outside `TranscriptedCore` changes. `MeetingInputDeviceSelectionPolicy.routeReadiness`'s signature and logic are untouched.

## Tests

Added `Tests/TranscriptedCoreTests/AudioTests/VoiceProcessingBindResultTests.swift`:
- `VoiceProcessingBindResult` Equatable shape.
- `routeReadiness` fed bind-result-shaped values reproduces the exact 1.1.52 scenario (`.ready` for a bound-before-wrap device that VPIO then wraps) and the mismatch case (`.deviceMismatch` for a bind result carrying the wrong pre-wrap device ID even with `enabled: true`).

`armVoiceProcessing` itself needs a live `AVAudioInputNode`, which this test
directory has never constructed directly (no precedent in
`AudioStateTransitionTests`/`AudioDiagnosticsSnapshotTests`/etc.) — kept
consistent with that convention rather than introducing a first hardware-
dependent unit test.

## Verification (all synchronous foreground)

- `bash build-deps.sh --force` — pass
- `bash build.sh --no-open` — pass (signed, launch smoke OK, 1788.8ms launch-to-interactive, within 3000ms budget)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12757 tests, 12757 passed, 0 failed
- `bash run-integration-smoke.sh` — pass (core smoke, wake smoke, recovery merge package tests)
- `swift test` — 967 tests, 13 skipped (macOS-version gated), 0 failed

## Concerns / for the reviewer

- This touches the same function (`makeReadyMeetingInputGraph`) and exact
  values (`boundInputDeviceIDBeforeVoiceProcessing`, `voiceProcessingEnabled`)
  as yesterday's unconfirmed field fix for issue #1624. I verified the
  refactor is value-identical by inspection (same lock scope, no
  intervening code in either version) and all four verification suites
  pass, but this has **not** been confirmed against real third-party-HAL
  hardware — that confirmation is still pending from the original reporter
  and is out of scope for what a code review can prove.
- The remaining ambient-cache/double-condition pattern in
  `armVoiceProcessing`/`disarmVoiceProcessing` is intentionally left alone
  (see "what remained" above) — flagging in case a future pass wants to
  revisit it once 1.1.52 has real field confirmation and there's more
  appetite for risk in this file.

### DIFF
--- a/Sources/TranscriptedCore/Audio/Audio.swift
+++ b/Sources/TranscriptedCore/Audio/Audio.swift
@@ -491,6 +491,16 @@ public class Audio: ObservableObject, @unchecked Sendable {
     // WebRTC activates AUVoiceProcessingIO on the shared input device which
     // hands every other reader an attenuated stream; running our own VPIO
     // gives us our own AGC'd copy.
+    //
+    // This is the cross-time cache: it is what makes VPIO state visible to
+    // work that happens well after the arm call returned (disarm at stop,
+    // live diagnostics snapshots, log lines) where no synchronous return
+    // value is available. Callers that run immediately alongside
+    // `armVoiceProcessing(on:)` — same lock scope, nothing intervening —
+    // should prefer its returned `VoiceProcessingBindResult` instead of
+    // re-reading this var, so route-identity decisions have exactly one
+    // source of truth for "what did the arm call just observe." See
+    // `makeReadyMeetingInputGraph` for the canonical example.
     var voiceProcessingEnabled: Bool = false
 
     /// Whether to arm Apple's AUVoiceProcessingIO (VPIO) on the meeting mic
@@ -1404,11 +1414,18 @@ public class Audio: ObservableObject, @unchecked Sendable {
                         // replaces the node's device identity with its private
                         // wrapper. The wrapper ID is not a reliable indication
                         // of which physical input CoreAudio already bound.
+                        // `armVoiceProcessing` returns both the pre-wrap device
+                        // ID and the resulting VPIO state as one atomic value,
+                        // so route readiness and format selection below thread
+                        // that same value through instead of separately
+                        // capturing the device ID beforehand and re-reading
+                        // the ambient `voiceProcessingEnabled` cache after.
                         let selection = meetingInputSelectionSnapshot()
-                        let boundInputDeviceIDBeforeVoiceProcessing =
-                            freshInputNode.auAudioUnit.deviceID
-                        armVoiceProcessing(on: freshInputNode)
-                        let recordingFormat = self.recordingFormat(for: freshInputNode)
+                        let voiceProcessingBind = armVoiceProcessing(on: freshInputNode)
+                        let recordingFormat = self.recordingFormat(
+                            for: freshInputNode,
+                            voiceProcessingEnabled: voiceProcessingBind.enabled
+                        )
                         guard let recordingSnapshot = AudioRecordingFormatPolicy.snapshot(recordingFormat) else {
                             throw NSError(
                                 domain: "Audio",
@@ -1423,11 +1440,11 @@ public class Audio: ObservableObject, @unchecked Sendable {
                         let actualInputDeviceID = freshInputNode.auAudioUnit.deviceID
                         let routeReadiness = MeetingInputDeviceSelectionPolicy.routeReadiness(
                             selection: selection,
-                            boundInputDeviceIDBeforeVoiceProcessing: boundInputDeviceIDBeforeVoiceProcessing,
+                            boundInputDeviceIDBeforeVoiceProcessing: voiceProcessingBind.boundInputDeviceIDBeforeWrap,
                             actualInputDeviceID: actualInputDeviceID,
                             capturedSampleRate: recordingSnapshot.sampleRate,
                             selectedNominalSampleRate: selectedNominalRate,
-                            voiceProcessingEnabled: voiceProcessingEnabled
+                            voiceProcessingEnabled: voiceProcessingBind.enabled
                         )
                         guard routeReadiness == .ready else {
                             AppLogger.audioMic.warning("Meeting microphone route did not settle", [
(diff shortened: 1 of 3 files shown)


r3dbars:
Both findings fixed in `46e3d7d2`, same branch.

**P1 (hard-coded `enabled: false`)** — confirmed correct. `armVoiceProcessing`'s preference-off branch now captures `engine?.isRunning` and the ambient `voiceProcessingEnabled` **before** calling `disarmVoiceProcessing`, then computes the returned `enabled` value through a new pure function, `Audio.voiceProcessingEnabledAfterDisarmAttempt(engineIsRunning:priorVoiceProcessingEnabled:)`:

```swift
engineIsRunning ? priorVoiceProcessingEnabled : false
```

This exactly mirrors `disarmVoiceProcessing`'s real branching: it returns immediately (cache untouched) when the engine is running, and unconditionally sets `voiceProcessingEnabled = false` on every other path (both the already-disabled early return and the full disable attempt, success or failure). So for the scenario you flagged — preference toggled off mid-session, VPIO still armed, engine running — the bind result now correctly carries `enabled: true`, matching what the ambient read on `main` would have observed, instead of the hard-coded `false` that would have flipped `recordingFormat`/`routeReadiness`.

**P2 (arm-level test coverage)** — went with the pure-function fallback: `armVoiceProcessing`/`disarmVoiceProcessing` need a live `AVAudioInputNode`, and this test target has no precedent for constructing a real `AVAudioEngine` (checked `AudioStateTransitionTests`, `AudioDiagnosticsSnapshotTests`, etc. — none of them touch `.inputNode`). Extracted the branch's decision into the pure function above and pinned it in `Tests/TranscriptedCoreTests/AudioTests/VoiceProcessingDisarmOutcomeTests.swift`:
- `testEngineRunningLeavesPriorCacheValueUnchanged` — the exact regression case (engine running, prior `true` → stays `true`; prior `false` → stays `false`).
- `testEngineStoppedAlwaysClearsToFalseRegardlessOfPriorValue` — engine stopped always clears to `false`.

Production code (`armVoiceProcessing`) now actually calls this pure function rather than just being documented alongside it, so the test is exercising the real decision path, not just a parallel spec. Residual gap, documented in the test file's header comment: these tests still don't exercise `disarmVoiceProcessing`'s real CoreAudio calls or `armVoiceProcessing`'s full branch dispatch end-to-end — if `disarmVoiceProcessing`'s own branching on `engine.isRunning`/`voiceProcessingEnabled` ever changes without updating this mirror, these tests won't catch the drift. Left a "keep in sync" note on the pure function's doc comment for future maintainers.

## Verification (foreground)

- `bash build-deps.sh --force` — pass
- `bash build.sh --no-open` — pass (signed, launch smoke OK)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12757 tests, 12757 passed, 0 failed
- `bash run-integration-smoke.sh` — pass (core smoke, wake smoke, recovery merge package tests)
- `swift test` — 969 tests, 13 skipped, 0 failed (2 new tests added, both passing)

Not merging — leaving that to the coordinator per instructions.

PR #1642, opened 2026-08-05
Replace shared-meeting-mic Bool with a session-bound claim handshake

r3dbars:
## Summary

`ParakeetEngine.sharedMeetingMicRecording` was a bare `var Bool` flipped from 8+ MainActor call sites with no versioning and no tie to the meeting side's own session identity. Two device-recovery guards (`handleSystemWake`, `ParakeetDeviceRecovery.handleAudioConfigChange`) trusted it with no expiry: if the meeting session that set it died without its own clear path running (capture-stack crash, error teardown ordering), dictation's device-recovery stayed suppressed forever.

This replaces the Bool with `ParakeetEngine.sharedMeetingMicClaim: SharedMeetingMicClaim?` — a claim minted by `MeetingSessionController.startDictationFromActiveMeetingMic()`, bound to `activeRecordingIdentity` (the narrowest stable per-recording identity introduced by #1634's state collapse) and carrying a weak-captured liveness probe into `isCaptureSessionActive` (also from #1634 — documented there as "the answer every … gate should use"). Speech never imports Meeting types; the probe closure is the entire contract.

**The staleness rule (one sentence):** the two device-recovery guards resolve the claim through `resolveSharedMeetingMicClaimStatus()`, which treats a claim whose minting meeting session no longer reports itself alive as `.absent` — releasing it and reporting a privacy-safe `shared_meeting_mic_claim_stale` event — so recovery runs instead of staying suppressed forever; every other shared-mic call site keeps testing raw presence (`sharedMeetingMicClaim != nil`) unchanged, because a claim on file, dead or alive, still means there is no local `AVAudioEngine` to tear down.

## Site migration table

| Site | Old | New |
|---|---|---|
| Declaration (ParakeetEngine.swift) | `var sharedMeetingMicRecording = false` | `var sharedMeetingMicClaim: SharedMeetingMicClaim?` |
| `startSharedMeetingMicRecording()` | `sharedMeetingMicRecording = true` | `func startSharedMeetingMicRecording(claim:)`, `sharedMeetingMicClaim = claim` |
| `finishSharedMeetingMicRecording` | `sharedMeetingMicRecording = false` | `sharedMeetingMicClaim = nil` |
| `resetAfterFailedRecordingStart()` | `sharedMeetingMicRecording = false` | `sharedMeetingMicClaim = nil` |
| `abandonBlockedRecordingStart()` | `sharedMeetingMicRecording = false` | `sharedMeetingMicClaim = nil` |
| `cancel()` | `sharedMeetingMicRecording = false` | `sharedMeetingMicClaim = nil` |
| `cleanup()` | `sharedMeetingMicRecording = false` | `sharedMeetingMicClaim = nil` |
| `isRecordingFromSharedMeetingMic` (public read) | `sharedMeetingMicRecording` | `sharedMeetingMicClaim != nil` (presence-only, unchanged behavior) |
| `updateSharedMeetingMicAudioLevel` guard | `guard sharedMeetingMicRecording else { return }` | `guard sharedMeetingMicClaim != nil else { return }` (presence-only) |
| `resumeRegularRecordingAfterSharedMeetingMicEndedIfNeeded` guard | `guard sharedMeetingMicRecording else { return }` | `guard sharedMeetingMicClaim != nil else { return }` (presence-only) |
| `performStopRecording` (2 checks) | `if sharedMeetingMicRecording ...` | `if sharedMeetingMicClaim != nil ...` (presence-only) |
| `handleSystemWake()` guard **(staleness-aware)** | `ParakeetSystemWakePolicy.decision(sharedMeetingMicRecording: sharedMeetingMicRecording)` | `ParakeetSystemWakePolicy.decision(sharedMeetingMicRecording: isSharedMeetingMicClaimCurrent)` |
| `ParakeetDeviceRecovery.handleAudioConfigChange()` guard **(staleness-aware)** | `if sharedMeetingMicRecording { ... }` | `if isSharedMeetingMicClaimCurrent { ... }` |
| Meeting-side mint (`MeetingSessionController.startDictationFromActiveMeetingMic()`) | `sttRouter.startRecordingFromSharedMeetingMic()` | mints `SharedMeetingMicClaim(sessionIdentity: activeRecordingIdentity, isSessionAlive: { [weak self] in self?.isCaptureSessionActive ?? false })`, threaded through `STTRouter.startRecordingFromSharedMeetingMic(claim:)` |

Only the two guards marked **(staleness-aware)** get new behavior (stale → treated as absent + logged). Every presence-only site is behavior-identical to the old Bool.

## `SharedMeetingMicTransitionState`

Left as-is — it guards a different concern (fencing stale async continuations across `await` boundaries within a single resume attempt), not claim ownership across meeting sessions. Documented the relationship in `SharedMeetingMicRecorder.swift`'s header.

## SupersessionEpoch/ClaimSlot

Considered but not used: `SupersessionEpoch` answers "did something newer supersede this" via an owner-side monotonic bump, which can't detect a dead owner that never bumps anything (the exact failure mode being fixed). The claim instead carries a live state-machine-derived liveness probe (`isCaptureSessionActive`), which is the authoritative "is this session still alive" check already established by #1634 and documented there as the canonical gate for this kind of question.

## Tests

`Tests/SharedMeetingMicClaimTests.swift` — pure decision coverage for `SharedMeetingMicClaimStatus`/`SharedMeetingMicClaimPolicy.isSharingActive` across current/stale/absent, plus composition with `ParakeetSystemWakePolicy.decision` to confirm the wake guard's outcome for each status. Added `Sources/Speech/SharedMeetingMicClaim.swift` to `run-tests.sh`'s `APP_SOURCES`.

## Verification

- `bash build-deps.sh --force` (twice — rebased onto main after #1639/SCK-parity merged mid-flight, which touched `Sources/TranscriptedCore/Audio/**`)
- `bash build.sh --no-open` — pass
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12764 tests, 12764 passed, 0 failed
- `bash run-integration-smoke.sh` — pass (app/core linkage, wake recovery, MicRecordingFileMerger)

All run synchronously in the foreground.

## Concerns / follow-ups

- Rebased cleanly onto main after #1639 (SCK parity) merged mid-flight; no file overlap (that PR touched `Sources/TranscriptedCore/Audio/**` only).
- The `claim_session` UUID logged in the stale-claim event is an opaque per-recording identity, not PII — consistent with other generation/session IDs already logged elsewhere (e.g. `ParakeetZombieRecoveryTerminal.generation`).
- Did not rename `ParakeetSystemWakePolicy.decision(sharedMeetingMicRecording:)`'s parameter label to avoid touching that file/its existing test beyond what's needed — it now receives the resolved (post-staleness-check) boolean rather than the raw field, semantics preserved.

### DIFF
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -256,17 +256,31 @@ final class MeetingSessionController: ObservableObject {
     /// `clearActiveRecordingIdentity()` call sites), i.e. exactly the window
     /// `canShareMicWithDictation` covers. `isSessionAlive` closes over
     /// `self` weakly so Speech never needs a reference to
-    /// `MeetingSessionController` itself, and reads `isCaptureSessionActive`
-    /// — the broader "is the capture pipeline live for this recording"
-    /// check that also covers this session's own teardown window — rather
-    /// than `isRecording`/`canShareMicWithDictation`, which would go false
-    /// the instant this session starts stopping and make every claim look
-    /// stale during completely normal teardown.
+    /// `MeetingSessionController` itself, and delegates to
+    /// `SharedMeetingMicClaimPolicy.isClaimSessionAlive`, which requires
+    /// BOTH `isCaptureSessionActive` — the broader "is the capture pipeline
+    /// live for this recording" check that also covers this session's own
+    /// teardown window, rather than `isRecording`/`canShareMicWithDictation`,
+    /// which would go false the instant this session starts stopping and
+    /// make every claim look stale during completely normal teardown — AND
+    /// that `activeRecordingIdentity` is still the exact identity minted
+    /// below (`mintedIdentity`), captured by value here so a later
+    /// recording on this same controller (a fresh identity) cannot make an
+    /// orphaned claim from an earlier, already-ended recording read as
+    /// alive just because something is active again.
     func startDictationFromActiveMeetingMic() -> Bool {
         guard canShareMicWithDictation, let activeRecordingIdentity else { return false }
+        let mintedIdentity = activeRecordingIdentity
         let claim = SharedMeetingMicClaim(
-            sessionIdentity: activeRecordingIdentity,
-            isSessionAlive: { [weak self] in self?.isCaptureSessionActive ?? false }
+            sessionIdentity: mintedIdentity,
+            isSessionAlive: { [weak self] in
+                guard let self else { return false }
+                return SharedMeetingMicClaimPolicy.isClaimSessionAlive(
+                    mintedIdentity: mintedIdentity,
+                    currentIdentity: self.activeRecordingIdentity,
+                    isCaptureSessionActive: self.isCaptureSessionActive
+                )
+            }
         )
         return sttRouter.startRecordingFromSharedMeetingMic(claim: claim)
     }
--- a/Sources/Meeting/MeetingSessionController.swift
+++ b/Sources/Meeting/MeetingSessionController.swift
@@ -248,9 +248,27 @@ final class MeetingSessionController: ObservableObject {
 
     /// Check capture ownership and arm borrowed dictation without an `await`
     /// between those actions, so meeting stop cannot slip into the handoff.
+    ///
+    /// Mints the `SharedMeetingMicClaim` dictation will hold for the
+    /// lifetime of the borrow: `activeRecordingIdentity` is always non-nil
+    /// here because it is coined before `.startingRecording` and only
+    /// cleared when this session leaves `.recording` (see
+    /// `clearActiveRecordingIdentity()` call sites), i.e. exactly the window
+    /// `canShareMicWithDictation` covers. `isSessionAlive` closes over
+    /// `self` weakly so Speech never needs a reference to
+    /// `MeetingSessionController` itself, and reads `isCaptureSessionActive`
+    /// — the broader "is the capture pipeline live for this recording"
+    /// check that also covers this session's own teardown window — rather
+    /// than `isRecording`/`canShareMicWithDictation`, which would go false
+    /// the instant this session starts stopping and make every claim look
+    /// stale during completely normal teardown.
     func startDictationFromActiveMeetingMic() -> Bool {
-        guard canShareMicWithDictation else { return false }
-        return sttRouter.startRecordingFromSharedMeetingMic()
+        guard canShareMicWithDictation, let activeRecordingIdentity else { return false }
+        let claim = SharedMeetingMicClaim(
+            sessionIdentity: activeRecordingIdentity,
+            isSessionAlive: { [weak self] in self?.isCaptureSessionActive ?? false }
+        )
+        return sttRouter.startRecordingFromSharedMeetingMic(claim: claim)
     }
 
     // MARK: - State transitions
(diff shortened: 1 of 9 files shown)


r3dbars:
Addressed both Codex review findings on the same branch.

**P1 (data loss) — fixed.** `resolveSharedMeetingMicClaimStatus()` was only clearing `sharedMeetingMicClaim` on staleness, never draining `sharedMeetingMicRecorder`, so the borrowed audio buffered since `begin()` was silently dropped. Added `finalizeStaleSharedMeetingMicClaim()`, which now routes through the same path a normal share-end takes:
1. `sharedMeetingMicTransition.invalidate()` (fence any in-flight resume, matching the other teardown sites)
2. `finishSharedMeetingMicRecording(keepRecordingState: false)` — drains `sharedMeetingMicRecorder` into `recoveredRecordingTimeline` *before* the claim is cleared (that method's first statement), then sets `isRecording = false`
3. `interruptRecordingPreservingRecoveredTimeline()` — marks the recording interrupted so the recovered audio surfaces to the user, exactly like any other capture cut off mid-recording

`keepRecordingState: false` mirrors `resumeRegularRecordingAfterSharedMeetingMicEndedIfNeeded`'s own choice, but a stale claim can't attempt `startRecording(isRecoveryAttempt:)` itself (the meeting session is dead) — it only preserves what was captured. Because `isRecording` is already `false` by the time the calling guard's own teardown logic runs, `handleSystemWake()`/`handleAudioConfigChange()` correctly skip their `if isRecording { preserveCurrentRecordingBuffersForRecovery() ... }` branch instead of touching a shared-mic recorder that no longer has anything queued.

**P2 (cross-session confusion) — fixed.** The `isSessionAlive` closure only checked `isCaptureSessionActive`, never comparing the claim's minted identity against the *current* recording identity. Added `SharedMeetingMicClaimPolicy.isClaimSessionAlive(mintedIdentity:currentIdentity:isCaptureSessionActive:)`, a pure two-condition check requiring both an active session AND identity equality. `MeetingSessionController.startDictationFromActiveMeetingMic()` now captures its own minted identity by value and delegates the closure to this check, so an orphaned claim from a superseded recording resolves stale even if a *different* recording is active — closing the "unexpected stop → async resume scheduled → new meeting starts first → old claim reads alive → new meeting's relay feeds the old recorder" race.

**Tests added:**
- `Tests/SharedMeetingMicClaimTests.swift` — extended with the exact P2 race at the pure-decision level: claim minted for identity A, current identity B, `isCaptureSessionActive: true` → `isClaimSessionAlive` returns `false`. Plus same-identity/inactive and nil-identity cases.
- `Tests/SharedMeetingMicRecorderTests.swift` (new) — drain-before-clear contract at the primitive level `finalizeStaleSharedMeetingMicClaim()` depends on (`ParakeetEngine` itself isn't in the fast-test seam, so this is the closest testable seam for P1's fix).

**Verification (all synchronous foreground):**
- `bash build.sh --no-open` — pass (deps stamp was fresh, no rebuild needed)
- `TRANSCRIPTED_DISABLE_FILE_LOGGER=1 bash run-tests.sh` — 12774 tests, 12774 passed, 0 failed (10 new tests: 4 `isClaimSessionAlive` cases + 4 `SharedMeetingMicRecorder` drain cases + 2 already-present suites re-verified)
- `bash run-integration-smoke.sh` — pass (app/core linkage, wake recovery, `MicRecordingFileMerger`)

Not merging — leaving that to the coordinator.

PR #1659, opened 2026-08-07
Fall back to the standard mic path when meeting-start voice processing fails to arm

r3dbars:
## What

Bounded meeting-start fallback for the "Apple voice processing requested but
not active" state (Matthew Bentley's Aug 3 `meeting_start_failed` /
`failure_kind=mic_unavailable` report).

## Root cause

With the opt-in VPIO mic mode on, `Audio.makeReadyMeetingInputGraph` arms
`setVoiceProcessingEnabled(true)` on each fresh input node. When that call
fails, `armVoiceProcessing` catches it and continues with `enabled=false` —
but the failed wrap can leave the node's `auAudioUnit.deviceID` untrustworthy,
so `MeetingInputDeviceSelectionPolicy.routeReadiness`'s raw-path device check
fails (`device_mismatch`), the graph attempt throws "The microphone route did
not become ready", and the built-in retry re-arms VPIO identically and fails
the same way. The start aborts before `isRecording`, so the watchdog/device
recovery never exists (`recovery_attempt_count=0`), and
`MeetingStartFailureClassifier` greps "microphone" → `mic_unavailable`, which
reads like a permission dead end even though permissions are granted.

Dictation survives the identical state because
`ParakeetEngine.applyDictationVoiceProcessingPreference` catches the VPIO
failure and continues with the current node mode — which is why "dictation
works, meetings don't" was the field signature.

## Fix (smallest safe shape)

- `VoiceProcessingStartFallbackPolicy` (pure, in `AudioCaptureStartState.swift`):
  the existing second graph attempt runs with VPIO suppressed **only** when the
  first attempt showed VPIO requested-but-inactive. One retry per graph build,
  never re-arms, never triggers for software-AGC/raw users, never triggers when
  VPIO armed fine.
- `armVoiceProcessing(on:suppressedByStartFallback:)` forces the existing
  preference-off/disarm branch for that one retry without mutating
  `enableVoiceProcessing`, so diagnostics keep reporting the user's real request.
- Permission gating (`MeetingRecordingStartGate`, TCC checks in `Audio.start()`)
  and the mic/system readiness latches (`AudioCaptureStartState.meetingCaptureOutcome`,
  first-frame streaming gates, 12s start deadline) are untouched — the fallback
  graph must still stream real frames from both sources to report ready.
- If the fallback also fails, the thrown error names voice processing, keeps
  the fallback attempt's underlying error (message + `NSUnderlyingErrorKey`),
  and classifies as new stable `failure_kind=voice_processing_unavailable`
  instead of `mic_unavailable`.
- Sanitized aggregate telemetry: new `voice_processing_start_fallback`
  (`none`/`attempted`) snapshot field rides `meeting_recording_started` /
  `meeting_recording_start_failed` (analytics registry + reviewed-properties +
  reliability-packet allowlist updated). Combined with the event name this
  distinguishes initial failure vs fallback attempt vs final outcome. No raw
  device names, paths, or text.

Note: `makeReadyMeetingInputGraph` is shared with mid-recording device
recovery, so a VPIO arm failure during recovery's rebuild gets the same
bounded non-VPIO retry. Recovery counts/cooldowns/give-up behavior unchanged.

## Deterministic coverage

- `Tests/TranscriptedCoreTests/AudioTests/VoiceProcessingStartFallbackPolicyTests.swift` (new):
  recoverable state selects fallback; bounded to one attempt; VPIO-off users
  unchanged; VPIO-armed-fine failures keep the requested mode; pre-arm failures
  don't engage; fallback path still requires both streams for readiness;
  telemetry rawValues stable.
- `Tests/MeetingStartFailureClassifierTests.swift`: the exact core
  fallback-failure message → `voice_processing_unavailable`; permission wording
  still wins; system-audio and mic buckets unchanged.
- `Tests/TranscriptedCoreTests/AudioTests/AudioPipelineDiagnosticsSnapshotShapeTests.swift`:
  new field mapped, shape guarded.
- Existing untouched guards: `MeetingRecordingStartGateTests` (genuine missing
  mic permission still blocked), `AudioCaptureStartStateTests` (system-audio
  failure still blocked), `VoiceProcessingBindResultTests`/`VoiceProcessingDisarmOutcomeTests`.

## Proof boundary

Deterministic tests + full build/test battery only. **No hardware proof**: the
failing HAL state (VPIO enable rejected on built-in mic @48kHz) was not
reproduced on real hardware, and no AirPods/Bluetooth/Zoom/TCC claims are made.
Matthew's fix needs a retest on his machine after release.


r3dbars:
Integration note: branch updated onto main (now including the quiet-library redesign #1660 and the dead-code sweep #1662). Full local verification on the merged tree: `build-deps --force` + `build.sh --no-open` + 12,073 fast tests + integration smoke + 984 Core package tests, all green. Outstanding caveat unchanged: the Bentley-machine retest gates *release*, not this merge.

PR #1675, opened 2026-08-10
Storage/About simplification: one cleanup action, tri-state updates

r3dbars:
## Summary

Part 4 of the settings-simplification stack (stacked on the General regroup).

**Storage**
- Merge the three overlapping destructive cache buttons (Reclaimable = stale + optional Whisper) into one primary **"Free Up Space"** action; per-cache metrics and the narrower Whisper/stale removals move behind a "Storage details" disclosure. All callbacks + confirmation alerts unchanged.
- Remove the "Compress WAV to M4A automatically" row — styled like a setting, but it had no control (always-on behavior). Its copy folds into the "Delete audio after" caption.
- Retitle "Support Folders" to "Advanced" (pure debug path-reveals shouldn't sit on par with Capture Library).

**About**
- Fold the two coupled Sparkle booleans into one segmented **"Automatic updates"** control: Off / Notify me / Download automatically. The controller already enforces the ladder invariant (checks-off forces downloads-off, downloads-on forces checks-on), so the four bool combinations collapse 1:1 onto these three states — no migration. Writes route through the existing setters so `update_setting_changed`/`settings_toggle_changed` telemetry is unchanged; the Download position is hidden when Sparkle reports downloads unavailable (previously a silently disabled toggle).
- "Send diagnostics" now explains *why* it's disabled (unconfigured crash reporting, or the General > Advanced > Privacy toggle) instead of greying out silently.

`docs/sparkle-updates.md` updated per its own same-change convention.

## Verification

```
export TRANSCRIPTED_DISABLE_FILE_LOGGER=1
bash build.sh --no-open   # passed
bash run-tests.sh         # 12075 tests, 0 failed
```

## Test plan

- [ ] Storage shows one Free Up Space action; details disclosure reveals per-cache metrics/removals
- [ ] Whisper removal still blocked while a Whisper model is selected
- [ ] About's Automatic updates control cycles Off / Notify me / Download automatically and persists across relaunch
- [ ] With crash reporting off, Send diagnostics shows the explanatory caption

### DIFF
--- a/Sources/UI/Settings/Pages/AboutSettingsPage.swift
+++ b/Sources/UI/Settings/Pages/AboutSettingsPage.swift
@@ -263,8 +263,11 @@ struct AboutSettingsPage: View {
 
     private var currentAutomaticUpdatePolicy: AutomaticUpdatePolicy {
         let settings = sparkleUpdater.automaticUpdateSettings
-        if settings.automaticDownloadsEnabled { return .download }
-        if settings.automaticChecksEnabled { return .notify }
+        // Clamp downloads-enabled to .notify when Sparkle reports downloads
+        // unavailable: the .download segment isn't offered then, and checks
+        // are what actually still run.
+        if settings.automaticDownloadsEnabled, settings.automaticDownloadsAllowed { return .download }
+        if settings.automaticChecksEnabled || settings.automaticDownloadsEnabled { return .notify }
         return .off
     }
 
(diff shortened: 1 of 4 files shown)


r3dbars:
Codex review verdict: PASS with one P2, fixed in 1b1d3732: the tri-state binding now only calls a Sparkle setter when the underlying boolean actually flips (the controller's setters emit update_setting_changed and kick refreshUpdateStatus unconditionally, so Download→Notify was firing a phantom automatic_checks event plus an unneeded update probe). Also clamped the derived policy to Notify when downloads are enabled but Sparkle reports them unavailable, so the segmented control always has a selectable current value.

PR #1682, opened 2026-08-11
Offer retry for failed meetings whose audio survived

r3dbars:
## Why

A beta user reported that failed meetings show up on Home with no way to force them through. They were right, and the reason is a real bug rather than a missing feature.

When a meeting's microphone track breaks but system audio records cleanly, the recording is fully transcribable — `TranscriptionPipeline` already drops an unusable mic and continues as a system-audio-only run. Home refused to offer it anyway, because retry availability came entirely from the persisted failure message, and `microphoneAudioUnusable` is hardcoded permanent. `HomeFailedMeetingInlinePresentation` then returned `canShowRetryAction: false`, so **Try again** was not merely disabled — it was never constructed. The row read "Needs attention" next to perfectly good audio of everyone else on the call.

The product contradicted itself: `PipelineError.microphoneAudioUnusable.localizedDescription` tells the user to "Open Transcripted Home to retry the saved meeting", while the policy hid that button.

The asymmetry is the giveaway that this was an oversight rather than a decision:

| Surviving source | Old verdict |
|---|---|
| mic fine, system audio missing | retryable (mic-only path, with an explicit carve-out) |
| system audio fine, mic unusable | permanent |

Same shape of loss, opposite answers. The first case got a rescue path and an exception in the code; the mirror case never did.

Note the affected population: current builds already handle a broken mic gracefully (`testSavedTranscriptResolutionExcludesUnusableMicAndMarksDegraded` pins that), so this mostly strands **rows persisted by older builds** whose pipeline aborted when one source broke. Most of those predate `errorKind` and classify through the legacy string fallback, so the fix is written not to depend on which exact message they carry.

## Product Impact

- Affects: `meetings`
- Lane: `meeting reliability`
- Why this matters: recoverable meeting audio is currently unreachable through the UI. The only workaround is Show Audio → reveal in Finder → re-import as a fresh file, which loses the local track, the original title, and the recording date, and leaves the dead row behind. Nothing in the app hints that this is even possible.

## What changed

- **`PipelineErrorKind.describesRecoverableSource`** (new) splits retry availability into the two questions it always was. Does the failure kind name one broken capture source (recoverable — the pipeline works around it), or audio that genuinely holds nothing? `emptyAudioFile`, `microphoneAudioUnusable`, `invalidAudioFormat`, and `missingSystemAudio` are recoverable; `noSpeechDetected` and `recordingTooShort` stay permanent so a retry cannot burn inference time reproducing a known-empty result.
- **`FailedTranscription.isRetryable`** now consults that alongside `isRetryable`, on both the typed and legacy paths. The legacy permanent-keyword list drops the single-source entries and keeps only the content-empty ones. `legacyPipelineError` became `legacyErrorKind` so both paths run one policy instead of two that can drift.
- **`FailedRecordingSignalProbe`** (new, Core) answers the factual half: do the surviving files actually carry audible signal? It streams bounded windows rather than `loadAndResample`-ing whole artifacts, since it runs for every failed row and a 2h capture allocates hundreds of MB.
- **`FailedMeetingUsableAudio`** (new, app) is the presentation-side tri-state. `FailedMeetingStore` owns the background probe pass and caches verdicts per audio URL, so WAV→M4A recompression invalidates entries for free and an `.inconclusive` file is not re-probed on every refresh.
- **Copy**: the unusable-mic detail advised re-recording without mentioning the recovery that actually works. It now leads with trying again for the other side of the call.
- Subsystem `CLAUDE.md` files and the fast-test source list updated for the two new source files.

### The probe is deliberately three-valued

Worth a reviewer's attention, because it is the one place I deviated from the obvious design.

My first version sampled six windows spread across the file to keep cost flat. That is wrong in the dangerous direction: a call that opens with silence, has real conversation in the middle, and ends quiet can fall between the windows. The probe would report silence and hide the retry button — reintroducing this exact bug in a new form.

So `.absent` is now returned **only** after the entire artifact has been read. Unreadable files and anything past the scan budget (20 min of audio) return `.inconclusive`, and `FailedMeetingUsableAudio.unknown` keeps retry offered. The probe can only ever take the button away when it is certain; every uncertain path stays optimistic. `testProbeFindsAudioStrandedInTheMiddle` fails against the original design and passes against this one.

### Two inverted assertions

These encoded the old behavior, so they were part of the bug. Both are inverted with rationale in the test body:

- `testFailedTranscriptionRetryabilityKeepsUnusableMicrophonePermanent` → `...OffersUnusableMicrophoneWhenSystemAudioSurvived`
- `testFailedTranscriptionManagerLoadsLegacyQueueFileWithoutErrorKind` no longer expects an "Empty audio file" row to be permanent

`PipelineError.isRetryable` is untouched. It answers a different question (was the error transient, while it is in flight) and is separately contract-tested; only the row-level policy moved.

## How I checked it

- [x] Selected checks from `.agents/test-matrix.yml` for the files changed (ran the union for `Sources/TranscriptedCore/**`, `Sources/Meeting/**`, root `Tests/*.swift`, and `scripts/entrypoints/run-tests.sh`)
- [x] `bash run-tests.sh` — 12174 tests, 12174 passed, 0 failed
- [x] Performance budget passed (bundle gate via `build.sh --no-open`; launch to interactive 1203.5ms against a 3000ms budget)
- [x] `bash run-integration-smoke.sh` — exit 0
- [x] `swift test` — 1026 tests, 0 failures, 13 skipped
- [x] `python3 scripts/dev/check-build-source-lists.py`, `bash -n` on both test-runner scripts
- [ ] Manual check: **not done — see below**

### Not verified

Calling this out rather than letting the checklist imply otherwise:

1. **The button has not been clicked.** The app builds and the logic is unit-tested, but I did not launch it and drive a failed row through Home. The end-to-end click, the probe landing and the row updating, and the resulting transcript are unproven.
2. **No genuine 1.5-era row was available.** Coverage uses a synthetic legacy row with the `errorKind` key stripped from the JSON, which matches the shape on disk. It is not a real one from an affected machine.

## Risk Review

- [x] Privacy / local-first behavior reviewed — the probe reads local audio already owned by the failed queue, decodes in memory, and emits no telemetry, paths, or content
- [x] Storage path or migration impact reviewed — no schema, no new persisted field, no migration. `errorKind` is untouched and the verdict cache is in-memory only
- [x] Public-facing copy stays concrete and matches current product scope — one detail string reworded, its assertion updated
- [x] Release/update impact reviewed — no version, appcast, or packaging change
- [ ] Agent PRs link the issue/workpad and stay draft until human review — **draft; no issue exists yet for this report**
- [x] UI changes include sanitized `.agent-review/visuals/` evidence — **N/A: no visual change.** The affected control is an existing button whose visibility condition changed; no new layout, and I have no screenshot because the app was not launched

### Reviewer notes on risk

The behavioral risk is narrow but worth a second opinion on two points:

1. **`missingSystemAudio` is now retryable unconditionally**, where before it was retryable only when `systemAudioURL == nil`. Retry re-transcribes existing files and never re-captures, so the permission state that caused the failure is irrelevant by then; the result is a mic-only transcript. I believe that is strictly better than a dead row, but it is a deliberate widening.
2. **More rows will now accept a retry**, which means more opportunities to occupy the single-flight pipeline. `retryFailedTranscription` still guards on `activeTasks.isEmpty`, so concurrency is unchanged — but the frequency of user-initiated retries could go up.

## Notes

Follow-ups deliberately left out of scope, from the same user report:

- No drag-and-drop anywhere in the app; audio import is file-picker only.
- `allowsMultipleSelection = false` on the import panel, so no batch import.
- No paired import: a single file lands in the system-audio slot, so there is no way to hand the app a matched local + remote pair as one meeting. This one needs an alignment story before it is safe — two files from separate processes share no clock, and guessing the offset produces a confidently mis-ordered transcript, which is worse than two separate ones.

## Agent handoff

`COORD_DONE: GREEN | this PR | split failed-meeting retry availability into failure-kind policy plus a fail-safe audio signal probe, inverted 2 assertions that encoded the bug, reworded unusable-mic copy | none | whether to widen missingSystemAudio unconditionally (see Risk Review) | build-deps --force, build.sh --no-open, run-tests.sh 12174 passed, swift test 1026 passed, run-integration-smoke.sh exit 0, check-build-source-lists.py, agent-preflight.sh | launch the app and click Try again on a failed row to close the manual gap`


r3dbars:
Reviewed by Sonnet (MERGE_WITH_NITS) and Codex (PASS). Codex's strongest P2 — an utterance straddling the probe's 8s window boundary could split below the 0.2s threshold on both sides and wrongly report .absent, re-introducing the false-negative class this PR exists to kill — is fixed in the commit just pushed (overlapping windows sized from the detector's own threshold, plus a regression test). Remaining follow-ups noted by the reviewers (mtime-aware verdict cache, per-source retry URLs, missing-file OR-gate in audioFilesExist, untested store scheduling) are quality improvements, not regressions, and are left for follow-up work.

PR #1684, opened 2026-08-11
Settings and Home polish: update menu, row rename, and dependency hygiene

r3dbars:
## What merged

This PR was a combined Settings, Home, and repository-hygiene pass:

- kept the `Settings` title but removed the redundant “Everything in one place” tagline
- replaced the cramped automatic-update segmented control with a readable menu
- renamed the update-off choice from `Off` to `Never check`
- moved meeting rename from inline click-to-edit into the row's `…` menu and wired its accessibility identifier
- removed transcript/playback highlighting and changed the collapsed transcript to the first eight lines
- deleted four accidentally tracked absolute `deps-*` symlinks and hardened their ignore rules
- updated the affected UI and Sparkle docs

The static transcript behavior is intentional: retained audio still plays, but transcript rows do not highlight or follow the playhead. Follow-up #1685 keeps that behavior and changes only the updater label to the truthful `Check on launch`, while recording the product capabilities future simplification work must preserve.

## Verification at merged head `7db88fa`

- `bash build.sh --no-open` — passed
- `bash run-tests.sh` — 12,125/12,125 passed
- GitHub Swift CI — passed (`checks`, `app-build`, `spm-tests`, and `build-and-test`)
- GitHub Repo Hygiene — passed
- hardware smoke — not run by CI

No preference keys, analytics IDs, or storage paths changed.

## Merge

Merged into `main` as `561d49af` on August 11, 2026.


r3dbars:
Correction pushed: the first commit removed the whole intro block, which left the page with no heading — the settings window sets `titleVisibility = .hidden`, so nothing else was displaying "Settings", and every sibling page still shows its own title. The "Settings" title is back in the title-only form the Speakers page already uses; only the "Everything in one place" tagline stays removed.

PR #1712, opened 2026-09-04
Release the system-audio write hold when a recovery is abandoned; drop deleted system URLs at stop

r3dbars:
## Summary

Two confirmed defects from the adversarial post-merge review of #1711 (verdict: ship with fix).

**1. System-audio write hold leaked on every non-successful recovery (silent data loss).** The hold was armed on `.deviceSwitch` and released only in `recordSystemAudioGap`, which runs on `.gap`, and `.gap` is sent only after a restarted stream delivers a buffer within 2 s. A recovery that was cancelled, superseded, threw, or got no buffer in time left the hold armed, and `AudioFileManager` then dropped every system buffer for the rest of the meeting. The new wake wiring kicks `recoverAfterSystemWake()` unconditionally, so a slow post-wake restart was exactly the case that lost all remaining system audio.

- New `SystemAudioRecoveryEvent.recoveryAbandoned`, sent from a `defer` in the SCK recovery block on every exit that never reached `.gap`, before `finishRecovery` so a successor attempt cannot arm ahead of the release.
- `Audio` releases the hold on `.recoveryAbandoned` on the sending thread (the same thread that armed it), and `recordSystemAudioGap` releases it even when it bails on `isRecording`.

**2. A failed system-audio start left a phantom system URL.** The WAV was removed but `originalSystemAudioFileURL` stayed set, so `stop()` handed the pipeline a URL for a deleted file: the meeting skipped the missing-system-audio warning and reported `system_stream_present: true`. `resolvedSystemAudioFileURL` now returns nil for a candidate that is not on disk, and the start-failure path clears the original URL.

## Tests

- `testAbandonedRecoveryReleasesSystemWriteHold`, `testGapWhileNotRecordingStillReleasesSystemWriteHold` (SystemAudioRecoveryParityTests)
- `testCancelledRecoveryReportsAbandonmentInsteadOfAGap` (SCKAudioCaptureInterleavingTests)
- `testResolvedSystemAudioURLIgnoresOriginalThatNoLongerExists` (AudioFileManagerTests)

## Verification

`bash build-deps.sh --force` → `bash build.sh --no-open` → `bash run-tests.sh` → `bash run-integration-smoke.sh` → `swift test`, all green locally on this branch.

Still owed on real hardware before the 1.1.57 cut: sleep/wake during a live meeting with audio playing, confirming the saved system WAV contains audio after the wake.

## Not in this PR (review follow-ups)

- Raw-mode quiet-mic detector can prompt for mic boost when the user only listens for 30 s (product call).
- A placeholder-only failed row passes `audioFilesExist()` and can be retried forever (several tests pin placeholder rows as retryable when a system track survives; needs a narrower fix).

### DIFF
--- a/Sources/TranscriptedCore/Audio/Audio.swift
+++ b/Sources/TranscriptedCore/Audio/Audio.swift
@@ -402,7 +402,7 @@ public class Audio: ObservableObject, @unchecked Sendable {
         guard isRecording else {
             // No pad to write, but the hold `.deviceSwitch` armed must not
             // outlive the recovery that armed it.
-            setSystemRecoveryWriteHold(false)
+            releaseSystemRecoveryWriteHold()
             return
         }
         appendRecordingGap(AudioGap(
@@ -414,7 +414,7 @@ public class Audio: ObservableObject, @unchecked Sendable {
             duration: duration,
             generation: recordingSessionGeneration
         )
-        setSystemRecoveryWriteHold(false)
+        releaseSystemRecoveryWriteHold()
     }
 
     /// Commit recovery artifacts only while the recovery still owns the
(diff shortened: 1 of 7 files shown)


r3dbars:
## Codex review (two passes) and follow-up

**Built-in `codex review`:** one P1. On the first-buffer timeout the restarted stream was left running; once the attempt was abandoned the hold came down and a late buffer would be written with no pad for the outage, shifting the rest of the system track against the microphone. **Fixed** in the follow-up commit: the timed-out stream is stopped before the attempt is abandoned, and the wait is 5 s instead of 2 s (a post-wake ScreenCaptureKit restart can take that long to deliver its first frame; giving up at 2 s costs the rest of the meeting's system audio).

**Adversarial `codex exec` pass:** two P1s.
- A predecessor recovery's `.gap` is handled on main, so a successor could arm before the predecessor's release ran and lose its hold. **Fixed:** the hold is a count now (arm on `.deviceSwitch`, release on `.gap` or `.recoveryAbandoned`, reset at recording start), with `testOverlappingRecoveriesKeepTheHoldUntilTheLastOneEnds`.
- A pre-existing race between `stop()` resolving the system URL and a superseded setup's `cleanupAbandonedSetup` deleting that WAV. **Not addressed here:** it predates this PR, the existence check narrows but does not close it, and the pipeline already falls back to mic-only when the file is gone. Worth its own issue.

Codex also confirmed: the `.recoveryAbandoned` ordering is sound, no non-success exit bypasses the `defer`, no allocation/lock/I/O was added inside a render or sample-buffer callback, the existence check cannot reject an ordinarily live file, and all four original tests fail against the old code.

Local verification after the follow-up: `build-deps --force` → `build.sh --no-open` → `run-tests.sh` (12,424 passed) → `run-integration-smoke.sh` → `swift test`, all green.

PR #1713, opened 2026-09-04
Stop the Sentry SDK from filing failed HTTP requests as errors

r3dbars:
## Summary

Sentry Cocoa captures every URLSession response with a 5xx status as its own error event by default. On 1.1.56 that produced two `HTTPClientError: HTTP Client Error with status code: 503` issues (APPLE-MACOS-28 and 29) from one machine, with no URL attached: the appcast or analytics endpoint having a bad hour, filed next to real capture failures. The app never acts on those responses, so the events carry nothing actionable and only bury the issues that are.

- `options.enableCaptureFailedRequests = false` in the Sentry bootstrap, next to the other automatic reporting it already keeps off. The bundled Sentry 9.10.0 framework exposes the option.
- New fast test `testCrashReporterOptions` pins the option set (PII, session tracking, breadcrumbs, stack attachment, failed requests) so a refactor cannot quietly widen what the SDK reports on its own.

No app behavior changes. Only what the SDK reports changes.

## Verification

`bash build-deps.sh --force` → `bash build.sh --no-open` → `bash run-tests.sh`, green locally. Codex review (built-in and adversarial pass) attached in the comments.

## After merge

Archive APPLE-MACOS-28 and 29 in Sentry.

### DIFF
--- a/Sources/Observability/CrashReporter.swift
+++ b/Sources/Observability/CrashReporter.swift
@@ -38,6 +38,11 @@ final class CrashReporter {
         // choice and later opt-outs aligned with Release Health envelopes.
         options.enableAutoSessionTracking = false
         options.enableNetworkBreadcrumbs = false
+        // The SDK otherwise files every URLSession 5xx as its own error issue
+        // (`HTTPClientError: HTTP Client Error with status code: 503`) with no
+        // URL attached. Those are the appcast or analytics endpoints having a
+        // bad hour, not app failures, and they bury the real issues.
+        options.enableCaptureFailedRequests = false
         options.maxBreadcrumbs = 0
         options.attachStacktrace = false
         options.beforeBreadcrumb = { _ in nil }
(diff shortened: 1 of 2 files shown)


r3dbars:
## Codex review, three rounds

- **Round 1 (adversarial pass):** MERGE. Confirmed the bundled Sentry 9.10.0 exposes `enableCaptureFailedRequests`, that nothing in the app consumes the automatic 5xx issues (analytics retries 5xx itself, Sparkle handles update failures through its delegates), and that the new fast test follows the runner convention. One P3: a substring pin could pass with a commented-out or later-overridden assignment.
- **Round 2:** DO NOT MERGE on two test-only P3s: `/* */` blocks and `#if false` regions still counted as live, and nothing checked the option is applied before `SentrySDK.start`.
- **Round 3 (final commit):** MERGE, no findings. Verified exact-once live assignment, dead code ignored, and every pinned option preceding the start call.

The production change is one line; the rest is the pin test getting stricter each round.

Local verification: `build-deps --force` → `build.sh --no-open` → `run-tests.sh`, green on the branch (the final commit is test-only and is re-running the same chain).

PR #1714, opened 2026-09-04
Report why a meeting's capture grade is what it is

r3dbars:
## Summary

`capture_quality` reaches Sentry, PostHog, the reliability packet, and the stop diagnostics with every input the grade is computed from, but not the grade's cause. On 1.1.56 that left 505 `recording_capture_degraded` events across 72 devices that could only be split by guessing: silence (fixed in #1710) explained some, gaps and route changes some more, and a large block had healthy system audio, near-perfect buffers, and nothing switching, with no way to tell the warning latch from anything else.

`RecordingHealthInfo.qualityReason` is one categorical token set where the grade is decided:

- from the live snapshot: `system_audio_failed`, `buffer_loss`, `device_switches`, `interruptions`, `none`
- from the finalization markings: `system_audio_missing`, `system_audio_warning`, `microphone_unusable`

It rides along as `quality_reason` on the capture-health snapshot, the degraded diagnostic, the stop diagnostic and analytics, allowlisted in `SentryEventPolicy` and `Resources/analytics-events.psv`, the reliability packet, and the local sanitizer. No behavior changes; the grade itself is untouched.

## Tests

`testQualityReasonNamesTheDominantCauseOfTheGrade` covers every branch, including buffer loss outranking the one-step interruption downgrade and mic attenuation staying a fact rather than a reason.

## Verification

## Why this matters for 1.1.57

With this key in place the first week of 1.1.57 answers "what is left in the degraded bucket" directly, instead of a second audit guessing at it.


r3dbars:
## Codex review, four rounds

- **Round 1 (adversarial pass):** DO NOT MERGE, five findings, all fixed in the next commit: `quality_reason` had to be registered in `Resources/analytics-reviewed-properties.psv` (the taxonomy fast test enforces it); the stop path's warning latch overwrote a more specific `system_audio_failed`; `successRate == 0` was a proxy for an explicit failure (mislabelling a genuine 0% buffer rate and failure-plus-three-switches); `meeting_capture_stopped_under_controller` also emits the key via `snapshotProperties` but its allowlist row lacked it; the mic-only fallback initializer left the reason at `none`.
- **Round 2:** all five confirmed RESOLVED; three new: the reliability packet still filed a degraded stop as `success`; two live counters were read several times while recovery can bump them; the unexpected-stop diagnostic lacked the key. All fixed.
- **Round 3:** all three RESOLVED; one new P2: the derived packet precedence also reclassified `meeting_recording_cancelled`. Fixed by gating the derived verdicts to success-default events, with a cancellation test.
- **Round 4:** RESOLVED, no new findings. MERGE.

No behavior change to the grade itself. The packet outcome for a normal stop now follows: timeout → no audio → degraded → recovered → success.

PR #1715, opened 2026-09-04
Treat a placeholder-only failed row as unretryable

r3dbars:
## Summary

Follow-up from the adversarial review of #1711.

`FailedTranscription.audioFilesExist()` answers true when either audio file survives, so a row whose system track is gone but whose silent microphone placeholder remains passes both retry gates in `TranscriptionTaskManager.retryFailedTranscription`. The retry transcribes 2.5 s of silence, fails with "no speech detected", and the row returns to the queue with "the audio was kept, try again" copy: a dead row the user can retry forever.

- The placeholder now counts only alongside a surviving system track. `FailedTranscription.isMicrophonePlaceholder` matches by the `microphone_placeholder` prefix, so task-suffixed and compressed (`.m4a`) placeholders are covered.
- Rows with a real mic recording, and placeholder rows whose system track is still on disk, are unchanged; the existing tests for both stay green.
- A placeholder-only row now takes the existing "audio files no longer exist" path: one heal attempt, then removal from the queue.

## Tests

`testAudioFilesExistIsFalseWhenOnlyThePlaceholderMicSurvives` (plain, task-suffixed, and compressed placeholders; retryable again once the system track exists) and `testAudioFilesExistIsTrueWhenOnlyARealMicFileSurvives`.

## Verification

### DIFF
--- a/Sources/Meeting/MeetingAudioStorageManager.swift
+++ b/Sources/Meeting/MeetingAudioStorageManager.swift
@@ -1220,7 +1220,10 @@ enum MeetingAudioStorageManager {
         guard !isSymbolicLink(url, fileManager: fileManager) else { return false }
         let values = try? url.resourceValues(forKeys: [.isRegularFileKey])
         guard values?.isRegularFile == true else { return false }
-        return url.deletingPathExtension().lastPathComponent == "microphone_placeholder"
+        // Prefix match, shared with the pipeline: the archiver can suffix a
+        // colliding placeholder (`microphone_placeholder-2.wav`), and an exact
+        // stem would then treat the whole directory as not-ours.
+        return FailedTranscription.isMicrophonePlaceholder(url)
     }
 
     private static func isStaleTemporaryAudioFile(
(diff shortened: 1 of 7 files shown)


r3dbars:
## Codex review, four rounds

- **Round 1 (adversarial pass):** DO NOT MERGE. Three paths where the placeholder check did not reach or did harm: partial capture-library relocation could strand real system audio behind a placeholder mic; the archiver renamed placeholders to `microphone.<ext>` so a promoted placeholder lost its identity; placeholder-only rows were not pruned at load. All addressed.
- **Round 2:** relocation and load pruning RESOLVED; archive identity partially (new archives only). New: my relocation test asserted the wrong list; the compression ownership gate broke on a `-N` collision suffix; a row can be persisted while its system copy failed (kept as-is: the original scratch track still exists then, and load-time pruning covers later cleanup); legacy rows already renamed by older builds cannot be recognised by name (deferred: needs a migration, and a filename heuristic could misclassify real recordings).
- **Round 3:** the relocation guard produced a row split across two libraries, hidden in both; the prefix predicate was over-broad (`microphone_placeholder_interview.wav` would have read as silence). Both fixed: a placeholder row whose system track was not copied stays whole in the old library (test drives both managers), and the predicate matches only the generated forms (exact stem, `_<UUID>`, numeric `-N`).
- **Round 4:** all RESOLVED, no diff-introduced path that drops or hides legitimately retryable audio. MERGE. Codex notes a symmetric pre-existing relocation gap when only system audio is copied, already on `main` and not worsened here.

Follow-up worth its own issue: legacy placeholder rows renamed to `microphone.wav` by older builds.

PR #1716, opened 2026-09-04
Never prompt for mic boost without AGC gain evidence

r3dbars:
## Summary

Follow-up from the adversarial review of #1711.

The quiet-mic detector's gain-absent branch (Raw and Apple voice processing, where no software AGC runs) qualified a tick on raw level alone: 30 s under about -26 dBFS with some room tone. That is what a normal meeting looks like while the user listens, so the branch fired the mic-boost prompt mid-meeting and stamped `micAttenuatedByCallApp` into the saved health metadata and the capture-health analytics for users whose mic was fine.

The AGC path has real evidence: software gain pinned at max while the raw mic stays quiet means something upstream is holding the device down. With no gain to pin there is nothing to distinguish "held down" from "not speaking", so the detector now never fires without it.

- Default (auto-level) users are unchanged.
- Raw and voice-processed users lose the in-meeting boost prompt that #1711 added for them. It was firing on no evidence, so the prompt they lose is the false one; stop-time classification is untouched.

## Tests

`testRawOffQuietActivityNeverFires` replaces `testRawOffQuietActivityFiresBoostCue`; the total-silence and AGC-path tests are unchanged.

## Verification

### DIFF
--- a/Sources/Meeting/MeetingCaptureSupport.swift
+++ b/Sources/Meeting/MeetingCaptureSupport.swift
@@ -513,7 +513,8 @@ enum MeetingCaptureVolumeDiagnostics {
                 captured: capturedInputChange.droppedState,
                 capturedContextPresent: capturedInputContextPresent,
                 defaultInput: context["default_input_volume_dropped"]
-            )
+            ),
+            agcActive: context["realtime_agc"] == "true"
         )
 
         context["output_ducking_detected"] = outputDuckingState(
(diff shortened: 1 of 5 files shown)


r3dbars:
## Codex review, four rounds

- **Round 1 (adversarial pass):** DO NOT MERGE. The live detector was fixed, but the app's stop-time classifier still labelled a quiet mic with no scalar drop as `voice_processed` without gain evidence, and that label stamps `micAttenuatedByCallApp` on the saved transcript.
- **Round 2:** gating the stop-time label on `realtime_agc` was wrong in both directions: the flag only proves an AGC object existed, not that gain was pinned, and accepting Boost swaps the AGC out for voice processing so a detector-confirmed episode was erased at save time.
- **Round 3:** the session-latched live cue (`micAttenuationCueObserved`) became the evidence, but it was consulted after the lifetime-peak gate (a later loud stretch could still read `none`), two stop contexts omitted it, and two fixtures were stale.
- **Round 4:** all RESOLVED. The cue is evaluated first (a scalar drop still wins), all three stop contexts carry it, fixtures fixed. Codex judged the deliberate `quiet_mic_unrecovered` gate on the saved-transcript note sound and conservative. MERGE.

Net behavior: default (auto-level) users unchanged; Raw and Apple voice-processing users no longer get an evidence-free mic-boost prompt or attenuation stamp; a detector-confirmed episode is recorded as `voice_processed` regardless of what the lifetime peaks did afterwards.

PR #1717, opened 2026-09-04
Prepare 1.1.57

r3dbars:
## Summary

Source version metadata only: `Info.plist` moves to 1.1.57. `docs/appcast.xml` still advertises 1.1.56 until a notarized DMG is published, which the nightly security check reports as the expected `appcast-release-candidate` watch item rather than a finding.

This is the preparatory bump the release docs call for before packaging.

## Verification

`python3 scripts/ops/nightly-security-check.py --strict ...` scores 100/100 with no deterministic findings. Per `scripts/dev/agent-preflight.sh` for this diff: `bash build.sh --no-open`, `bash run-tests.sh`, `python3 scripts/dev/check-analytics-emitters.py`.

## What still gates the actual release

Two credentials are absent on the build machine and only the owner can supply them:

- no notarytool keychain profile, so a distribution DMG cannot be notarized
- no Sparkle EdDSA private key, so an appcast entry cannot be signed

Until both exist, publishing would ship a Gatekeeper-blocked DMG and an update feed no installed client would accept.


r3dbars:
## Blocked: this bump cannot land on its own

`ReleaseMetadataContractTests` fails with 7 assertions, and it is right to. The contract is that `Info.plist`, `docs/appcast.xml`, and `Casks/transcripted.rb` all describe the same release:

```
Homebrew cask version should match the app bundle version — expected 1.1.57, got 1.1.56
latest appcast item should match the app bundle version — expected 1.1.57, got 1.1.56
latest appcast enclosure should point at the matching GitHub DMG — .../v1.1.57/... vs .../v1.1.56/...
```

So a standalone preparatory bump is not a valid repo state, whatever the nightly security check tolerates as a watch item. The version, the appcast entry, and the cask have to land in one change, and the appcast entry needs a real DMG URL, asset length, and Sparkle EdDSA signature, which in turn need the notarized artifact.

Converted to draft. This branch becomes the release PR once the two missing credentials exist on the build machine:

- a notarytool keychain profile, so the DMG can be notarized
- the Sparkle EdDSA private key matching `SUPublicEDKey` in Info.plist, so the appcast entry can be signed

At that point the release commits (appcast + cask) join this one and the contract test goes green.

